PR #92, opened 2021-10-11
deploy: Add external-label support

ArthurSens:
Looking forward to installing parca in a multi-cluster setup. Support for external-labels in the jsonnet library would be great 🙂


ArthurSens:
I've removed the trailing comma anyways

PR #109, opened 2021-10-23
systemd: Deal with non-resolvable units and slice locations

zecket495:
My system.slice on OpenSUSE tumbleweed seems to be at a different place. Also don't fail to render the status page when a unit can not be resolved.

### DIFF
--- a/pkg/agent/systemdmanager.go
+++ b/pkg/agent/systemdmanager.go
@@ -47,6 +47,7 @@ type SystemdManager struct {
 	mtx               *sync.RWMutex
 	tmpDir            string
 	profilingDuration time.Duration
+	cgroupPath        string
 }
 
 type SystemdUnitTarget struct {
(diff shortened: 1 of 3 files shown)


pkg/agent/systemdmanager.go
```diff
@@ -143,14 +148,14 @@ func (m *SystemdManager) Run(ctx context.Context) error {
...
 
 func (m *SystemdManager) reconcileUnit(ctx context.Context, unit string) error {
-	f, err := os.Open(fmt.Sprintf("/sys/fs/cgroup/systemd/system.slice/%s/cgroup.procs", unit))
+	f, err := os.Open(fmt.Sprintf("%s/%s/cgroup.procs", m.cgroupPath, unit))
 	if os.IsNotExist(err) {
 		m.mtx.Lock()
 		p := m.unitProfilers[unit]
 		if p != nil {
 			p.Stop()
 		}
```
kakkoyun, line 157:
```suggestion
		if p, ok := m.unitProfilers[unit]; ok && p != nil {
			p.Stop()
		}
```

Just a nitpicking. With the below change this would be more accurate.

kakkoyun (approved):
LGTM. Thanks a lot.

PR #116, opened 2021-10-28
pkg/agent: Add batching support for profile write requests

### LINKED ISSUE #14: Batch write requests
Currently, whenever a profile has been recorded, it is individually sent directly to a configured server. With potentially hundreds of profilers running on a single machine, it would make sense to batch these requests and only send them out once every 10 seconds or so. The write API already supports batched writes both in the series dimension as well as the number of samples.

Sylfrena:
Fixes #14

### DIFF
--- a/pkg/agent/batcher.go
+++ b/pkg/agent/batcher.go
@@ -1,0 +1,123 @@
+package agent
+
+import (
+	"context"
+
+	"sync"
+	"time"
+
+	"github.com/cespare/xxhash/v2"
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	profilestorepb "github.com/parca-dev/parca/gen/proto/go/parca/profilestore/v1alpha1"
+)
+
+type Batcher struct {
+	series      map[uint64]*profilestorepb.RawProfileSeries
+	writeClient profilestorepb.ProfileStoreServiceClient
+	logger      log.Logger
+
+	mtx                sync.RWMutex
+	lastProfileTakenAt time.Time
+	lastError          error
+}
+
+func NewBatcher(wc profilestorepb.ProfileStoreServiceClient) *Batcher {
+	return &Batcher{
+		series:      make(map[uint64]*profilestorepb.RawProfileSeries),
+		writeClient: wc,
+	}
+}
+
+func (b *Batcher) loopReport(lastProfileTakenAt time.Time, lastError error) {
+	b.mtx.Lock()
+	defer b.mtx.Unlock()
+	b.lastProfileTakenAt = lastProfileTakenAt
+	b.lastError = lastError
+}
+
+func (b *Batcher) Run(ctx context.Context) error {
+	// TODO(Sylfrena): Make ticker duration configurable
+	const tickerDuration = 10000000000
+
+	ticker := time.NewTicker(tickerDuration)
+	defer ticker.Stop()
+
+	var err error
+	for {
+		select {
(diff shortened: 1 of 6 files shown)


pkg/agent/batcher.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	return nil
+}
+
+func (batcher *Batcher) Scheduler(profileSeries profilestorepb.RawProfileSeries) {
+	batcher.mtx.Lock()
+	defer batcher.mtx.Unlock()
+
+	labelsetHash := Hash(*profileSeries.Labels)
```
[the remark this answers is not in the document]
brancz: I would keep it very simple until we have data it needs to be more sophisticated. Just maintain a slice and compare label-sets one by one. The default pod limit per Kubernetes node is 120 pods, most people don't modify this. So let's take an extreme example and say each pod has 10 containers, that's 1200 items to insert over 10 seconds. Even linear inserts should be totally fine at that scale, and if it's not (and data shows this) we do binary search.

pkg/agent/batcher.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func NewBatcher(wc profilestorepb.ProfileStoreServiceClient) *Batcher {
+	return &Batcher{
+		series:      make(map[uint64]*profilestorepb.RawProfileSeries),
+		writeClient: wc,
+	}
+}
+
+func (b *Batcher) loopReport(lastProfileTakenAt time.Time, lastError error) {
```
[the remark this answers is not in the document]
brancz: I actually like this to show in the web UI when the last batch send happened and whether there was an error.

pkg/agent/batcher.go
```diff
@@ -0,0 +1,137 @@
...
+		series:      []*profilestorepb.RawProfileSeries{},
+		writeClient: wc,
+		mtx:         &sync.RWMutex{},
+	}
+}
+
+func (b *Batcher) loopReport(lastProfileTakenAt time.Time, lastError error) {
+	b.mtx.Lock()
+	defer b.mtx.Unlock()
+	b.lastProfileTakenAt = lastProfileTakenAt
```
brancz, line 49: `lastProfileTakenAt` is probably not the accurate name, more like `lastBatchSentAt`

pkg/agent/batcher.go
```diff
@@ -0,0 +1,137 @@
...
+	b.lastError = lastError
```
brancz, line 50: don't feel as strongly as with the other variable, but if you're already changing that one, maybe this would could also instead be `lastBatchSendError`?

pkg/agent/batcher.go
```diff
@@ -0,0 +1,137 @@
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-ticker.C:
+		}
+
+		err := b.batchLoop(ctx)
+		b.series = []*profilestorepb.RawProfileSeries{}
```
brancz, line 68: This needs to be protected by the mutex because it can occur concurrently, I would say it would be best to do this at the end of the `batchLoop` function instead.

brancz (commented):
Only a few very small things, I think we're super close to merging this. Really nice work!! It reads super easily!

brancz (approved):
Really awesome! Nice work! :tada: :rocket:

kakkoyun (approved):
💯 👏

PR #171, opened 2021-12-16
pkg/agent: Support cgroups v2 unified, add basic test

### LINKED ISSUE #23: Add support for cgroup v2
`parca-agent` supports only `cgroup` v1 right now. The `systemd` already started supporting v2. 

Kubernetes changes are in the pipeline (https://github.com/kubernetes/kubernetes/pull/85218 & https://github.com/kubernetes/enhancements/issues/2254).

derekparker:
This patch adds support for the cgroups v2 unified filesystem. We now
attempt to detect at runtime which version of cgroup the host is using
and adapt appropriately.

Additionally this patch adds a small sanity check test for the profiler.
For now it only asserts that we have indeed received some profiling
sample data, in the future we can add more assertions.

Fixes #23

### DIFF
--- a/pkg/agent/profile_test.go
+++ b/pkg/agent/profile_test.go
@@ -1,0 +1,80 @@
+package agent
+
+import (
+	"context"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/containerd/cgroups"
+	"github.com/go-kit/log"
+	"github.com/parca-dev/parca-agent/pkg/debuginfo"
+	"github.com/parca-dev/parca-agent/pkg/ksym"
+)
+
+func TestCgroupProfiler(t *testing.T) {
+	var (
+		gotSample = make(chan struct{})
+
+		// Put this here specifically so we only close over the t and gotsample variables
+		// and don't bloat this closure with a bunch of variables.
+		sink = func(r Record) {
+			if len(r.Profile.Sample) == 0 {
+				t.Fatal("expected at least one sample")
+			}
+			gotSample <- struct{}{}
+		}
+
+		unit           = "upower.service" // Ensure we profile a noisy and ubiquitous process.
+		logger         = log.NewNopLogger()
+		ksymCache      = ksym.NewKsymCache(logger)
+		ctx            = context.Background()
+		duration       = time.Second * 10
+		errCh          = make(chan error)
+		externalLabels = map[string]string{"systemdunit": unit}
+	)
+
+	f, err := os.CreateTemp(os.TempDir(), "test.tmp")
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Seems backwards but since defers stack this will
+	// close first then remove.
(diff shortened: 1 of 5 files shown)


pkg/agent/systemdmanager.go
```diff
@@ -66,7 +69,17 @@ func (t *SystemdUnitTarget) Labels() []*profilestorepb.Label {
...
 
 func (t *SystemdUnitTarget) PerfEventCgroupPath() string {
-	return fmt.Sprintf("/sys/fs/cgroup/perf_event/system.slice/%s/", t.Name)
+	switch t.cgroupMode {
+	case cgroups.Legacy, cgroups.Hybrid:
+		return fmt.Sprintf("/sys/fs/cgroup/perf_event/system.slice/%s/", t.Name)
+	case cgroups.Unified:
+		return fmt.Sprintf("/sys/fs/cgroup/system.slice/%s/", t.Name)
+	case cgroups.Unavailable:
+		// TODO(derekparker): we should probably log this at least.
```
kakkoyun, line 78: :+1:

pkg/agent/profile_test.go
```diff
@@ -0,0 +1,80 @@
...
+		// Put this here specifically so we only close over the t and gotsample variables
+		// and don't bloat this closure with a bunch of variables.
+		sink = func(r Record) {
+			if len(r.Profile.Sample) == 0 {
+				t.Fatal("expected at least one sample")
+			}
+			gotSample <- struct{}{}
+		}
+
+		unit           = "upower.service" // Ensure we profile a noisy and ubiquitous process.
```
kakkoyun, line 28: :100:

kakkoyun (approved):
LGTM :1st_place_medal: 

The only missing piece is failing tests. I think we need to run tests as `sudo`:
`profile_test.go:73: new bpf module: error setting rlimit: operation not permitted`

derekparker:
Of course the CI server doesn't have `upower.service`...

```
=== RUN   TestCgroupProfiler
    profile_test.go:69: waiting for profiler to collect data
    profile_test.go:73: open cgroup: open /sys/fs/cgroup/perf_event/system.slice/upower.service/: no such file or directory
--- FAIL: TestCgroupProfiler (0.00s)
```

I'll look into this today and try to fix by Monday.

derekparker:
Alright so it likely is there it's just under `/sys/fs/cgroup/systemd/system.slice/`.

I'll dig into this more on Monday.

derekparker:
Closing this in favor of https://github.com/parca-dev/parca-agent/pull/178. 

That PR is simpler, just adding support for cgroupv2. It does not add a test because the last refactor made it a bit difficult to test the CgroupProfiler in isolation because it relies on the discoverer to resolve the systemd cgroup path, which I didn't want to reimplement in the test file itself. I think the better path forward will be to write integration tests since we need the discoverer and target manager running to setup the profilers correctly.

PR #177, opened 2021-12-20
Refactor service discovery mechanism in parca-agent

Sylfrena:
Currently, service discovery logic is contained in pkg/agent where
systemdmanager.go and podmanager.go individually discover systemd and
kubernetes targets respectively which are then called by the
target-manager.

This patch introduces a new Discoverer interface which is implemented
by the SystemdDiscoverer and PodDiscoverer types for fetching from
systemd and Kubernetes target sources respectively. Targets from these
providers are then pushed by the discovery-manager onto the target-manager
which reconciles the target groups.

Additionally, observer pattern logic for fetched profiles has been
moved from agent/targetmanager.go to agent/listener.go .

Co-developed-by: Frederic Branczyk <fbranczyk@gmail.com>
Co-developed-by: Sumera Priyadarsini <sylphrenadin@gmail.com>


cmd/parca-agent/main.go
```diff
@@ -102,71 +104,50 @@ func main() {
...
 
 	if len(flags.StoreAddress) > 0 {
-		conn, err := grpcConn(reg, flags)
+		_, err := grpcConn(reg, flags)
 		if err != nil {
 			level.Error(logger).Log("err", err)
 			os.Exit(1)
 		}
-
-		wc = profilestorepb.NewProfileStoreServiceClient(conn)
```
brancz, line 118: whops I think this was removed accidentally, without this we won't be sending anything to a Parca server

cmd/parca-agent/main.go
```diff
@@ -102,71 +104,50 @@ func main() {
...
-			dc,
-			flags.TempDir,
-			flags.ProfilingDuration,
-			flags.SystemdCgroupPath,
-		)
-		targetSources = append(targetSources, sm)
+		))
+	}
+
+	if flags.ExternalLabel == nil {
```
brancz, line 140: How about we move this external label handling to a separate function? The main is already quite busy maybe we can try to refactor everything we touch a bit.

cmd/parca-agent/main.go
```diff
@@ -276,13 +266,13 @@ func main() {
 
 			w.Header().Set("Content-Type", "application/vnd.google.protobuf+gzip")
 			w.Header().Set("Content-Disposition", "attachment;filename=profile.pb.gz")
-			err = profile.Write(w)
```
brancz, line 279: I think this was important for the download feature on the `Show Profile` page.

pkg/agent/listener.go
```diff
@@ -0,0 +1,127 @@
...
+func (l *ProfileListener) NextMatchingProfile(ctx context.Context, matchers []*labels.Matcher) (*profile.Profile, error) {
+	pCh := make(chan []byte)
+	defer close(pCh)
+
+	o := l.Observe(func(r *profilestorepb.WriteRawRequest) {
+
+		var searchedSeries *profilestorepb.RawProfileSeries
+
+		for _, series := range r.Series {
+			profileLabels := model.LabelSet{} //map[string]string{}
```
brancz, line 99: We can probably live without that comment.

It wouldn't be the worst to comment all of the functions and structs in this file, but we didn't really before so I wouldn't block this PR on that. We should take a note for the future though (or do it now if you feel like it).

pkg/agent/profile.go
```diff
@@ -62,7 +68,7 @@ type Record struct {
 
 type CgroupProfilingTarget interface {
```
brancz, line 69: I believe this interface is now unused, so I think we can safely remove it.

pkg/agent/listener.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+func (l *ProfileListener) WriteRaw(ctx context.Context, r *profilestorepb.WriteRawRequest, opts ...grpc.CallOption) (*profilestorepb.WriteRawResponse, error) {
+	l.ObserveProfile(r)
+	return l.next.WriteRaw(ctx, r, opts...)
+}
+
+type observer struct {
+	f func(*profilestorepb.WriteRawRequest)
+}
```
metalmatze, line 51: Verrrrry nice! :100:

pkg/discovery/systemd.go
```diff
@@ -0,0 +1,165 @@
...
+				Source:  unit,
+			})
+		}
+
+		up <- targetGroups
+	}
+}
+
+func (c *SystemdDiscoverer) ReconcileUnit(ctx context.Context, unit string) (model.LabelSet, error) {
+	f, err := os.Open(fmt.Sprintf("/sys/fs/cgroup/systemd/system.slice/%s/cgroup.procs", unit))
```
brancz, line 101: These paths will conflict quite a bit with what is done in https://github.com/parca-dev/parca-agent/pull/171

I don't think we should wait on either front for the other one, just something to keep in mind, but wouldn't hurt to move the templating to a separate function I think that will probably make things easier either way.

brancz (commented):
Super nice job!

pkg/agent/profile.go
```diff
@@ -145,31 +146,35 @@ func (p *CgroupProfiler) LastError() error {
...
-		})
-	for key, value := range p.externalLabels {
-		labels = append(labels, &profilestorepb.Label{
-			Name:  key,
-			Value: value,
-		})
+func (p *CgroupProfiler) Labels() model.LabelSet {
+	labels :=
+		model.LabelSet{
+			"__name__": "parca_agent_cpu",
```
metalmatze, line 160: I can see this hasn't been a constant before, but in a future Pull Request we can make this a constant too. :)
For the label there should also be `labels.MetricName` be available.

brancz: It's just coincidental that we chose `__name__` as well, using `labels.MetricName` would be confusing I think.

cmd/parca-agent/main.go
Sylfrena: right, good catch!

cmd/parca-agent/main.go
Sylfrena: oops, will add this back.

pkg/agent/listener.go
Sylfrena: Agreed, the comment was only for my reference.  Will remove it.

metalmatze (approved):

cmd/parca-agent/main.go
```diff
@@ -295,21 +292,46 @@ func main() {
 		})
 	}
 
-	if len(flags.SystemdUnits) > 0 {
+	// Run group for discovery manager
+	var m *discovery.Manager
+	var err error
+
+	{
```
derekparker, line 299: Just out of curiosity, why put this code (and that below) in it's own lexical block?

derekparker: Ah, I guess below it scopes `ctx, cancel := context.WithCancel(ctx)` so that makes sense. I don't think there's any such scoping here.

This is definitely just a small nit though, feel free to ignore, I was just curious!

pkg/discovery/discoverymanager.go
```diff
@@ -0,0 +1,321 @@
...
+			Name: "parca_agent_sd_updates_delayed_total",
+			Help: "Total number of update events that couldn't be sent immediately.",
+		},
+	)
+	sentUpdates = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Name: "parca_agent_sd_updates_total",
+			Help: "Total number of update events sent to the SD consumers.",
+		},
+	)
```
yeya24, line 59: Is it better to make these metrics non-global? Maybe better to put them under a struct and with promauto.

cmd/parca-agent/main.go
brancz: Specifically for run groups I kind of like this pattern, but it's certainly debatable. If we want to change that I think we should do that outside of this PR though, as is it's just conforming to the rest of the code.

brancz (approved):
Incredible work! Let's go ahead with this I'd say and we can do some of the non-architectural changes as follow-ups? I think the promauto suggestion from @yeya24 makes a lot of sense.

pkg/discovery/discoverymanager.go
Sylfrena: Thanks for the suggestion! Made a [PR](https://github.com/parca-dev/parca-agent/pull/182) for the same.

pkg/agent/profile.go
```diff
@@ -185,7 +181,7 @@ func (p *CgroupProfiler) Run(ctx context.Context) error {
 		return fmt.Errorf("load bpf object: %w", err)
 	}
 
-	cgroup, err := os.Open(p.target.PerfEventCgroupPath())
+	cgroup, err := os.Open(string(p.target[model.LabelName("__cgroup_path__")]))
```
kakkoyun, line 184: This gives me a weird feeling. I would rather prefer the previous version.

kakkoyun: At the very least we should have used the constant. `const CgroupPathLabelName = model.LabelName("__cgroup_path__")`

I still think this should be a method and the implementation details should be hidden under the hood.

cmd/parca-agent/main.go
```diff
@@ -116,57 +128,32 @@ func main() {
 		}
 
 		wc = profilestorepb.NewProfileStoreServiceClient(conn)
-		dc = parcadebuginfo.NewDebugInfoClient(conn)
```
kakkoyun, line 119: I think this is unintentionally removed. With the current state we only use noopClient for the debuginfo.

pkg/agent/profile.go
brancz: Feel free to fix this, I agree that at the very least we should use the constant. We did intentionally move to this being a label-set so over all I do think this is more consistent, but I agree with the use of the constant.

cmd/parca-agent/main.go
kakkoyun: Resolved by https://github.com/parca-dev/parca-agent/pull/188

PR #178, opened 2021-12-20
pkg/discovery: Support cgroupv2

### LINKED ISSUE #23: Add support for cgroup v2

derekparker:
With cgroup v2 the perf_events controller is implicitly enabled so we
don't have to do anything manually from the perspective of the
parca-agent.

Fixes #23

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/alecthomas/kong v0.2.22
 	github.com/aquasecurity/libbpfgo v0.2.2-libbpf-0.5.0
 	github.com/cespare/xxhash/v2 v2.1.2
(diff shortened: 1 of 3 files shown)


derekparker:
We should also handle hybrid mode in a follow up PR. I think we can see is `/sys/fs/cgroup/perf_event` exists in which case we ignore the `unified` directory since it's not recommended to mix and match cgroup v1 and v2.

brancz (approved):

brancz:
Nice!

PR #181, opened 2021-12-21
pkg/agent: Use eBPF map batch operations

### LINKED ISSUE #74: Use batch operations
https://github.com/parca-dev/parca-agent/blob/32327eae805c18a3483731cb3dea5fff697dbcb1/pkg/agent/profile.go#L280

libbpf supports the batch operations such as:

* `bpf_map_lookup_and_delete_batch`
* `bpf_map_lookup_batch`
* `bpf_map_delete_batch`

These make sense to be used. Because we already load all the data from the BPF maps into the memory of our user-space program, so might as well do it efficiently. And not only does it allow us to batch, but it also allows us to delete items at the same time, which is exactly what we do anyways after each profiling iteration.

They first need to be implemented in [libbpfgo](https://github.com/aquasecurity/libbpfgo).

derekparker:
This patch adds support for using eBPF batch operations on maps which
should dramatically improve the performance of the parca-agent by
enabling it to batch lookup and delete keys/values in maps in a single
operation.

This replaces the use of iterators in favor of iteration in batches.

Fixes #74

### DIFF
--- a/pkg/profiler/profiler.go
+++ b/pkg/profiler/profiler.go
@@ -52,7 +52,6 @@ var bpfObj []byte
 const (
 	stackDepth       = 127 // Always needs to be sync with MAX_STACK_DEPTH in parca-agent.bpf.c
 	doubleStackDepth = 254
-	batchSize        = 1024
 )
 
 // stackCountKey mirrors the struct in parca-agent.bpf.c.
@@ -315,7 +314,9 @@ func (p *CgroupProfiler) profileLoop(ctx context.Context, now time.Time, counts 
 			// TODO(derekparker): Should we log or increment missing stack trace count?
 			continue
 		}
-		stackTraces.DeleteKey(unsafe.Pointer(&userStackID))
+		if err := stackTraces.DeleteKey(unsafe.Pointer(&userStackID)); err != nil {
+			return fmt.Errorf("unable to delete stack trace key: %w", err)
+		}
 
 		// Twice the stack depth because we have a user and a potential Kernel stack.
 		stack := [doubleStackDepth]uint64{}
@@ -330,7 +331,9 @@ func (p *CgroupProfiler) profileLoop(ctx context.Context, now time.Time, counts 
 				//profile.MissingStacks++
 				continue
 			}
-			stackTraces.DeleteKey(unsafe.Pointer(&kernelStackID))
+			if err := stackTraces.DeleteKey(unsafe.Pointer(&kernelStackID)); err != nil {
+				return fmt.Errorf("unable to delete stack trace key: %w", err)
+			}
 
 			err = binary.Read(bytes.NewBuffer(stackBytes), byteOrder, stack[stackDepth:])
 			if err != nil {
--- a/pkg/profiler/profiler.go
+++ b/pkg/profiler/profiler.go
@@ -18,12 +18,13 @@ import (
 	"context"
 	_ "embed"
 	"encoding/binary"
+	"errors"
 	"fmt"
-	"io"
 	"os"
 	"runtime"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -93,7 +93,7 @@ Flags:
                                   socket. Leave this empty to use the defaults.
       --profiling-duration=10s    The agent profiling duration to use. Leave
                                   this empty to use the defaults.
-      --systemd-cgroup-path="/sys/fs/cgroup/systemd/system.slice"
+      --systemd-cgroup-path=STRING
```
kakkoyun, line 96: I think this is a left-over change from the other PR #206. Let's keep things clean.

cmd/parca-agent/main.go
```diff
@@ -79,7 +79,7 @@ type flags struct {
 	TempDir            string            `kong:"help='Temporary directory path to use for object files.',default='/tmp'"`
 	SocketPath         string            `kong:"help='The filesystem path to the container runtimes socket. Leave this empty to use the defaults.'"`
 	ProfilingDuration  time.Duration     `kong:"help='The agent profiling duration to use. Leave this empty to use the defaults.',default='10s'"`
-	SystemdCgroupPath  string            `kong:"help='The cgroupfs path to a systemd slice.',default='/sys/fs/cgroup/systemd/system.slice'"`
+	SystemdCgroupPath  string            `kong:"help='The cgroupfs path to a systemd slice.'"`
```
kakkoyun, line 82: ditto

pkg/profiler/profiler.go
```diff
@@ -53,6 +54,16 @@ const (
 	doubleStackDepth = 254
 )
 
+// stackCountKey mirrors the struct in parca-agent.bpf.c.
+//
+// TODO(derekparker) Perhaps in order to keep these in sync we should write a Go generator to
```
kakkoyun, line 59: Awesome idea! Let's do it.

kakkoyun: Already the opened the issue for it https://github.com/parca-dev/parca-agent/issues/207

pkg/profiler/profiler.go
```diff
@@ -229,80 +249,74 @@ func (p *CgroupProfiler) Run(ctx context.Context) error {
...
-		if err != nil {
-			return fmt.Errorf("get count value: %w", err)
-		}
-		value := byteOrder.Uint64(valueBytes)
+		value := byteOrder.Uint64(vals[i])
 
 		stackBytes, err := stackTraces.GetValue(unsafe.Pointer(&userStackID))
 		if err != nil {
-			//profile.MissingStacks++
+			// TODO(derekparker): Should we log or increment missing stack trace count?
```
kakkoyun, line 314: :+1: Let's add a metric for it, not necessarily in this PR though. https://github.com/parca-dev/parca-agent/issues/208

pkg/profiler/profiler.go
```diff
@@ -484,51 +498,18 @@ func (p *CgroupProfiler) profileLoop(ctx context.Context, now time.Time, counts,
...
+	return nil
+}
 
-		key := it.Key()
-		prev = make([]byte, len(key))
-		copy(prev, key)
+// memsetCountKeys will reset the given slice to the given value.
+// This function makes use of the highly optimized copy builtin function
+// and is able to fill the entire slice in O(log n) time.
+func memsetCountKeys(in []stackCountKey, v stackCountKey) {
```
kakkoyun, line 507: :heart:

kakkoyun (approved):
It looks good to me.

We just need to clean up a minor thing.

kakkoyun (changes requested):
Let's clean up first.

derekparker:
> Let's clean up first.

Done!

kakkoyun (approved):

PR #189, opened 2022-01-05
debug-info: Make sure it is safe to remove the section

kakkoyun:

With this PR we check all the debug sections and only remove them if they do not have any bits. This clean up was introduced in https://github.com/parca-dev/parca-agent/pull/102 for another special case for Go binaries where we had  `.debug_gdb_scripts` section without any bits.

I hope the PR will be a solution for both.

Fixes https://github.com/parca-dev/parca-agent/issues/185

cc @derekparker

### DIFF
--- a/pkg/debuginfo/debuginfo.go
+++ b/pkg/debuginfo/debuginfo.go
@@ -243,39 +243,26 @@ func (di *Extractor) extract(ctx context.Context, buildID string, file string) (
 		level.Debug(di.logger).Log("msg", "failed to determine if binary is a Go binary", "path", file, "err", err)
 	}
 
-	sections, err := sectionsToRemove(file)
+	toRemove, err := sectionsToRemove(file)
 	if err != nil {
 		level.Debug(di.logger).Log("msg", "failed to determine sections to remove", "path", file, "err", err)
 	}
 
-	var (
-		cmd           *exec.Cmd
-		debugInfoFile string
-		cleanup       func() error
-	)
+	var debugInfoFile string
 	switch {
 	case hasDWARF:
-		cmd, debugInfoFile, cleanup = di.useStrip(ctx, tmpDir, file, sections)
+		debugInfoFile, err = di.useStrip(ctx, tmpDir, file, toRemove)
 	case isGo:
-		cmd, debugInfoFile, cleanup = di.useObjcopy(ctx, tmpDir, file, sections)
+		debugInfoFile, err = di.useObjcopy(ctx, tmpDir, file, toRemove)
 	default:
-		cmd, debugInfoFile, cleanup = di.useStrip(ctx, tmpDir, file, sections)
+		debugInfoFile, err = di.useStrip(ctx, tmpDir, file, toRemove)
 	}
-	defer func() {
-		if err := cleanup(); err != nil {
-			level.Debug(di.logger).Log("msg", "failed to clean up", "err", err)
-		}
-	}()
-
-	if out, err := cmd.CombinedOutput(); err != nil {
-		level.Error(di.logger).Log(
-			"msg", "external binutils command call failed",
-			"output", strings.ReplaceAll(string(out), "\n", ""),
-			"file", file,
-		)
-		return "", fmt.Errorf("failed to extract debug information from binary: %w", err)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


brancz:
Happy with this being a follow-up, but could we have a test for this, maybe with an actual Rust binary?

brancz:
On second thought I think I might prefer a real e2e test where we add containers to a kubernetes cluster of all kinds of languages.

kakkoyun:
> On second thought I think I might prefer a real e2e test where we add containers to a kubernetes cluster of all kinds of languages.

I was going to suggest exactly this. We need a more flexible and extensible solution. Our support matrix will ever grow. I'm going to update the concerning issue with more context.

pkg/debuginfo/debuginfo.go
```diff
@@ -243,18 +243,30 @@ func (di *Extractor) extract(ctx context.Context, buildID string, file string) (
 		level.Debug(di.logger).Log("msg", "failed to determine if binary is a Go binary", "path", file, "err", err)
 	}
 
+	sections, err := sectionsToRemove(file)
```
derekparker, line 246: nit: I would rename this to something more like `removeable` or `toRemove`.

pkg/debuginfo/debuginfo.go
```diff
@@ -243,18 +243,30 @@ func (di *Extractor) extract(ctx context.Context, buildID string, file string) (
...
-		cmd, debugInfoFile = di.useStrip(ctx, tmpDir, file)
+		cmd, debugInfoFile, cleanup = di.useStrip(ctx, tmpDir, file, sections)
 	case isGo:
-		cmd, debugInfoFile = di.useObjcopy(ctx, tmpDir, file)
+		cmd, debugInfoFile, cleanup = di.useObjcopy(ctx, tmpDir, file, sections)
 	default:
-		cmd, debugInfoFile = di.useStrip(ctx, tmpDir, file)
+		cmd, debugInfoFile, cleanup = di.useStrip(ctx, tmpDir, file, sections)
 	}
+	defer func() {
```
derekparker, line 264: Why defer the cleanup?

derekparker: Also, why return a function at all? It seems that `useStrip` is the only method returning a cleanup function that does anything. Why not leave that defer in `useStrip` which cleans the interim file itself?

pkg/debuginfo/debuginfo.go
```diff
@@ -347,49 +359,49 @@ func hasDebugInfo(path string) (bool, error) {
...
-		case strings.HasPrefix(s.Name, ".debug_"):
-			return s.Name[7:]
-		case strings.HasPrefix(s.Name, ".zdebug_"):
-			return s.Name[8:]
-		case strings.HasPrefix(s.Name, "__debug_"): // macos
-			return s.Name[8:]
-		default:
-			return ""
-		}
+var dwarfSuffix = func(s *elf.Section) string {
```
derekparker, line 376: nit: I'm being incredibly pedantic here but this should be `dwarfPrefix` as opposed to `dwarfSuffix`.

derekparker (commented):
Just a few nits / questions.

pkg/debuginfo/debuginfo.go
kakkoyun: This is actually, removing the detected prefix and returning the suffix. Thus the name. Maybe I should a comment to clarify more.

derekparker: Aah, ok I understand now. Sorry for the noise!

pkg/debuginfo/debuginfo.go
derekparker: Ack, ok I see now, that makes sense.

However, is there any reason to not simply execute the command in the function we call and return the output directly? It seems like the caller (`di.extract` in this case) is only concerned with the output and doesn't need a handle to the `*exec.Command` pointer.

kakkoyun: I think this we can do :+1: Thanks for the suggestion.

kakkoyun:
I have applied the suggestion @derekparker. PTAL.

derekparker (approved):
LGTM

Issue #184, opened 2021-12-31
Parca agent makes Oracle Kubernetes Engine worker nodes down


tonychoe:
The subject may sound weird, but that's what I have experienced. Hope you guys shed some light. :) 

I started testing Parca in OKE (Oracle Kubernetes Engine) and found installing Parca agent made the worker node unresponsive to SSH and must be hard-rebooted. At first, I suspected the issue is with OKE or its underlying cloud infrastructure. But over time, I was able to isolate that this issue only happens after Parca agent is installed. The reproduction was done across multiple clusters, node pools, nodes, etc. 

But I still don't know what exactly caused this issue. I've scanned the system stats and logs but they looked clean. The system resources were fine. I could have missed something as I have no idea where to look at.

Parca agent showed this log from all nodes it was installed, but I am not sure if it is relevant (anyway, the log says "please report").
```
level=error ts=2021-12-31T05:29:33.514162476Z caller=debuginfo.go:259 namespace=kube-system pod=coredns-94d6cc8b6-br22d container=coredns component=debuginfoextractor msg="external binutils command call failed" output="objcopy: BFD (GNU Binutils for Debian) 2.31.1 internal error, aborting at ../../bfd/elf.c:7085 in rewrite_elf_program_headerobjcopy: Please report this bug." file=/proc/16763/root/coredns
```

Environment
- Oracle Kubernetes Engine (Kubernetes 1.21)
- OS: Oracle Linux Server 7.9 (a cousin to RHEL 7.9)

Steps to reproduce
1. kubectl create namespace parca
2. kubectl apply -f kubectl apply -f https://github.com/parca-dev/parca-agent/releases/download/v0.3.0/kubernetes-manifest.yaml
3. Watch the node become "not ready" and unresponsive within minutes. Sometimes in a few mins, sometimes takes a long time (so the nodes in the cluster which Parca agent is installed as DaemonSet went unresponsive one by one).
4. hard-reboot the node

brancz:
cc @kakkoyun

kakkoyun:
xref: https://github.com/parca-dev/parca/issues/514

kakkoyun:
@tonychoe Thanks for reporting the issue. I'm looking into this.
I would be glad if you could share more information about your current setup.

Since the log is about an external program that we use to extract debug information on the host, have you seen any zombie processes before the node went down? It's just a wild guess, but we might be leaking file descriptors.

That being said, I'm looking into this in more detail.

[...]

PR #198, opened 2022-01-06
debug-info: Use child process reaper for more reliable external command execution

kakkoyun:

Used https://github.com/containerd/containerd/blob/main/sys/reaper/reaper_unix.go

Could be a mitigation for https://github.com/parca-dev/parca-agent/issues/184

cc @derekparker

### DIFF
--- a/cmd/parca-agent/main.go
+++ b/cmd/parca-agent/main.go
@@ -23,11 +23,13 @@ import (
 	"net/http/pprof"
 	"net/url"
 	"os"
+	"os/signal"
 	"sort"
 	"strings"
 	"time"
 
 	"github.com/alecthomas/kong"
+	"github.com/containerd/containerd/sys/reaper"
 	"github.com/go-kit/log/level"
 	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
 	"github.com/oklog/run"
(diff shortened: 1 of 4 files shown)


brancz:
If I'm understanding this correctly the two things this does is:

1) Continuously reap zombie processes while Parca Agent continues to run.
2) Reap all processes when Parca Agent shuts down.

If that's correct, then lgtm :+1:

kakkoyun:
> If I'm understanding this correctly the two things this does is:
> 
> 1. Continuously reap zombie processes while Parca Agent continues to run.

Yes :+1: 

> 2. Reap all processes when Parca Agent shuts down.
I'll just add the necessary line :)

derekparker:
Hm, I'd be kind of be surprised if we're not reaping these processes already. We run them with `cmd.CombinedOutput()` which internally uses `cmd.Run` which also internally calls `cmd.Wait()` which should wait until the process has actually exited so the `Wait` call itself should reap the child.

Have we observed these commands creating zombie processes in the wild? I saw the associated issue but I'd be surprised if executing these commands causes pid exhaustion or something else to lock up the system. Difficult to say without more information though. We should ask the reporter of the original issue to test this patch if possible.

kakkoyun:
@derekparker Exactly as you said. We don't for sure. This is just mitigation for a hunch, in case we're exhausting PID space. We make sure we clean as much as possible. This doesn't have a major downside if you don't have major concerns I will go with it. Until we hear back from them

brancz:
We sadly don't push container images on PRs/branches, so I would be ok with merging to make it as easy as possible for the user to try.

derekparker:
> @derekparker Exactly as you said. We don't for sure. This is just mitigation for a hunch, in case we're exhausting PID space. We make sure we clean as much as possible. This doesn't have a major downside if you don't have major concerns I will go with it. Until we hear back from them

SGTM!

PR #199, opened 2022-01-06
pkg/discovery: take user provided systemd cgroup path into account

### LINKED ISSUE #192: Multiple systemd units with the same name
I have just started playing with `parca` and this is the first thing that came up. ;)

Use case - multiple containers with systemd inside:
Cgroup v2 structure:
```
/sys/fs/cgroup/machine.slice/machine-libpod_pod_<some_hash>.slice/libpod-<some_hash>.scope/container/system.slice
/sys/fs/cgroup/machine.slice/machine-libpod_pod_<other_hash>.slice/libpod-<other_hash>.scope/container/system.slice
/sys/fs/cgroup/machine.slice/machine-libpod_pod_<antother_hash>.slice/libpod-<another_hash>.scope/container/system.slice
```
Each of the containers runs systemd unit with the same name. Obviously, from the container perspective those are separate systemd instances, but from the host perspective when using single `parca-agent` there is a problem, I can't just use `--systemd-units=my-app.service` - `parca-agent` will try to find that service under `/sys/fs/cgroup/system.slice` (BTW default for now is `/sys/fs/cgroup/systemd/system.slice` which is wrong for cgroup v2. Shouldn't this be autodetected?)

Can this be supported, or should I run multiple agents for each container with `--systemd-cgroup-path=` pointing at each `system.slice` inside `machine.slice`?

Env: RHEL 8 and podman.

derekparker:
This patch wires up the --systemd-cgroup-path flag and actually plumbs
the user provided value into the systemd config and subsequently into
the systemd discoverer to be used when trying to reconcile unit files.

Fixes #192

### DIFF
--- a/cmd/parca-agent/main.go
+++ b/cmd/parca-agent/main.go
@@ -152,6 +152,7 @@ func main() {
 	if len(flags.SystemdUnits) > 0 {
 		configs = append(configs, discovery.NewSystemdConfig(
 			flags.SystemdUnits,
+			flags.SystemdCgroupPath,
 		))
 	}
 
--- a/pkg/discovery/systemd.go
+++ b/pkg/discovery/systemd.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path"
 	"sync"
 	"syscall"
 	"time"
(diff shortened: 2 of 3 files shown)


kakkoyun (approved):
LGMT :rocket:

PR #206, opened 2022-01-13
Remove default systemd cgroup path

derekparker:
There was a previous commit which wired up this flag throughout the
codebase, however that code didn't take into account that this flag
has a default parameter. Remove the default and let the user set it
if it's really necessary.


kakkoyun:
Related https://github.com/parca-dev/parca-agent/pull/199

kakkoyun (approved):
LGTM :rocket: 

Amazing, thanks for taking care of this!

PR #210, opened 2022-01-18
pkg/containerutils/containerd: Automatically discover k3s containerd

brancz:
Previously the k3s containerd socket had to be explicitly configured,
however, similar to the standard Kubernetes containerd configuration k3s
has a (different) standard location as well. In order to not force users
to make special changes for k3s, this patch automatically attempts
discovering the k3s path as well.


kakkoyun (approved):
LGTM

PR #225, opened 2022-01-25
Small fixes in Makefile/Dockerfile

mrueg:
* Reduce packages fetched by --no-install-recommends
* Replace cd && make with make -C

Looking into multi-arch containers, which might be a bit more tricky for this project, so I push these smaller changes early.


brancz (approved):
Nice!! Thank you!! lgtm :+1:

PR #252, opened 2022-02-22
Add support for Position Independent Code

### LINKED ISSUE #256: Cache debuginfo.Exists responses
As far as I'm aware the requests send with `debuginfo.Exists` are to check if a binary symbol is already present on the Parca server side. 
Given that the build IDs are pretty unique I propose to add caching on Parca agent's side to reduce the amount of Exists requests sent.

kakkoyun:

Fixes https://github.com/parca-dev/parca/issues/326

also fixes #256 and #264

* Add support for PIC
* Refactor debuginfo package
* Fix a couple of linter errors


pkg/debuginfo/cache.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/parca-dev/parca-agent/pkg/objectfile"
+)
+
+type cache struct {
+	files map[string]*debugInfoFile
+}
+
+func newCache() *cache {
+	return &cache{
+		files: make(map[string]*debugInfoFile),
```
brancz, line 26: we're going to want this to be a cache that actually evicts, I'd say let's start with an lru cache

pkg/debuginfo/cache.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+}
+
+func (c *cache) debugInfoFile(objFile *objectfile.ObjectFile) *debugInfoFile {
```
brancz, line 30: it's quite confusing to me as the reader that this function is both used to read and write from the cache, I think making this a read-through cache would be easier to understand for the user, maybe even wrapping the gRPC client, so the caching logic doesn't spill into the business logic

pkg/debuginfo/debuginfo_file.go
```diff
@@ -0,0 +1,121 @@
...
+	dbgOnce      sync.Once
+	dbgErr       error
+	hasDebugInfo bool
+
+	localDbgOnce       sync.Once
+	localDbgErr        error
+	localDebugInfoPath string
+}
+
+func (f *debugInfoFile) LocalHostDebugInfoPath() (string, error) {
```
brancz, line 54: can we rename this to just `LocalDebugInfoPath`? `LocalHost` strongly suggested to me as the reader that what we're dealing here has something to do with the hostname

pkg/debuginfo/debuginfo_file.go
```diff
@@ -0,0 +1,121 @@
...
+	f.localDbgOnce.Do(func() {
```
brancz, line 55: the `sync.Once` only holds true for this instance of the `debugInfoFile` struct, so doing it once at instantiation is equally good, I think this makes it pretty confusing to read

pkg/objectfile/object_file.go
```diff
@@ -0,0 +1,270 @@
...
+
+func (f *ObjectFile) FullPath() string {
+	if f.path != "" {
+		return f.path
+	}
+	return path.Join(f.Root(), f.File)
+}
+
+func (f *ObjectFile) ObjAddr(addr uint64) (uint64, error) {
+	f.baseOnce.Do(func() { f.baseErr = f.computeBase(addr) })
```
brancz, line 184: we already have the `NewObjectFile` function that we use to instantiate the struct, let's just do the computation there once

pkg/maps/mapping.go
```diff
@@ -49,31 +48,16 @@ func (m *Mapping) PidAddrMapping(pid uint32, addr uint64) (*profile.Mapping, err
...
 	for _, pid := range m.pids {
 		maps := m.pidMappings[pid]
 		for _, mapping := range maps {
 			if mapping.BuildID != "" {
-				buildIDFiles[mapping.BuildID] = BuildIDFile{
-					PID:  pid,
-					File: mapping.File,
-				}
+				// TODO(kakkoyun): Use objectfile.FromProcess() and add cache!
+				objFiles = append(objFiles, objectfile.NewObjectFile(pid, mapping))
```
brancz, line 60: Still todo?

pkg/objectfile/object_file.go
```diff
@@ -0,0 +1,270 @@
...
+		},
+	}, nil
+}
+
+func NewObjectFile(pid uint32, m *profile.Mapping) *ObjectFile {
+	return &ObjectFile{
+		PID:     pid,
+		File:    m.File,
+		BuildID: m.BuildID,
+		m: &mapping{
```
brancz, line 148: the `kernelOffset` is not set here, is that right?

pkg/profiler/profiler.go
```diff
@@ -215,36 +276,36 @@ func (p *CgroupProfiler) Run(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("get stack traces map: %w", err)
 	}
+	p.bpfMaps = &bpfMaps{counts, stackTraces}
```
brancz, line 279: can we use field names here? I was confused and had to look at the `bpfMaps` type to understand what's going on here

pkg/profiler/profiler.go
```diff
@@ -257,6 +318,7 @@ func (p *CgroupProfiler) profileLoop(ctx context.Context, now time.Time, counts,
 
 	mapping := maps.NewMapping(p.pidMappingFileCache)
 	kernelMapping := &profile.Mapping{
+		// TODO(kakkoyun): Check if this conflicts with https://github.com/google/pprof/pull/675/files
```
brancz, line 321: nice find!

pkg/debuginfo/cache.go
kakkoyun: This is actually a read-through cache. The idea is to reduce the redundant work of find the debug information and maybe extraction. So it's not directly related to gRPC calls. This is one layer above. I realized I can do a better job of explaining this and possibly improve the interface. I'll do so.

pkg/objectfile/object_file.go
kakkoyun: In order to compute the base properly we need an address, that's why we need to defer the calculation until we received an address from the profiler.

kakkoyun:
There is a regression, in the branch. I'm trying to find it. Thus, I've marked it as draft again.

brancz (approved):

pkg/profiler/profiler.go
```diff
@@ -132,7 +198,7 @@ func (p *CgroupProfiler) Stop() {
 	p.mtx.Lock()
 	defer p.mtx.Unlock()
 	level.Debug(p.logger).Log("msg", "stopping cgroup profiler")
-	if !prometheus.Unregister(p.missingStacks) {
+	if !p.reg.Unregister(p.missingStacks) {
```
kakkoyun, line 201: cc @Sylfrena

PR #266, opened 2022-03-02
pkg/maps: Skip more pseudo-paths and anonymous mappings 

mrueg:
This should prevent trying to access the root symlink because the m.File is empty.
```
level=warn ts=2022-03-02T15:50:14.508461455Z caller=maps.go:102 msg="failed to read obj build ID" obj=/proc/12497/root err="failed to open elf: read /proc/12497/root: is a directory"
``` 

(Hopefully)
Fixes: #82

### DIFF
--- a/pkg/maps/maps.go
+++ b/pkg/maps/maps.go
@@ -91,8 +91,8 @@ func (c *PidMappingFileCache) mappingForPid(pid uint32) ([]*profile.Mapping, err
 	for _, m := range mapping {
 		// Try our best to have the BuildID.
 		if m.BuildID == "" {
-			// TODO(brancz): These need special cases.
-			if m.File == "[vdso]" || m.File == "[vsyscall]" {
+			// TODO(brancz): These need special cases. See pseudo-paths in proc's man page
+			if m.File == "[vdso]" || m.File == "[vsyscall]" || m.File == "[stack]" || m.File == "[heap]" || m.File == "" {
 				continue
 			}
 


kakkoyun (approved):
LGTM.

Funny enough, I did a similar change on another WIP PR.

mrueg:
Ideally there should be a check for the existence of the file, as there are more possible pseudopaths.

See `man proc`

```
There are additional helpful pseudo-paths:

              [stack]
                     The initial process's (also known as the main
                     thread's) stack.

              [stack:<tid>] (from Linux 3.4 to 4.4)
                     A thread's stack (where the <tid> is a thread ID).
                     It corresponds to the /proc/[pid]/task/[tid]/ path.
                     This field was removed in Linux 4.5, since
                     providing this information for a process with large
                     numbers of threads is expensive.

              [vdso] The virtual dynamically linked shared object.  See
                     [vdso(7)](https://man7.org/linux/man-pages/man7/vdso.7.html).

              [heap] The process's heap.

              If the pathname field is blank, this is an anonymous
              mapping as obtained via [mmap(2)](https://man7.org/linux/man-pages/man2/mmap.2.html).  There is no easy way to
              coordinate this back to a process's source, short of
              running it through [gdb(1)](https://man7.org/linux/man-pages/man1/gdb.1.html), [strace(1)](https://man7.org/linux/man-pages/man1/strace.1.html), or similar.
```

mrueg:
@kakkoyun sounds good, I added the other pseudo paths as well to reduce misleading log message.

kakkoyun (approved):
LGTM

PR #269, opened 2022-03-03
discovery: Support systems without systemd

### LINKED ISSUE #268: Incorrect cgrouppath for a pod
# Environment
-  `parca-agent` `0.5.0`
- `kernel` `5.10.61`
- `kubernetes` `1.21` with both cgroupv1 and cgroupv2 mounted
- `containerd` `1.4.9`

# What's happening
When looking into a pod using parca, we noticed that multiple processes were shown in the profile that are unrelated to each other.

Screenshot 2022-03-02 at 15 56 57

This is the target we used to see this result: http://localhost:7071/query?debug=1&query=%7Bcontainer%3D%22envoy-app%22%2C+containerid%3D%22docker%3A%2F%2F65e940dca4baa227bb0b3c1a91226c9717c0fbb5071df1c9f3ba04ab3a39152c%22%2C+namespace%3D%22app%22%2C+node%3D%22kube-worker2.x.y.z%22%2C+pod%3D%22app-864f4b66d6-gqfc2%22%7D

Digging a a bit more into it with @brancz, we noticed that `__cgroup_path__` was set to `/sys/fs/cgroup/perf_event` to all of them. This value is set in https://github.com/parca-dev/parca-agent/blob/1110c48ae92ff9c1a3eee622c278dae016a72173/pkg/k8s/k8s.go#L172-L188 and the struct always has the same values when called:
- `c.CgroupV1=""`
- `c.CgroupV2="/docker"`
- `c.Cgrouppath="/sys/fs/cgroup/unified/docker"`

Since this node has both cgroupv1 and cgroupv2 mounted, it might be related to #180.

# What's expected
The correct Cgrouppath is set and profiles only contain mappings for the requested pod.

xiu:
Before this commit, we only supported discovering cgroups via systemd by
reading /proc/$PID/cgroup:
```
[...]
1:name=systemd:/kubepods/burstable/pod1fa4240d-10d4-494f-88d2-80a639e26ac5/de86e0bfc1f648084db99ead71c756df43ec7e936342144c61d58961859a564b
[...]
```

On systems running openrc for instance, that line doesn't exist. Instead
we have:
```
1:name=openrc:/docker
```

With this, parca-agent would discover the __cgroup_path__ for the job to
be `/sys/fs/cgroup/perf_event`, leading to parca-agent profiling
everything for every pod.

This commit adds logic to support discovering the cgroup via perf_event.
If the system is running systemd, it'll always fall back to using that
line as it is the second from the last.

Fixes #268

### DIFF
--- a/pkg/containerutils/containerutils.go
+++ b/pkg/containerutils/containerutils.go
@@ -114,6 +114,12 @@ func GetCgroupPaths(pid int) (string, string, error) {
 			if err != nil {
 				break
 			}
+			// Fallback in case the system the agent is running on doesn't run systemd
+			if strings.Contains(line, ":perf_event:") {
+				cgroupPathV1 = strings.SplitN(line, ":", 3)[2]
+				cgroupPathV1 = strings.TrimSuffix(cgroupPathV1, "\n")
+				continue
+			}
 			if strings.HasPrefix(line, "1:name=systemd:") {
 				cgroupPathV1 = strings.TrimPrefix(line, "1:name=systemd:")
 				cgroupPathV1 = strings.TrimSuffix(cgroupPathV1, "\n")


brancz:
Awesome! Thank you so much for having a look into this! lgtm 🎉 

cc @kakkoyun any comments from your side?

kakkoyun (approved):
LGTM :rocket: 

Awesome job :100:

xiu:
Thanks folks!

PR #272, opened 2022-03-03
Upgrade libbpf to 7.0.0

kakkoyun:

### DIFF
--- a/deploy/tilt/parca-server-deployment.yaml
+++ b/deploy/tilt/parca-server-deployment.yaml
@@ -5,7 +5,7 @@ metadata:
     app.kubernetes.io/component: observability
     app.kubernetes.io/instance: parca
     app.kubernetes.io/name: parca
-    app.kubernetes.io/version: v0.8.0
+    app.kubernetes.io/version: v0.8.1
   name: parca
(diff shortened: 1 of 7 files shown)


brancz (approved):

mrueg:
Looks like with this change a couple of bpf functions and maps got deprecated:
 
```
libbpf: map 'counts' (legacy): legacy map definitions are deprecated, use BTF-defined maps instead

libbpf: map 'stack_traces' (legacy): legacy map definitions are deprecated, use BTF-defined maps instead

libbpf: Use of BPF_ANNOTATE_KV_PAIR is deprecated, use BTF-defined maps in .maps section instead
```

kakkoyun:
Hmmm :thinking: I will check this out. I have run it and haven't caught these. Agent seemed functional.

I'm going to take care of a couple of other things before releasing the next version of agent around libbpf.

If we migrate to BTF-defined maps, we need to bump the oldest supported kernel version to 5.2. It's not a blocker, for upcoming features we have to do this anyway.

Thanks for pointing it out @mrueg 

Please let me know if you observe any functional regressions.

Issue #275, opened 2022-03-03
Time-based invalidation of debug information caches


kakkoyun:

brancz:
Why time-based? I've always found size- or count-based to be more predictable for users in practice.

kakkoyun:
I should have more specific. I specifically meant debug info exists response cache. Now that we have another way to find debug information, it would be more correct to invalidate local cache.

brancz:
Ah, that makes perfect sense, yes! We probably have to also differentiate between the result of the response.

PR #282, opened 2022-03-15
Add TTL to exists response cache

kakkoyun:

Fixes https://github.com/parca-dev/parca-agent/issues/275

Depends https://github.com/parca-dev/parca/pull/767


brancz:
Looks good, just for clarification though, these are LRU caches not time-based, right?

kakkoyun:
> Looks good, just for clarification though, these are LRU caches not time-based, right?

Unless an option is given e. g. `cache.WithExpireAfterAccess(2*time.Minute)`, they are size-based LRU caches.

https://pkg.go.dev/github.com/goburrow/cache#section-readme

I have changed the existing one that don't need time-based invalidation just to eliminate additional dependency.

brancz:
Got it, and since we're not using that, these are "just" LRU caches. I was just thrown off by the title and commit.

Love it!

PR #299, opened 2022-03-21
pkg/target: Add functionality for using specified sampling ratio

### LINKED ISSUE #286: Utilize SamplingRatio flag
It seems like we accidentally remove the ability to specify sampling ratio. Let's bring it back.

cc @Sylfrena @metalmatze

Sylfrena:
Fixes #286

### DIFF
--- a/cmd/parca-agent/main.go
+++ b/cmd/parca-agent/main.go
@@ -165,6 +165,7 @@ func main() {
 		profileListener, debugInfoClient,
 		flags.ProfilingDuration,
 		externalLabels(flags.ExternalLabel, flags.Node),
+		flags.SamplingRatio,
 		flags.TempDir,
 	)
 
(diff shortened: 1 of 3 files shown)


kakkoyun:
I've always imagined adjusting the frequency or limit of the events we collected from the Kernel. But now I'm confused, @brancz what was the intended behavior when sampling ratio flag is specified? The existing flag description didn't make it clear for me :sweat_smile:

Sylfrena:
@kakkoyun, afaiu, the intended behaviour is still adjusting the number of profiles, except instead of doing this while collecting the events, this is done right before profiling the samples, so only the specified fraction of samples is profiled instead. I should have been clearer in the commit message.

kakkoyun:
I think I'm gonna pass the decision on this one to @brancz, because lack of context on my side.

brancz:
This flag was originally to control the targets that would be sampled, similar to the label selector for pods. This is likely to change with the host-wide profiler architecture but until then we should make the flag work as intended (for now).

PR #318, opened 2022-03-29
discovery: Sanitize Kubernetes label names

brancz:
Take two ...

Turns out in #318 I incorrectly sanitized the label-values not the label-names. Label values can be anything, they do not have to be sanitized.


kakkoyun (approved):
:100:

PR #321, opened 2022-03-30
debuginfo/find: Improve separate debug file discovery

kakkoyun:
With this PR, parca-agent adopts the same behavior as GDB to discovery separate debug files.


pkg/debuginfo/find.go
```diff
@@ -144,18 +151,52 @@ func (f *Finder) find(root, buildID, path string) (string, error) {
...
-func (f *Finder) generatePaths(root, buildID, path string) []string {
+func readDebuglink(path string) (string, error) {
+	// A debug link is a special section of the executable file named .gnu_debuglink. The section must contain:
+	//
+	// A filename, with any leading directory components removed, followed by a zero byte,
+	//  zero to three bytes of padding, as needed to reach the next four-byte boundary within the section, and
+	//  a four-byte CRC checksum, stored in the same endianness used for the executable file itself.
+	// The checksum is computed on the debugging information file’s full contents by the function given below,
+	// passing zero as the crc argument.
+	file, err := elf.Open(path)
```
Sylfrena, line 155: I can't figure out exactly which  part/function is calculating the checksum here. :thinking:

Sylfrena (approved):

pkg/debuginfo/find.go
kakkoyun: Good catch! The documentation bit is from the tooling's original docs. I'll re-check the implementation and provide examples for this.

kakkoyun: Should be fixed by https://github.com/parca-dev/parca-agent/pull/321/commits/851e570ec4b4f5c8a6a8dfb096f3e4bd2fc37474

Sylfrena (approved):

PR #325, opened 2022-04-05
pkg/target: validate labels

Sylfrena:
Fixes https://github.com/parca-dev/parca/issues/817

Also see https://github.com/parca-dev/parca/pull/842 for the fix on Parca


brancz:
I think we misunderstood each other here. I meant when there is a validation error in Parca because of invalid labels when Parca Agent tries to send invalid labels I would expect the Parca gRPC request to return `InvalidArgument` and Parca Agent to print that error when it receives it.

Sylfrena:
@brancz , does it not already do that here? Or would you prefer to add something more specific like this?
 ```
 if st, _ := status.FromError(err); st.Err() != nil {
	level.Warn(b.logger).Log("msg", "Invalid gRPC write request", "err", st.Code().String())
}
 ```

brancz:
I wasn't sure that we had this, all I meant was I think we should have that. Looks like we're all set already on the parca agent side!

PR #326, opened 2022-04-05
profiler: Use BPF batch operations

kakkoyun:


kakkoyun:
Blocked by https://github.com/aquasecurity/libbpfgo/issues/147

pkg/profiler/profiler.go
```diff
@@ -334,77 +294,114 @@ func (p *CgroupProfiler) Run(ctx context.Context) error {
...
+	if err := it.Err(); err != nil {
+		return fmt.Errorf("iterate over counts map: %w", err)
+	}
 
-		r := bytes.NewBuffer(keyBytes)
+	if batchSize == 0 {
+		return nil
+	}
+	level.Debug(p.logger).Log("msg", "fetching stack trace counts in batch", "batchSize", batchSize)
+	time.Sleep(1 * time.Second)
```
kakkoyun, line 356: @brancz :scream:

v-thakkar: So profiler needs to wait until the BPF program gives the information about stack trace counts. Is that correct? In that case, do you think we can use something like SECCOMP_RET_USER_NOTIF (https://lwn.net/Articles/756233/)?

kakkoyun: This wait should've happened. We still don't know the reason why it works with it. I'm still investigating. Please see the linked issue. 

And thanks for the tip that might help us here, I'll check.

pkg/profiler/profiler.go
```diff
@@ -334,77 +294,114 @@ func (p *CgroupProfiler) Run(ctx context.Context) error {
...
+			pid           = key.pid
+			userStackID   = key.userStackID
+			kernelStackID = key.kernelStackID
+		)
 
-		kernelStackIDBytes := make([]byte, 4)
-		if _, err := io.ReadFull(r, kernelStackIDBytes); err != nil {
-			return fmt.Errorf("read kernel stack ID bytes: %w", err)
+		if pid == 0 {
+			break
```
kakkoyun, line 391: @kakkoyun 
```suggestion
			continue
```

PR #327, opened 2022-04-06
makefile: do not show errors while removing bpf files

javierhonduco:
The `-f` option ignores nonexistent files (https://linux.die.net/man/1/rm)


kakkoyun (approved):
LGTM :100: 

Thanks for your first contribution :smile:

PR #341, opened 2022-04-13
Add Golang build information

javierhonduco:
This commit adds some small wrapper to gather the Go build
information we care about and shows some of these values
during the initialisation of the agent.

Having this information is useful to understand the precise
version that people run.

Now it shows both the commit and the date when it was built
as well as the architecture:
```
$ sudo dist/parca-agent --node=test --systemd-units=docker.service --log-level=debug --kubernetes=false --insecure

ts=2022-04-13T11:38:14.915638561Z caller=main.go:119 msg=starting... node=test store=
level=debug ts=2022-04-13T11:38:14.915780145Z caller=main.go:120 msg="parca-agent initialized" version= commit=d7870a66fa1aa31a9555ba9bb66a6574a5898a17 date=2022-04-13T11:36:50Z builtBy= config="{debug :7071 test map[]    true 10s false 1 false  [docker.service] /tmp  10s  false}" arch=arm64
```

Some possible follow-ups:
- Show this info in `--help`
- Add this information to our BPF code

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -23,7 +23,7 @@ builds:
       - -trimpath
       - -v
     ldflags:
-      - main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}} -X main.builtBy=goreleaser`
+      - main.version={{.Version}} -X main.commit={{.FullCommit}} -X main.date={{.Date}} -X main.builtBy=goreleaser -X main.goArch={{.Runtime.Goarch}}
   - main: ./cmd/debug-info/
     id: "debug-info"
     binary: parca-debug-info
@@ -37,7 +37,7 @@ builds:
       - -trimpath
       - -v
     ldflags:
-      - main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}} -X main.builtBy=goreleaser`
+      - main.version={{.Version}} -X main.commit={{.FullCommit}} -X main.date={{.Date}} -X main.goArch={{.Runtime.Goarch}}
 archives:
   - replacements:
       linux: Linux
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -23,7 +23,7 @@ builds:
       - -trimpath
       - -v
     ldflags:
-      - main.version={{.Version}} -X main.commit={{.FullCommit}} -X main.date={{.Date}} -X main.builtBy=goreleaser -X main.goArch={{.Runtime.Goarch}}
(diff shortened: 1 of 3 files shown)


javierhonduco:
Removed extraneous newline + `FindBuildInfo` => `FetchBuildInfo`

cmd/parca-agent/main.go
```diff
@@ -103,6 +105,16 @@ func main() {
 
 	logger := logger.NewLogger(flags.LogLevel, logger.LogFormatLogfmt, "")
 
+	// Fetch build info such as the git revision we are based off
+	buildInfo, err := buildinfo.FetchBuildInfo()
```
kakkoyun, line 109: Overall, this PR looks good. It would help us to discover  build information during runtime.

However, those variables are set by our build tool from outside while cross-building the binaries. https://github.com/parca-dev/parca-agent/blob/ff872f449ea39ae9d37988bbc4162e905f955db0/.goreleaser.yml#L26

So we should check if those variables have already values before we set them.

kakkoyun (changes requested):

javierhonduco:
Address review feedback + pass the architecture flag in the gorelease file

.goreleaser.yml
```diff
@@ -23,7 +23,7 @@ builds:
       - -trimpath
       - -v
     ldflags:
-      - main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}} -X main.builtBy=goreleaser`
```
javierhonduco, line 26: I am a bit confused by the backtick, I am assuming it was a typo? 😀

kakkoyun: It probably is :rofl:

cmd/parca-agent/main.go
```diff
@@ -60,6 +61,7 @@ var (
 	commit  string
 	date    string
 	builtBy string
```
kakkoyun, line 63: nit: I saw that it's not used anymore, let's remove `builtBy`, WDTY?

javierhonduco: Oh, thought it was used by `goreleaser`! Happy to remove it otherwise!

javierhonduco:
Thanks for retrying it! Unfortunately it seems like it's a legit issue:

```
2022-04-13T15:12:34.1268657Z pkg/buildinfo/buildinfo.go:34:28: bi.Settings undefined (type *debug.BuildInfo has no field or method Settings)
```

I suppose that we are running an older Golang version in CI, checking :)

kakkoyun (approved):
:100:

PR #343, opened 2022-04-13
profiler: Properly free and clean resources

kakkoyun:

cc @v-thakkar @javierhonduco @Sylfrena

### DIFF
--- a/pkg/profiler/profiler.go
+++ b/pkg/profiler/profiler.go
@@ -26,6 +26,7 @@ import (
 	"runtime"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 	"unsafe"
 
@@ -250,7 +251,6 @@ func (p *CgroupProfiler) Run(ctx context.Context) error {
 
 	cpus := runtime.NumCPU()
 	for i := 0; i < cpus; i++ {
-		// TODO(branz): Close the returned fd
 		fd, err := unix.PerfEventOpen(&unix.PerfEventAttr{
 			Type:   unix.PERF_TYPE_SOFTWARE,
 			Config: unix.PERF_COUNT_SW_CPU_CLOCK,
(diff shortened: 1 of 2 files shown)


pkg/profiler/profiler.go
```diff
@@ -261,6 +261,11 @@ func (p *CgroupProfiler) Run(ctx context.Context) error {
 		if err != nil {
 			return fmt.Errorf("open perf event: %w", err)
 		}
+		defer func() {
```
javierhonduco, line 264: Sorry, just noticed something here! If I am not mistaken, this might not be what we want, and we won't close all the FDs we want! What is worse, we might be accidentally closing FDs we don't want to (if we race with another process).

I _think_ we want to pass the FD as an argument, such as: https://go.dev/play/p/tB-jjUUaNZ4 instead of https://go.dev/play/p/e9vpgsF3364.

This is due to the way defer seems to work, as it evaluates the arguments at the time `defer` is assembled, instead of when the function is finally called, but currently we used the variable that's in the scope, which will always be the last FD.

I might be 100% wrong, let's learn some Go together 😄

pkg/profiler/profiler.go
```diff
@@ -269,10 +274,15 @@ func (p *CgroupProfiler) Run(ctx context.Context) error {
 
 		// Because this is fd based, even if our program crashes or is ended
 		// without proper shutdown, things get cleaned up appropriately.
-		// TODO(brancz): destroy the returned link via bpf_link__destroy
-		if _, err := prog.AttachPerfEvent(fd); err != nil {
+		link, err := prog.AttachPerfEvent(fd)
+		if err != nil {
 			return fmt.Errorf("attach perf event: %w", err)
 		}
+		defer func() {
```
javierhonduco, line 281: This might be potentially the same issue as above, only that slightly less dangerous I think 😄

pkg/profiler/profiler.go
kakkoyun: :see_no_evil: Amazing catch! I missed that we're in a loop!

kakkoyun: Now that, I've checked it again. We use `fd` in the defer closure and `fd` is local so, it should be a problem.

kakkoyun: https://go.dev/play/p/VxHz5xAekZH

javierhonduco: You are 100% right! https://go.dev/play/p/ucsC6MeLYsX My bad :)

PR #345, opened 2022-04-13
profiler: Increase rlimit

kakkoyun:

Until we have the following feature lands in the libbpf, we need a way to increase rlimit. libbpf-go sets it to 512MB when the module loaded, however I always hits this limit whenever I change something regarding BPF maps so, I'm introducing a way to increase it and also by default this Pr increases that limit to 1024MB.

Check out rlimits if you ever encounter `operation not permitted` error.

cc @javierhonduco @v-thakkar @Sylfrena


javierhonduco (commented):
LGTM!

Issue #328, opened 2022-04-06
parca-agent: /lib64/libc.so.6: version `GLIBC_2.28' not found (required by parca-agent)


tkcreddy:
While starting up the parca-agent, we are getting this error. this is on Amazon Linux instance.

kakkoyun:
@tkcreddy Thanks for reaching out :+1: 

Could you please describe your setup a little more? I assume you install `parca-agent` directly on Amazon Linux and use systemd to discover cgroups? Or this is a Kubernetes setup that uses Amazon Linux (which is more worrisome because I see that dependency presents in the container image)?

https://github.com/parca-dev/parca-agent/blob/bbafe10412a384b64d252f24fdec90d6e3a1e1e2/Dockerfile#L42-L55

In either case, we'll make sure we update our docs or make sure the dependencies are present (either by linking dependencies statically or having dependencies in the container)

SINGH-IOT2020:
I am also getting the same error message while starting the parca-agent in amazon linux ec2 instance. 
Details  are below 👎 
parca-agent.service - Parca Agent
   Loaded: loaded (/etc/systemd/system/parca-agent.service; enabled; vendor preset: disabled)
   Active: activating (auto-restart) (Result: exit-code) since Thu 2022-04-07 14:59:16 UTC; 9s ago
  Process: 1460 ExecStart=/usr/bin/parca-agent --http-address=":7071" --node=systemd-test --systemd-units=parca.service,parca-agent.service --kubernetes=false --store-address=localhost:7070 --insecure (code=exited, status=1/FAILURE)
------------------------------------
To fix the above issue I ran command   -systemctl  enable parca-agent 
Still same error message as above
----------------------------------------
Try to start parca agent manually . Got below error message 👍 

# ./parca-agent
./parca-agent: /lib64/libc.so.6: version `GLIBC_2.28' not found (required by ./parca-agent)

Please suggest and help to resolve this.

kakkoyun:
@SINGH-IOT2020 Could you try installing certain dependencies and try again? For a Debian based distro they look like below:
https://github.com/parca-dev/parca-agent/blob/bbafe10412a384b64d252f24fdec90d6e3a1e1e2/Dockerfile#L15-L18

SINGH-IOT2020:
Hi, Thanks for response. 
I install the dependency and still getting the same error message :
[root@ip-172-31-30-170 ec2-user]# yum install llvm binutils elfutils
[...]

PR #351, opened 2022-04-14
Produce statically linked binaries

### LINKED ISSUE #332: Statically link or remove libc from binaries
- To static link: `go build --ldflags '-linkmode external -extldflags "-static"` or `go build -ldflags="-extldflags=-static"`
- To (potentially) remove: `go build -tags netgo` `!! This might not be possible with CGO and this could change the behaviour of DNS resolution, so we should test it.`

> User/group resolution, and the non-DNS aspects for host resolution, are code that only exists in the libc, and the way glibc implements those is by dynamically loading libraries based on config files. Unless you explicitly opt out (use pure-Go implementations that imitate some of the behavior), if you use either of those facilities, you will get a dynamically linked glibc.

ref: 
https://ops.tips/notes/glibc-golang-learnings/
https://www.arp242.net/static-go.html

kakkoyun:

Fixes #332 #328 

# TODO

* [x] Clean up Dockerfiles
* [x] Make sure CGO cross-compilation works (using Goreleaser)
* [x] Test against arm64

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -24,6 +24,10 @@ builds:
       - -v
     ldflags:
       - main.version={{.Version}} -X main.commit={{.FullCommit}} -X main.date={{.Date}} -X main.goArch={{.Runtime.Goarch}}
+      - -extldflags=-static
+    tags:
+      - osusergo
+      - netgo
   - main: ./cmd/debug-info/
     id: "debug-info"
     binary: parca-debug-info
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,24 +36,12 @@ RUN make build
 
 FROM ${DEBIAN_BASE} as all
 
-ARG LINUX_ARCH
-ENV LINUX_ARCH=$LINUX_ARCH
-
 COPY --from=build /etc/nsswitch.conf /etc/nsswitch.conf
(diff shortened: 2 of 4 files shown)


brancz:
Does this mean we can also get rid of copying them in our dockerfile here? https://github.com/parca-dev/parca-agent/blob/0c05f0048b6773011bcbeeb26221e6e8cadb5e0e/Dockerfile#L45-L52

kakkoyun:
@brancz Good call. Let me make check that.

brancz (approved):
😍 lgtm

javierhonduco:
Thanks again for implementing this, just learned that ASAN doesn't run on statically linked binaries

```
github.com/parca-dev/parca-agent/pkg/discovery
github.com/parca-dev/parca-agent/cmd/parca-agent
# github.com/parca-dev/parca-agent/cmd/parca-agent
/usr/lib/golang/pkg/tool/linux_amd64/link: running clang failed: exit status 1
/usr/bin/ld: /usr/lib64/clang/14.0.0/lib/linux/libclang_rt.asan-x86_64.a(asan_linux.cpp.o): in function `__asan::AsanDoesNotSupportStaticLinkage()':
(.text+0x37): undefined reference to `_DYNAMIC'
clang-14: error: linker command failed with exit code 1 (use -v to see invocation)
```

so `build-dyn` unblocked me 🎉

PR #352, opened 2022-04-14
profiler: Refactor profileLoop for more readability 

kakkoyun:

* Split profileLoop into semantically sound functions
* Use binary.Read to ready BPF keys to Go structs
* Add recovery mechanism to profileLoop for unexpected panics
* Handle known errors

### DIFF
--- a/pkg/ksym/ksym.go
+++ b/pkg/ksym/ksym.go
@@ -56,6 +56,7 @@ func NewKsymCache(logger log.Logger) *Cache {
 	}
 }
 
+// TODO(kakkoyun): https://github.com/aquasecurity/libbpfgo/blob/main/helpers/kernel_symbols.go
 func (c *Cache) Resolve(addrs map[uint64]struct{}) (map[uint64]string, error) {
 	c.mtx.RLock()
 	lastCacheInvalidation := c.lastCacheInvalidation
--- a/pkg/profiler/profiler.go
+++ b/pkg/profiler/profiler.go
@@ -23,7 +23,6 @@ import (
 	"fmt"
 	"os"
 	"runtime"
-	"runtime/debug"
 	"strings"
 	"sync"
 	"syscall"
@@ -62,8 +61,10 @@ const (
 	stackDepth       = 127 // Always needs to be sync with MAX_STACK_DEPTH in parca-agent.bpf.c
 	doubleStackDepth = 254
 
-	defaultRlimit = 1024 << 20 // ~1GB
+	defaultRLimit = 1024 << 20 // ~1GB
 )
+
+type stack [doubleStackDepth]uint64
 
 type bpfMaps struct {
 	counts      *bpf.BPFMap
@@ -71,6 +72,10 @@ type bpfMaps struct {
 }
 
 // stackCountKey mirrors the struct in parca-agent.bpf.c
+// NOTICE: The memory layout and alignment of the struct currently matches the struct in parca-agent.bpf.c.
+// However, keep in mind that Go compiler injects padding to align the struct fields to be a multiple of 8 bytes.
+// The Go spec says the address of a struct’s fields must be naturally aligned.
+// https://dave.cheney.net/2015/10/09/padding-is-hard
 // TODO: https://github.com/parca-dev/parca-agent/issues/207
 type stackCountKey struct {
 	PID           uint32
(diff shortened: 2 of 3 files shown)


pkg/ksym/ksym.go
```diff
@@ -56,6 +56,7 @@ func NewKsymCache(logger log.Logger) *Cache {
 	}
 }
 
+// TODO(kakkoyun): https://github.com/aquasecurity/libbpfgo/blob/main/helpers/kernel_symbols.go
```
brancz, line 59: that looks better but ultimately we should probably just implement a real parser

pkg/profiler/profiler.go
```diff
@@ -372,69 +374,52 @@ func (p *CgroupProfiler) Run(ctx context.Context) error {
...
-	mapping := maps.NewMapping(p.pidMappingFileCache)
-	kernelMapping := &profile.Mapping{
-		// TODO(kakkoyun): Check if this conflicts with https://github.com/google/pprof/pull/675/files
-		File: "[kernel.kallsyms]",
-	}
-	kernelFunctions := map[uint64]*profile.Function{}
-	userFunctions := map[[2]uint64]*profile.Function{}
+func (p *CgroupProfiler) profileLoop(ctx context.Context, captureTime time.Time) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
```
brancz, line 379: I think I would prefer if we crash and the user notices more drastically to encourage people to open bugs rather than just continue.

pkg/profiler/profiler.go
```diff
@@ -372,69 +374,52 @@ func (p *CgroupProfiler) Run(ctx context.Context) error {
...
-	// 2 uint64 1 for PID and 1 for Addr
-	locations := []*profile.Location{}
-	kernelLocations := []*profile.Location{}
-	kernelAddresses := map[uint64]struct{}{}
-	locationIndices := map[[2]uint64]int{}
-	samples := map[[doubleStackDepth]uint64]*profile.Sample{}
+	var (
+		mappings      = maps.NewMapping(p.pidMappingFileCache)
+		kernelMapping = &profile.Mapping{
+			// TODO(kakkoyun): Check if this conflicts with https://github.com/google/pprof/pull/675/files
```
brancz, line 389: I don't know if I did it correctly, but I chose the file name specifically to be consistent with what pprof does

pkg/profiler/profiler.go
```diff
@@ -372,69 +374,52 @@ func (p *CgroupProfiler) Run(ctx context.Context) error {
...
-		if _, err := io.ReadFull(r, pidBytes); err != nil {
-			return fmt.Errorf("read pid bytes: %w", err)
-		}
-		pid := p.byteOrder.Uint32(pidBytes)
-		if pid == 0 {
-			level.Debug(p.logger).Log("msg", "missing pid")
-			p.metrics.missingPIDs.Inc()
-			continue
+		var key stackCountKey
+		if err := binary.Read(bytes.NewBuffer(keyBytes), p.byteOrder, &key); err != nil {
```
brancz, line 408: Are you 100% sure this works as expected? I remember trying something like this but running into problems because of automatic byte alignment that the Go compiler applies to the `stackCountKey` struct.

pkg/profiler/profiler.go
```diff
@@ -550,15 +511,68 @@ func (p *CgroupProfiler) profileLoop(ctx context.Context, captureTime time.Time)
...
+	defer p.mtx.Unlock()
+
+	p.lastProfileTakenAt = lastProfileTakenAt
+	p.lastError = lastError
+}
+
+func (p *CgroupProfiler) buildProfile(
+	ctx context.Context,
+	captureTime time.Time,
+	samples map[[254]uint64]*profile.Sample,
```
brancz, line 541: ooooooo we can't use the constant for this 254 here, can we?

brancz (approved):
I would feel quite a bit more confident if we had true e2e tests, but it looks good.

pkg/profiler/profiler.go
kakkoyun: Happy to remove it. :+1:

pkg/profiler/profiler.go
kakkoyun: No, we can't. IDE generated signature, I've overlooked it. I will introduce a specific type for it.

pkg/profiler/profiler.go
kakkoyun: This just work because the struct we have is perfectly aligned! So I will add a notice in case we need to change it we should think about alignment.

https://dave.cheney.net/2015/10/09/padding-is-hard

kakkoyun:
I'm just gonna merge this and test it until our next release. (I'm planning to cut one next week)

PR #346, opened 2022-04-13
Initial try for .eh_frame/.debug_frame based stack unwinding using eBPF

kakkoyun:
Signed-off-by: Kemal Akkoyun <kakkoyun@gmail.com>

blocked by xref: https://github.com/aquasecurity/libbpfgo/issues/93

Fixes #293 

- [x] Convince eBPF verifier that I'm not trying to harm anyone
- [x] Make sure it is locally testable https://github.com/kubernetes/minikube/pull/12707
- [x] Support multiple kernel versions (>= 4.19, >= 5.2 (with BTF))
- [x] Increase memlock rlimit dynamically
- [x] Move process discovery before attaching the eBPF program
  - [ ] Test for the simple case
- [ ] Rebase
- [ ] Split and commit BTF, Linux 5.2+ changes 
- [ ] ~Implement BPF map of maps after we have the support in libbpf-go:~ Utilize https://github.com/aquasecurity/libbpfgo/pull/138
- [ ] SystemdUnit .PID() implement
- [ ] Use perf_map to find out executable
- [ ] Handle unwinding information of dynamically linked libraries (multiple mappings)
- [ ] Test and prove it works visually
- [ ] Add debug_frame support
- [ ] Conclude and properly vendor delve/dwarf package and ask upstream if they have these changes
- [ ] Clean up TODO comments in code

PR #354, opened 2022-04-19
profiler: Keep track of stack unwinding failures

kakkoyun:

As a pre-work to #346, add metrics to keep track of stack unwinding failures.

cc @v-thakkar

### DIFF
--- a/parca-agent.bpf.c
+++ b/parca-agent.bpf.c
@@ -17,7 +17,6 @@
 #define KBUILD_MODNAME "parca-agent"
 
 #undef container_of
-//#include "bpf_core_read.h"
 #include <bpf_core_read.h>
 #include <bpf_endian.h>
 #include <bpf_helpers.h>
@@ -33,6 +32,8 @@
 #define MAX_STACK_ADDRESSES 1024
 // Max depth of each stack trace to track
 #define MAX_STACK_DEPTH 127
+
+/*================================ eBPF MAPS =================================*/
 
 #define BPF_MAP(_name, _type, _key_type, _value_type, _max_entries)            \
   struct {                                                                     \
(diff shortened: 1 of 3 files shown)


parca-agent.bpf.c
```diff
@@ -97,9 +100,27 @@ int do_sample(struct bpf_perf_event_data *ctx) {
...
+
+  // get kernel stack id
+  key.kernel_stack_id = 0;
+  int kernel_stack_id = bpf_get_stackid(ctx, &stack_traces, 0);
+  if (kernel_stack_id >= 0)
+    key.kernel_stack_id = kernel_stack_id;
+
+  // TODO(kakkoyun): failed bpf_get_stackid() could indicate stack unwinding
+  // issues; this could be a useful place to hook eh_frame-based stack
+  // unwinding.
```
v-thakkar, line 117: :100:

v-thakkar:
LGTM!

parca-agent.bpf.c
```diff
@@ -97,9 +100,27 @@ int do_sample(struct bpf_perf_event_data *ctx) {
   // create map key
   stack_count_key_t key = {.pid = tgid};
```
javierhonduco, line 101: nit: for consistency, should we initialise the other fields here?

javierhonduco (commented):
Looks great!! 

I think you might have mentioned this at some point, but perhaps we could check that the addresses in userspace are all within the stack low<->top addresses? I am unsure if the kernel does this already (checking now!) but if it doesn't it might be worth doing the check ourselves

kakkoyun:
> Looks great!!
> 
> I think you might have mentioned this at some point, but perhaps we could check that the addresses in userspace are all within the stack low<->top addresses? I am unsure if the kernel does this already (checking now!) but if it doesn't it might be worth doing the check ourselves

I don't think kernel does this. I came to that conclusion because we received addresses that we can't map any object file. AFAIK (don't quote me on that), if we receive out-of-bounds addresses, we can say that the Kernel failed to unwind the stack. The only boundary check we have to find the mapping.

https://github.com/parca-dev/parca-agent/blob/e8afc39e6c19f9cb1fa2684a150ae85068d3de6b/pkg/maps/mapping.go#L82-L90

Wed Apr 20 02:19:27 PM CEST 2022

The addresses that we couldn't symbolize and don't have `[*]` prefix is how it manifests itself on icicle graphs.

We should investigate this more, though. Happy to happen an issue (if we haven't gotten one yet)!

PR #360, opened 2022-04-20
Update to Go 1.18

javierhonduco:


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/parca-dev/parca-agent
 
-go 1.17
+go 1.18
```
kakkoyun, line 3: This is actually not needed. This specified minimum required version to compile this module/codebase, if we don't need a specific version from 1.18 we shouldn't update this.

kakkoyun (changes requested):
LGTM. Thanks for taking care of it.

I only pointed one issue.

go.mod
javierhonduco: I think we need it for https://github.com/parca-dev/parca-agent/pull/341 (https://tip.golang.org/doc/go1.18 says that `Settings` was made public in that release)

kakkoyun: Ah, then yes. Sorry I've missed that. It'd have been better to add these changes to that PR then.

kakkoyun (approved):

PR #361, opened 2022-04-20
*: Enable more linters

kakkoyun:
Fix issues accordingly


kakkoyun:
cc @v-thakkar @javierhonduco

kakkoyun:
These actually helped to catch a couple of possible issues and helped to improve design

pkg/profiler/profiler.go
```diff
@@ -712,22 +713,24 @@ func (p *CgroupProfiler) sendProfile(ctx context.Context, prof *profile.Profile)
...
-			&profilestorepb.Label{
-				Name:  string(key),
-				Value: string(value),
-			})
+		labelOldFormat[i] = &profilestorepb.Label{
+			Name:  string(key),
+			Value: string(value),
+		}
+		i++
 	}
```
Sylfrena, line 726: Nothing I want changed, but curious, why is this a better way? :thinking:

kakkoyun: Rather than appending to a dynamically sized slice, this pre-allocates and fixes the slice size, then writing the value in specific positions. Just to avoid growing the slice. Really a micro optimization but static analyzer was smart enough to catch and point it.

pkg/profiler/profiler.go
```diff
@@ -712,22 +713,24 @@ func (p *CgroupProfiler) sendProfile(ctx context.Context, prof *profile.Profile)
...
+		labelOldFormat = make([]*profilestorepb.Label, 0, len(p.Labels()))
+		i              = 0
+	)
 	for key, value := range p.Labels() {
-		labeloldformat = append(labeloldformat,
-			&profilestorepb.Label{
-				Name:  string(key),
-				Value: string(value),
-			})
+		labelOldFormat[i] = &profilestorepb.Label{
```
javierhonduco, line 721: Since we have a slice with capacity=`len(p.Labels())`
 and length 0, we need to use `append` to avoid writing out of bounds, no? :D 

> 

https://go.dev/play/p/17_G4OQD4eU

PR #366, opened 2022-04-21
Add e2e tests

### LINKED ISSUE #317: Basic e2e test setup
We've had a couple of bugs by now that would have been caught by a simple e2e tests. The suggested setup for a test like this in CI:

* setup minikube
* setup parca
* setup parca-agent
* do a gRPC range query across all of time to ensure _some_ data is successfully returned

Once we have this setup we can extend it wit things like #190.

Sylfrena:
Very early version, so excuse the lazy shell scripting, plus need to add a lot more test cases. 

Use `make dev/e2e`  from root of the repo to run the tests.

Fixes #317

cc @v-thakkar @javierhonduco

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -258,3 +258,8 @@ dev/up: deploy/manifests
 .PHONY: dev/down
 dev/down:
 	source ./scripts/local-dev.sh && down
+
+.PHONY: test-e2e $(driver)
+test-e2e:
+	cd deploy; source ./../e2e/local-e2e.sh && run $(driver)
+	go test -v $(shell go list ./e2e | grep "e2e")
--- a/Makefile
+++ b/Makefile
@@ -147,7 +147,7 @@ test/profiler: $(GO_SRC) $(LIBBPF_HEADERS) $(LIBBPF_OBJ) bpf
 .PHONY: test
 ifndef DOCKER
 test: $(GO_SRC) $(LIBBPF_HEADERS) $(LIBBPF_OBJ) bpf test/profiler
-	$(go_env) go test -run '[^TestIntegrationGRPC]' -v $(shell go list ./... | grep -v "internal/pprof" | grep -v "pkg/profiler")
+	$(go_env) go test -v $(shell go list ./... | grep -v "internal/pprof" | grep -v "pkg/profiler")
 else
 test: $(DOCKER_BUILDER)
 	$(call docker_builder_make,$@)
@@ -262,4 +262,4 @@ dev/down:
 .PHONY: test-e2e $(driver)
 test-e2e:
 	cd deploy; source ./../e2e/local-e2e.sh && run $(driver)
-	go test -v $(shell go list ./e2e | grep "e2e")
+	go test -tags=integration -v $(shell go list -tags=integration ./e2e)
--- a/Makefile
+++ b/Makefile
@@ -147,7 +147,7 @@ test/profiler: $(GO_SRC) $(LIBBPF_HEADERS) $(LIBBPF_OBJ) bpf
 .PHONY: test
 ifndef DOCKER
 test: $(GO_SRC) $(LIBBPF_HEADERS) $(LIBBPF_OBJ) bpf test/profiler
-	$(go_env) go test -v $(shell go list ./... | grep -v "internal/pprof" | grep -v "pkg/profiler")
+	$(go_env) go test -v $(shell go list ./... | grep -v "internal/pprof" | grep -v "pkg/profiler" | grep -v "e2e")
 else
 test: $(DOCKER_BUILDER)
 	$(call docker_builder_make,$@)
@@ -262,4 +262,4 @@ dev/down:
 .PHONY: test-e2e $(driver)
 test-e2e:
 	cd deploy; source ./../e2e/local-e2e.sh && run $(driver)
-	go test -tags=integration -v $(shell go list -tags=integration ./e2e)
+	go test -v $(shell go list ./e2e)
--- a/Makefile
+++ b/Makefile
@@ -147,7 +147,7 @@ test/profiler: $(GO_SRC) $(LIBBPF_HEADERS) $(LIBBPF_OBJ) bpf
 .PHONY: test
 ifndef DOCKER
 test: $(GO_SRC) $(LIBBPF_HEADERS) $(LIBBPF_OBJ) bpf test/profiler
-	$(go_env) go test -v $(shell go list ./... | grep -v "internal/pprof" | grep -v "pkg/profiler")
+	$(go_env) go test -run '[^TestIntegrationGRPC]' -v $(shell go list ./... | grep -v "internal/pprof" | grep -v "pkg/profiler")
 else
 test: $(DOCKER_BUILDER)
 	$(call docker_builder_make,$@)
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -11,9 +11,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build integration
-// +build integration
-
 package e2e
 
 import (
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -1,6 +1,3 @@
-//go:build integration
-// +build integration
-
 // Copyright 2022 The Parca Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -13,6 +10,10 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+
+//go:build integration
+// +build integration
+
 package e2e
 
 import (
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -1,3 +1,6 @@
+//go:build integration
+// +build integration
+
 // Copyright 2022 The Parca Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -1,0 +1,47 @@
+// Copyright 2022 The Parca Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package e2e
+
+import (
+	"context"
+	"testing"
+
+	"github.com/prometheus/prometheus/model/timestamp"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	pb "github.com/parca-dev/parca/gen/proto/go/parca/query/v1alpha1"
+)
+
+func TestGRPC(t *testing.T) {
+	conn, err := grpc.Dial("127.0.0.1:7070", grpc.WithTransportCredentials(insecure.NewCredentials()))
+	require.NoError(t, err)
+	defer conn.Close()
+
+	c := pb.NewQueryServiceClient(conn)
+	ctx := context.Background()
+
+	queryRequest := &pb.QueryRangeRequest{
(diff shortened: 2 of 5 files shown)


e2e/e2e-test-parca.sh
```diff
@@ -0,0 +1,88 @@
+################################################################################
+#
+# This script is meant to be run from the root of this project
+#
+################################################################################
+
+#!/bin/bash
```
kakkoyun, line 7: This needs to at the top of the file, no?

e2e/e2e-test-parca.sh
```diff
@@ -0,0 +1,88 @@
...
+
+set -euo pipefail
+
+function e2erun() {
```
kakkoyun, line 11:
```suggestion
function e2e_run() {
```

e2e/e2e-test-parca.sh
```diff
@@ -0,0 +1,88 @@
...
+  minikube_cluster_up
+  deploy
+  # minikube_cluster_down genius, spin deploy and delete ugh
+}
+
+# Create local minikube cluster and deploys the dev env for parca and parca agent 
+function minikube_cluster_up() {
+  echo "Spinning up a parca dev cluster"
+  minikube start -p parca-dev \
```
kakkoyun, line 20:
```suggestion
  minikube start -p parca-e2e \
```
nit: just to be super specific

e2e/e2e-test-parca.sh
```diff
@@ -0,0 +1,88 @@
...
+set -euo pipefail
+
+function e2erun() {
+  minikube_cluster_up
+  deploy
+  # minikube_cluster_down genius, spin deploy and delete ugh
+}
+
+# Create local minikube cluster and deploys the dev env for parca and parca agent 
+function minikube_cluster_up() {
```
kakkoyun, line 18: Why not just `up`?

e2e/e2e-test-parca.sh
```diff
@@ -0,0 +1,88 @@
...
+# Delete minikube instance
+function minikube_cluster_down() {
+  echo "deleting parca dev cluster"
+  minikube delete -p parca-dev
+}
+
+# Fetches latest binaries for parca and parca agent 
+function deploy() {
+  SERVER_LATEST_VERSION=$(curl -s https://api.github.com/repos/parca-dev/parca/releases/latest | grep -oP '"tag_name": "\K(.*)(?=")' | xargs echo -n)
+  AGENT_LATEST_VERSION=$(curl -s https://api.github.com/repos/parca-dev/parca-agent/releases/latest | grep -oP '"tag_name": "\K(.*)(?=")' | xargs echo -n)
```
kakkoyun, line 39: We shouldn't be needing this. We want to build a container from current branch, no?

e2e/e2e-test-parca.sh
```diff
@@ -0,0 +1,88 @@
...
+
+ # jb install $(find . -name 'vendor' -prune -o -name '*.libsonnet' -print -o -name '*.jsonnet' -print)
+	
+  echo blah
+  rm -rf manifests/local 
+  mkdir -p manifests/local
+  
+  echo blahhh
+  jsonnet --tla-str version="e2e-f655897a0ab80c18" -J vendor main.jsonnet -m manifests/local | xargs -I{} sh -c 'cat {} | gojsontoyaml > {}.yaml; rm -f {}' -- {}
+  awk 'BEGINFILE {print "---"}{print}' manifests/local/* > manifests/local/manifest-e2e.yaml
```
kakkoyun, line 87: We don't need to merge that into a single file. We can just use the whole directory.
We merge them for production just to make it easier to deploy by just downloading a single file. In this case, we don't need it.

e2e/local-e2e.sh
```diff
@@ -0,0 +1,84 @@
...
+    --driver=kvm2 \
+    --kubernetes-version=v1.22.3 \
+    --cpus=4 \
+    --memory=16gb \
+    --disk-size=20gb \
+    --docker-opt dns=8.8.8.8 \
+    --docker-opt default-ulimit=memlock=9223372036854775807:9223372036854775807
+}
+
+#Delete minikube instance
```
kakkoyun, line 31:
```suggestion
# Delete minikube instance
```

e2e/local-e2e.sh
```diff
@@ -0,0 +1,84 @@
+################################################################################
+#
+# This script is meant to be run from the root of this project
+#
+################################################################################
+
+#!/bin/bash
```
kakkoyun, line 7: Ditto.

e2e/local-e2e.sh
```diff
@@ -0,0 +1,84 @@
...
+  SERVER_LATEST_VERSION=$(curl -s https://api.github.com/repos/parca-dev/parca/releases/latest | grep -oP '"tag_name": "\K(.*)(?=")' | xargs echo -n)
+  AGENT_LATEST_VERSION=$(curl -s https://api.github.com/repos/parca-dev/parca-agent/releases/latest | grep -oP '"tag_name": "\K(.*)(?=")' | xargs echo -n)
+
+  echo "Server version: $SERVER_LATEST_VERSION"
+  echo "Agent version: $AGENT_LATEST_VERSION"
+
+  kubectl apply -f https://github.com/parca-dev/parca/releases/download/"$SERVER_LATEST_VERSION"/kubernetes-manifest.yaml
+  kubectl apply -f ./manifests/local/manifest-e2e.yaml
+
+  sleep 1m  #add a probe or something
```
kakkoyun, line 48: You can check using kubectl if they are rollout.

https://kubernetes.io/docs/reference/kubectl/cheatsheet/#updating-resources
https://jamesdefabia.github.io/docs/user-guide/kubectl/kubectl_rollout/
https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#-em-status-em-
https://kubernetes.io/docs/tasks/manage-daemon/update-daemon-set/

e2e/e2e-test-parca.sh
kakkoyun: https://blog.hermes-technology.de/minikube/kubernetes/podman/2020/03/23/use-local-images-in-minikube-3-methods.html

https://minikube.sigs.k8s.io/docs/handbook/pushing/

Sylfrena:
fyi, the Go build in CI here fails since the test also requires the Parca server to be running.

e2e/e2e_test.go
```diff
@@ -0,0 +1,50 @@
...
+	queryRequest := &pb.QueryRangeRequest{
+		Query: `parca_agent_cpu:samples:count:cpu:nanoseconds:delta{container="parca-agent"}`,
+		Start: timestamppb.New(timestamp.Time(0)),
+		End:   timestamppb.New(timestamp.Time(9223372036854775807)),
+		Limit: 10,
+	}
+
+	resp, err := c.QueryRange(ctx, queryRequest)
+
+	fmt.Println(resp.String())
```
Sylfrena, line 46: Remove this.

Sylfrena, line 46:
```suggestion
```

e2e/local-e2e.sh
```diff
@@ -0,0 +1,115 @@
...
+  deploy
+}
+
+# Create local minikube cluster and deploys the dev env for parca and parca agent
+function minikube_up(){
+  echo "Spinning up a parca dev cluster"
+  minikube start -p parca-e2e \
+    --container-runtime=docker \
+    --insecure-registry="localhost:5000"\
+    --driver=kvm2 \
```
kakkoyun, line 23: Let's make this configurable. So that people on other platforms can run them by specifying it.

e2e/local-e2e.sh
```diff
@@ -0,0 +1,115 @@
...
+function deploy() {
+  SERVER_LATEST_VERSION=$(curl -s https://api.github.com/repos/parca-dev/parca/releases/latest | grep -oP '"tag_name": "\K(.*)(?=")' | xargs echo -n)
+  echo "Server version: $SERVER_LATEST_VERSION"
+
+  if ! check_ns_parca; then
+    kubectl create namespace parca
+  fi
+
+  kubectl apply -f https://github.com/parca-dev/parca/releases/download/"$SERVER_LATEST_VERSION"/kubernetes-manifest.yaml
+  while ! is_rollout_finished "deployment/parca" ; do
```
kakkoyun, line 50: `kubectl rollout status deployment` should already suppose to wait until deployment actually rollout out. The loop seemed overkill.

Also, there's a `--timeout` flag we should specify a reasonable timeout value to prevent this script to run forever.

e2e/local-e2e.sh
```diff
@@ -0,0 +1,115 @@
...
+    VERSION=$RELEASE_TAG
+  fi
+
+  rm -rf manifests/local
+  mkdir -p manifests/local
+
+  jsonnet --tla-str version="$VERSION" -J vendor e2e.jsonnet -m manifests/local | xargs -I{} sh -c 'cat {} | gojsontoyaml > {}.yaml; rm -f {}'
+  awk 'BEGINFILE {print "---"}{print}' manifests/local/* > manifests/local/manifest-e2e.yaml
+
+  docker build -t localhost:5000/parca-agent:"$VERSION" -f ./../Dockerfile.dev ./..
```
kakkoyun, line 114:
```suggestion
  docker build -t localhost:5000/parca-agent:"$VERSION" -f ./../Dockerfile ./..
```

We should use the Dockerfile for the production.

kakkoyun (commented):
Awesome job @Sylfrena!

I have a couple of nits. And let's add a section to CONTRIBUTING.md about end-to-end tests.

kakkoyun:
You might want to run `make go/deps` to fix failing checks.

e2e/local-e2e.sh
```diff
@@ -0,0 +1,114 @@
...
+  fi
+
+  rm -rf manifests/local
+  mkdir -p manifests/local
+
+  jsonnet --tla-str version="$VERSION" -J vendor e2e.jsonnet -m manifests/local | xargs -I{} sh -c 'cat {} | gojsontoyaml > {}.yaml; rm -f {}'
+  awk 'BEGINFILE {print "---"}{print}' manifests/local/* > manifests/local/manifest-e2e.yaml
+
+  # docker.io/golang:1.18.1-bullseye
+  GOLANG_BASE='docker.io/golang@sha256:ee752bc53c628ff789bacefb714cff721701042ffa9eb736f7b2ed4e9f2bdab6'
```
kakkoyun, line 106: To make these arch independent let's not specify the hash. Tag should be enough.

e2e/local-e2e.sh
```diff
@@ -0,0 +1,114 @@
...
+#
+# This script is meant to be run from the root of this project
+#
+################################################################################
+
+set -euo pipefail
+
+function run() {
+  # Driver set to docker by default if not specified
+  DRIVER=${1-docker}
```
kakkoyun, line 13: Parca Agent wouldn't work docker driver. Let's go with VirtualBox as default.

Makefile
```diff
@@ -147,7 +147,7 @@ test/profiler: $(GO_SRC) $(LIBBPF_HEADERS) $(LIBBPF_OBJ) bpf
 .PHONY: test
 ifndef DOCKER
 test: $(GO_SRC) $(LIBBPF_HEADERS) $(LIBBPF_OBJ) bpf test/profiler
-	$(go_env) go test -v $(shell go list ./... | grep -v "internal/pprof" | grep -v "pkg/profiler")
+	$(go_env) go test -run '[^TestIntegrationGRPC]' -v $(shell go list ./... | grep -v "internal/pprof" | grep -v "pkg/profiler")
```
Sylfrena, line 150: Is there a simpler way to do this? Ideally tags would be good but that would need changing all existing test files just to add the tag so used this hack instead.

kakkoyun: I think you can achieve this also just by tagging the integration test files. When running the unit tests, you can just exclude the tests that are tagged by integration.

brancz: I would exclude it in the same way as we exclude the `internal/pprof` and `pkg/profiler` packages, that's how I see it done in most places.

Makefile
```diff
@@ -258,3 +258,8 @@ dev/up: deploy/manifests
 .PHONY: dev/down
 dev/down:
 	source ./scripts/local-dev.sh && down
+
+.PHONY: test-e2e $(driver)
```
kakkoyun, line 262: I'm not sure about this Makefile syntax, have you double-checked this?
Especially with the `.PHONY` directive, I wouldn't expect this.

I'm assuming this would work like `make test-e2e driver=kvm2`, Would it work without specifying any args?

kakkoyun (commented):
This LGTM :+1: 

Only one nit about Make action. It could be my lack of knowledge. Could you check out my questions?

Makefile
Sylfrena: Yep, if no argument is passed, like `make test-e2e`, it just falls back to the default driver(virtualbox) specified in the `local-e2e.sh` script

kakkoyun (approved):

Makefile
kakkoyun: Since this hasn't used anywhere in the PR, I'm trusting your word on this ;)

PR #376, opened 2022-04-26
debuginfo: Use a faster go binary detection function

kakkoyun:
* Upgrade parca

xref: https://github.com/parca-dev/parca/issues/961

This solves the issue caused by lavishly allocated memory while testing whether the binaries are Go binaries or not.


metalmatze (approved):
Nice

PR #377, opened 2022-04-26
Rewrite eBPF profiler in Rust

kakkoyun:

> We use C for writing eBPF programs that live in kernel-space in Parca Agent. The act of writing C programs in an error-prone process. And the language itself has lots of ways to shoot oneself in the foot. Even though the eBPF verifier makes life a lot easier, it is still possible to write unsafe programs. Or when you make a mistake and the verifier doesn’t give the green light to the program, it is hard to pinpoint the mistakes that you made that can be preventable with simple compile-time checks (memory overflows, type-checking). The goal is to use Rust to write eBPF programs to provide more compile-time checks for safety and security.


yeya24:
Just curious any reason behind this? Is the golang agent not performant enough or it is simply because rust is cooler?

kakkoyun:
@yeya24 Hahaha, none of the above. I have updated the description for you. The gist of it, we would like to try Rust as an alternative to the C program we us in kernel-space. The idea is to guarantee more safety by preventing mistakes.

yeya24:
Got it. I made a mistake and you are just rewriting the kernel side ebpf collector. This makes sense to me. Thanks for the explanation.

kakkoyun:
cc @dave-tucker

kakkoyun:
PTAL @v-thakkar @Sylfrena @javierhonduco

bpf/cpu-profiler/src/main.rs
```diff
@@ -0,0 +1,120 @@
...
+// TODO(kakkoyun): It might be needed for BPF_F_USER_STACK!
+// #[allow(non_upper_case_globals)]
+// #[allow(non_snake_case)]
+// #[allow(non_camel_case_types)]
+// #[allow(dead_code)]
+// mod vmlinux;
+//
+// use vmlinux::task_struct;
+
+// TODO(kakkoyun): Learn about Rust structs.
```
v-thakkar: This should be constants.

.github/workflows/build.yml
```diff
@@ -11,18 +11,37 @@ jobs:
...
+      - name: Rust Cache
+        uses: Swatinem/rust-cache@v1.3.0
+
+      - name: Run eBPF toolchain setup
+        run: |
+          make -C bpf setup
+          cd bpf && cargo check
+
+      - name: Install llvm
+        run: sudo apt-get update -y && sudo apt-get install llvm libelf-dev
```
javierhonduco, line 44: nit: We are also installing `libelf-dev`, should we change the name so it's more descriptive?

.github/workflows/build.yml
```diff
@@ -11,18 +11,37 @@ jobs:
...
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          components: rustfmt, clippy
 
-      - name: Checkout
-        uses: actions/checkout@v3
+      - name: Set up Rust Nightly
+        uses: actions-rs/toolchain@v1
```
javierhonduco, line 30: This is so we install some components from 'stable' but rustc from nightly, right? In that case we should make sure we run something similar locally

.github/workflows/build.yml
```diff
@@ -31,7 +50,7 @@ jobs:
         run: make bpf
 
       - name: Set Env
-        run:  echo "BPF_OUT=$(pwd)/dist" >> $GITHUB_ENV
+        run: echo "BPF_OUT=$(pwd)/dist" >> $GITHUB_ENV
```
javierhonduco, line 53: TIL!

.github/workflows/build.yml
```diff
@@ -54,29 +73,44 @@ jobs:
...
+        uses: Swatinem/rust-cache@v1.3.0
+
+      - name: Run eBPF toolchain setup
+        run: |
+          make -C bpf setup
+          cd bpf && cargo check
+
       - name: Install llvm
-        run: sudo apt-get install llvm libelf-dev
+        run: sudo apt-get update -y && sudo apt-get install llvm libelf-dev
```
javierhonduco, line 106: Same as above

Dockerfile.dev
```diff
@@ -9,13 +9,19 @@ RUN apt-get update -y && \
       ln -s /usr/bin/clang-11 /usr/bin/clang && \
       ln -s /usr/bin/llc-11 /usr/bin/llc
 
+# Install Rust
+RUN curl https://sh.rustup.rs -sSf | bash -s -- -y
```
javierhonduco, line 13: Should we pin a Rust version here?

bpf/Makefile
```diff
@@ -0,0 +1,14 @@
+.PHONY: setup
+setup:
+	rustup install stable
+	rustup install nightly
```
javierhonduco, line 4: Same as above, should we pin a version?

bpf/cpu-profiler/src/main.rs
```diff
@@ -0,0 +1,97 @@
...
+        pid: ctx.tgid(),
+        user_stack_id: 0,
+        kernel_stack_id: 0,
+    };
+
+    match { STACK_TRACES.get_stackid(&ctx, BPF_F_USER_STACK.into()) } {
+        Ok(stack_id) => {
+            key.user_stack_id = stack_id as i32;
+        }
+        _ => {}
```
javierhonduco, line 63: Let's use the [`if let`](https://doc.rust-lang.org/rust-by-example/flow_control/if_let.html) pattern as we don't care about the other enum possibilities (in this case, `Err`)

bpf/cpu-profiler/src/main.rs
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    match { STACK_TRACES.get_stackid(&ctx, 0) } {
+        Ok(stack_id) => {
+            key.kernel_stack_id = stack_id as i32;
+        }
+        _ => {}
```
javierhonduco, line 70: Same as above

bpf/cpu-profiler/src/main.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    return try_update_count(&mut key);
+}
+
+#[inline(always)]
+unsafe fn try_update_count(key: &mut StackCountKey) -> Result<u32, u32> {
+    match { COUNTS.get(&key) } {
+        Some(count) => {
+            let val = AtomicU64::new(*count);
+            val.store(val.load(Ordering::SeqCst) + 1, Ordering::SeqCst);
```
javierhonduco, line 81: Perhaps `fetch_add` would be shorter and easier to read!

bpf/cpu-profiler/src/main.rs
```diff
@@ -0,0 +1,97 @@
...
+            match COUNTS.insert(&key, &val.load(Ordering::SeqCst), 0) {
+                Ok(_) => Ok(0),
+                Err(ret) => Err(ret as u32),
+            }
+        }
+        None => match COUNTS.insert(&key, &1, 0) {
```
javierhonduco, line 87: What's the last argument?

bpf/Makefile
```diff
@@ -0,0 +1,14 @@
...
+	cargo install bpf-linker
+	rustup component add rust-src --toolchain nightly-x86_64-unknown-linux-gnu
```
javierhonduco, line 6: Can we install `stable-aarch64-unknown-linux-gnu` if running on arm64?

javierhonduco (commented):
LGTM overall!! Currently testing it on my Fedora arm64 box, will follow-up as soon as I get some profiles

Dockerfile.dev
kakkoyun: This is rustup so it's ok to install latest!

.github/workflows/build.yml
kakkoyun: We have a make action to install these in a similar manner under /bpf

bpf/cpu-profiler/src/main.rs
kakkoyun: I will remove this to simplify.

bpf/cpu-profiler/src/main.rs
kakkoyun: I'll add a comment for it.

bpf/cpu-profiler/src/main.rs
```diff
@@ -0,0 +1,89 @@
...
+    if let Ok(stack_id) = STACK_TRACES.get_stackid(&ctx, 0) {
+        key.kernel_stack_id = stack_id as i32;
+    }
+
+    return try_update_count(&mut key);
+}
+
+#[inline(always)]
+unsafe fn try_update_count(key: &mut StackCountKey) -> Result<u32, u32> {
+    match { COUNTS.get(&key) } {
```
orhun, line 63: Hey, I'm just a random guy passing by, I was wondering if there was a specific reason why you used curly braces for `match` expressions?

kakkoyun: No, particular reason. Except being new in Rust :)

PR #378, opened 2022-04-28
.gitattributes: Mark some files as gen / 3rd party

javierhonduco:
So linguist doesn't tally them https://github.com/github/linguist/blob/master/docs/overrides.md

cc/ @kakkoyun

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+3rdparty/ linguist-vendored	
+vmlinux.h linguist-generated
+internal/pprof/ linguist-vendored
+internal/go/ linguist-vendored


.gitattributes
```diff
@@ -0,0 +1,3 @@
+3rdparty/ linguist-vendored	
+vmlinux.h linguist-generated
+internal/pprof/binutils/testdata/ linguist-vendored
```
kakkoyun, line 3:
```suggestion
internal/pprof/ linguist-vendored
internal/go linguist-vendored
```

I think this is more correct :)

kakkoyun (commented):
:rocket:

kakkoyun (approved):
:100: :heart: :rocket:

PR #387, opened 2022-05-04
ksym: Calculate cache hit rate

javierhonduco:
To help us understand the effectiveness of the Kernel symbolizer cache

**Test plan**

http://localhost:7071/metrics contains
```
parca_agent_profiler_kernel_symbolizer_cache_hit_rate{target="{__cgroup_path__=\"/sys/fs/cgroup/system.slice/docker-0cd1ac1dee8df9859b5dd10b794b38b235ef5745cedfde653896dd09264cf615.scope/\", node=\"test\", systemd_unit=\"docker-0cd1ac1dee8df9859b5dd10b794b38b235ef5745cedfde653896dd09264cf615.scope\"}"} 38.035961272475795
```

### DIFF
--- a/pkg/ksym/ksym.go
+++ b/pkg/ksym/ksym.go
@@ -16,6 +16,7 @@ package ksym
 import (
 	"bufio"
 	"errors"
+	"fmt"
 	"io/fs"
 	"os"
 	"sort"
@@ -32,6 +33,11 @@ import (
 
 var ErrFunctionNotFound = errors.New("kernel function not found")
 
+type CacheStats struct {
+	Hits  int
+	Total int
+}
+
 type Cache struct {
 	logger                log.Logger
 	fs                    fs.FS
@@ -39,6 +45,7 @@ type Cache struct {
 	lastCacheInvalidation time.Time
 	updateDuration        time.Duration
 	fastCache             map[uint64]string
+	Stats                 CacheStats
 	mtx                   *sync.RWMutex
 }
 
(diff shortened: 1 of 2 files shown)


brancz:
I would have expected this to be a counter and the `rate` would be calculated in Prometheus.

javierhonduco:
Good call! Thanks!

brancz:
lgtm

pkg/profiler/profiler.go
```diff
@@ -526,6 +536,11 @@ func (p *CgroupProfiler) profileLoop(ctx context.Context, captureTime time.Time)
 		level.Warn(p.logger).Log("msg", "failed to clean BPF maps", "err", err)
 	}
 
+	ksymCacheStats := p.ksymCache.Stats
+	level.Debug(p.logger).Log("msg", "Kernel symbol cache stats", "stats", ksymCacheStats.String())
+	p.metrics.ksymCacheHitRate.WithLabelValues("hits").Add(float64(ksymCacheStats.Hits))
+	p.metrics.ksymCacheHitRate.WithLabelValues("total").Add(float64(ksymCacheStats.Total))
```
kakkoyun, line 542: This is actually not correct regarding Prometheus conventions. It's like double counting. I would expect to see "hit", "miss" label values and if I would like to have the total I would just skip specifying the label value for the metric.

pkg/ksym/ksym.go
```diff
@@ -32,26 +33,41 @@ import (
 
 var ErrFunctionNotFound = errors.New("kernel function not found")
 
+type CacheStats struct {
```
kakkoyun, line 36: Rather than having this, I would prefer to have a metrics struct with the counters defined below.

kakkoyun (commented):
Sorry, I'm late to the game, but I have a couple of suggestions.

pkg/profiler/profiler.go
javierhonduco: I see, I will update the PR to submit the hits with the 'hits' label, and the total without labels. 

Never used Prometheus so I am not sure, but in the UI is it going to be clear that the 'total' is the one without a label?

pkg/ksym/ksym.go
javierhonduco: Are you referring to the counters created in `profiler.go` or something else? If so, I thought it would be a good separation of concerns to keep the stats here and create the Prometheus counters on the profiler pkg, but no strong opinions!

kakkoyun: What I meant was to have `metrics` struct in kysm package similar to one in profiler.go. And rather than having an additional struct for stats using counters directly in here. So we can pass a register to this type and handle everything separately. That's actually a pattern we use too often.

pkg/profiler/profiler.go
kakkoyun: You don't need to increase total without label. Labels are for slicing and dicing the a whole set, so a query without specifying a label will give you the total set out of the box. For example, when you have "hits" + "miss" = "total". With your proposed way, it would be  2 x "hits" = "total", so it's false.

Check this presentation, it would help https://promcon.io/2017-munich/slides/best-practices-and-beastly-pitfalls.pdf

PR #391, opened 2022-05-05
ksym: Move prometheus cache stats to the ksym package

javierhonduco:
Addressing feedback from https://github.com/parca-dev/parca-agent/pull/387

**Test plan**
```
parca_agent_profiler_kernel_symbolizer_cache_total{type="hits"} 639
parca_agent_profiler_kernel_symbolizer_cache_total{type="misses"} 580
```

### DIFF
--- a/pkg/ksym/ksym.go
+++ b/pkg/ksym/ksym.go
@@ -16,7 +16,6 @@ package ksym
 import (
 	"bufio"
 	"errors"
-	"fmt"
 	"io/fs"
 	"os"
 	"sort"
@@ -28,14 +27,16 @@ import (
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+
 	"github.com/parca-dev/parca-agent/pkg/hash"
 )
 
 var ErrFunctionNotFound = errors.New("kernel function not found")
 
-type CacheStats struct {
-	Hits  int
-	Total int
+type metrics struct {
+	ksymCacheHitRate *prometheus.CounterVec
 }
 
 type Cache struct {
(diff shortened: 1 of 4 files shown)


javierhonduco:
Oh, seems like a test spuriously crashed, taking a look

pkg/ksym/ksym.go
```diff
@@ -45,29 +49,34 @@ type Cache struct {
...
+				Help:        "Stats for the kernel symbolizer cache",
+				ConstLabels: map[string]string{"target": target.String()},
+			},
+			[]string{"type"},
+		),
+	}
 }
 
-func NewKsymCache(logger log.Logger) *Cache {
+func NewKsymCache(logger log.Logger, reg prometheus.Registerer, target model.LabelSet) *Cache {
```
kakkoyun, line 73: I think this is the perfect use case for https://github.com/prometheus/client_golang/blob/0222f88f4a941059843b49d9f56b42b9fea90d97/prometheus/wrap.go#L47-L52

Rather than passing target labels just wrap the register and pass it.

pkg/target/manager.go
```diff
@@ -60,7 +60,7 @@ func NewManager(
 		logger:            logger,
 		reg:               reg,
 		externalLabels:    externalLabels,
-		ksymCache:         ksym.NewKsymCache(logger),
+		ksymCache:         ksym.NewKsymCache(logger, reg, externalLabels),
```
brancz, line 63: rather than external labels, this should be the name of the service discovery config no?

javierhonduco: N00b question 😅 , but what would the name of the service discovery config be?

I might be wrong, but in the end, I don't think we really need any labels here, no?

kakkoyun: This is more accurate. These metrics are per manager.

javierhonduco: As far as I've seen, we only have one manager no? So we don't need labels to identify the manager (yet), no?

kakkoyun: I think so I need to double-check, in theory, there should only be one because systemd and kubernetes discovery, they're mutually exclusive

kakkoyun:
@javierhonduco How close is this to the finish line?

javierhonduco:
@kakkoyun forgot about this one, will take a look at it by the end of the day!

javierhonduco:
Will re-open this in some days, once system-wide lands

PR #399, opened 2022-05-11
Introduce spin locks around counter

kakkoyun:
cc @Sylfrena @v-thakkar

### DIFF
--- a/parca-agent.bpf.c
+++ b/parca-agent.bpf.c
@@ -64,9 +64,14 @@ typedef struct stack_count_key {
   int kernel_stack_id;
 } stack_count_key_t;
 
+struct counter {
+  u64 count;
+  struct bpf_spin_lock lock;
+} counter_t;
+
 /*================================ MAPS =====================================*/
 
-BPF_HASH(counts, stack_count_key_t, u64);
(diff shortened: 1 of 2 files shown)


kakkoyun:
https://lwn.net/Articles/779120/

> One significant added restriction is that BPF spinlocks cannot be used in tracing or socket-filter programs due to preemption-related issues. Those restrictions seem likely to be lifted in the future, but other types of BPF programs (including most networking-related programs) should be able to use BPF spinlocks once the feature goes upstream.

https://github.com/aquasecurity/libbpfgo/blob/aab8978ac89e2842886ad16193ae40124143011a/selftest/spinlocks/README.md

> In BTF annotated BPF maps, we can use a bpf_spin_lock to allow for atomic reads/updates of values in BPF maps.

 It is NOT supported in programs of type:

 - BPF_PROG_TYPE_KPROBE
 - BPF_PROG_TYPE_TRACEPOINT
 - BPF_PROG_TYPE_PERF_EVENT
 - BPF_PROG_TYPE_RAW_TRACEPOINT

So spinlocks are not allowed for our type of programs.

PR #403, opened 2022-05-13
Revert "profiler: Use BPF batch operations "

kakkoyun:
Reverts parca-dev/parca-agent#394

This feature requires a bump in the minimum required Linux version.

cc @javierhonduco  @v-thakkar @Sylfrena

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,4 +6,3 @@ Dockerfile
 /images
 /.github
 /tmp
-*.bpf.o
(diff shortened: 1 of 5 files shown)


javierhonduco:
LGTM!

Just for context for others, this works on our dev boxes which have more modern kernels, but not on mini-kube as @v-thakkar just checked.

We will re-add this feature once the mini-kube kernel is upgraded

PR #428, opened 2022-05-31
Add boot time checks for external dependencies

kakkoyun:
Signed-off-by: Kemal Akkoyun <kakkoyun@gmail.com>\

Thanks, @javierhonduco for the idea.

cc @v-thakkar 

### Example errors:

1. `level=error ts=2022-05-31T15:51:24.74057947Z caller=main.go:178 msg="failed to check whether external dependency is healthy" err="exit status 127" cmd=eu-strip output="eu-strip: error while loading shared libraries: libelf.so.1: cannot open shared object file: No such file or directory\n"`
2. `level=error ts=2022-05-31T15:14:27.640210588Z caller=main.go:168 msg="failed to find external dependency in the PATH; make sure it is installed and added to the PATH" cmd=objcopy`

### DIFF
--- a/cmd/parca-agent/main.go
+++ b/cmd/parca-agent/main.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"crypto/tls"
 	"errors"
@@ -25,6 +26,7 @@ import (
 	"net/http/pprof"
 	"net/url"
 	"os"
+	"os/exec"
 	"os/signal"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


cmd/parca-agent/main.go
```diff
@@ -158,6 +160,34 @@ func main() {
...
+
+				ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+				cmd := exec.CommandContext(ctx, c, "--help")
+				var stdout, stderr bytes.Buffer
+				cmd.Stdout = &stdout
+				cmd.Stderr = &stderr
+				if err := cmd.Run(); err != nil {
+					cancel()
+					level.Error(logger).Log(
+						"msg", "failed to check whether external dependency is healthy",
```
javierhonduco, line 184: Should we print the output of the binary that fails to have more context?

javierhonduco (approved):
LGTM

cmd/parca-agent/main.go
kakkoyun: :+1:

javierhonduco: Thanks!! ✨

PR #429, opened 2022-05-31
fix: set profile duration to real measurement

### LINKED ISSUE #375: Set duration to real measurement
The duration in a pprof profile is supposed to be the real duration, instead of the intended duration, so parca-agent should be setting the duration that capturing the profile actually took instead of the configured interval.

https://github.com/parca-dev/parca-agent/blob/80e9cb0df9ae987900b85f8e41150af7ade275cb/pkg/profiler/profiler.go#L556

maxbrunet:
Hello there, I am looking at the "good first issues" :wave:

I may be completely wrong, but I am trying to get to know the internals, so it looks like it is simply a matter of measuring `time.Since(captureTime)` in `buildProfile()`, set from here:

https://github.com/parca-dev/parca-agent/blob/4dcd283d602a97325cc2a5371634686196946fb5/pkg/profiler/profiler.go#L390-L391

The exact collection of the samples starting/finishing _slightly_ later/earlier, adding `start := time.Now()` before the BPF Maps loop:

https://github.com/parca-dev/parca-agent/blob/4dcd283d602a97325cc2a5371634686196946fb5/pkg/profiler/profiler.go#L415-L416

And measuring `duration := time.Since(start)` after exiting the loop, then pass it down to `buildProfile()` should also work.

The former is much simpler (1 line change), but I wonder how accurate we can/want/need to be

Closes #375

### DIFF
--- a/pkg/profiler/profiler.go
+++ b/pkg/profiler/profiler.go
@@ -572,7 +572,7 @@ func (p *CgroupProfiler) buildProfile(
 			Unit: "count",
 		}},
 		TimeNanos:     captureTime.UnixNano(),
-		DurationNanos: int64(p.profilingDuration),
+		DurationNanos: int64(time.Since(captureTime)),
 
 		// We sample at 100Hz, which is every 10 Million nanoseconds.
 		PeriodType: &profile.ValueType{
--- a/pkg/profiler/profiler.go
+++ b/pkg/profiler/profiler.go
@@ -204,9 +204,9 @@ type CgroupProfiler struct {
 	bpfMaps   *bpfMaps
 	byteOrder binary.ByteOrder
 
-	lastError          error
-	lastProfileTakenAt time.Time
-	nextProfileTakenAt time.Time
+	lastError                      error
+	lastSuccessfulProfileStartedAt time.Time
+	nextProfileStartedAt           time.Time
 
 	writeClient profilestorepb.ProfileStoreServiceClient
 	debugInfo   *debuginfo.DebugInfo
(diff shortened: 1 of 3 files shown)


brancz:
Conceptually this looks pretty close, we do need to initialize it at the very start before we start the profile loop otherwise the first profile duration is set to ~52years (current time-unix timestamp 0).

I just double-checked and `.TimeNanos` is supposed to be when the profile started, so `.TimeNanos` + `.DurationNanos` = when we wiped the BPF map. Meaning what you did now is correct.

I like that you kept the last profile taken at as we want that to continue to be 0 until a profile was actually taken successfully (it's often a good indicator that something is not working correctly). I would suggest to not return the time and just set it internally.

maxbrunet:
> we do need to initialize it at the very start

Indeed, it should now initialized `nextProfileTakenAt` right after attaching the eBPF program

> I would suggest to not return the time and just set it internally.

I have changed `nextProfileTakenAt` to be updated within `loopReport()` if no error occurred, if it is what you meant

I have not taken the time to set up my local environment yet, maybe the CI could give it a shot?

kakkoyun (approved):
LGTM :100: 

GIving @brancz the last approval, he is the expert on this :)

pkg/profiler/profiler.go
```diff
@@ -548,31 +559,34 @@ func (p *CgroupProfiler) profileLoop(ctx context.Context, captureTime time.Time)
 	return nil
 }
 
-func (p *CgroupProfiler) loopReport(lastProfileTakenAt time.Time, lastError error) {
+func (p *CgroupProfiler) loopReport(lastError error) {
 	p.mtx.Lock()
 	defer p.mtx.Unlock()
 
-	p.lastProfileTakenAt = lastProfileTakenAt
+	if lastError == nil {
```
brancz, line 566: I like this, but then I think we should rename the fields to `lastSuccessfulProfileStartedAt` and `nextProfileStartedAt`.

brancz (commented):
Last tiny nit, then I think this is ready to merge! Thank you so much for taking care of this!

kakkoyun (approved):

PR #431, opened 2022-06-02
[WIP WIP ] Fix the build for arm64 in CI

v-thakkar:
[ ] Check that libelf-dev works in GH actions
[ ] Check that BPF linker gets installed properly for arm64 


kakkoyun:
Wohoo :tada: All green :green_heart: Anything else left?

v-thakkar:
Yes, actually Aya still doesn't work for arm64: https://github.com/parca-dev/parca-agent/runs/6759705092?check_suite_focus=true#step:7:807 . I'm looking into it.

bpf/xtask/src/build_ebpf.rs
```diff
@@ -44,7 +44,7 @@ pub fn build_ebpf(opts: Options) -> Result<(), anyhow::Error> {
     let dir = PathBuf::from("cpu-profiler");
     let target = format!("--target={}", opts.target);
     let mut args = vec![
-        "+nightly",
+        "+nightly-2022-05-23",
```
kakkoyun, line 47: :+1:

v-thakkar:
I reported the bug in Aya and the fix has been merged now: https://github.com/aya-rs/aya/commit/ef3b029d26b1eeba3507c194ee24d6c1fcfc57e2 . For some reason, CI is not pulling the latest commit[ef3b029d2] from main. :disappointed: 

Also, we should use a stable tag - [aya-v0.11.0] once this fix is part of it.

v-thakkar:
Phew, this is finally done! @kakkoyun I've changed Dockerfile.dev as well to keep it in sync with the one we're using for CI. But let me know in case I've changed something that is required specifically for Tilt. Thanks

kakkoyun (approved):
LGTM :rocket: 

Awesome work @v-thakkar! I have seen that you put so much effort into this! Thanks a lot!



My only concern in here, and it's not specific to this PR, how we install LLVM. I think this is not reproducible. Because we don't use a snapshot repo. Do you happen to know a way to achieve this?

kakkoyun:
We don't need to handle the issue in this PR. We should handle it in subsequent PRs before the next major release.

PR #432, opened 2022-06-03
bpf: Pin the nightly in Makefile

v-thakkar:


v-thakkar:
Closing this as I think it's better to include this change as a separate commit in #431.

PR #433, opened 2022-06-03
chore: make development environment more portable

maxbrunet:
* Use `/usr/bin/env bash` instead of `/bin/bash`, `bash`'s location can differ from a system to another. For example, on MacOS, if it is installed with `brew`, it is `/usr/local/bin/bash`. And I use NixOS, it is `/run/current-system/sw/bin/bash`.
* Make Rust toolchain consistent in the whole repository using `rust-toolchain.toml`. It was set at multiple places to multiple values (nightly, nightly-date-pinned, default), it should be easier to manage.
* I noticed Rust was not mentioned in `CONTRIBUTING.md`, so I added it
* Add [`.editorconfig`](https://editorconfig.org/) for consistent config across IDEs

I can suggest to add a [`shell.nix` file](
https://nixos.wiki/wiki/Development_environment_with_nix-shell):

```nix
let
  unstable = import (fetchTarball https://nixos.org/channels/nixos-unstable/nixexprs.tar.xz) { };
in
{ pkgs ? import <nixpkgs> { } }:

pkgs.mkShell {
  nativeBuildInputs = with pkgs; [
    clang
    glibc
    glibc.static
    gnumake
    libbpf
    libelf
    llvm
    rustup
    zlib
    zlib.static
  ] ++
  (with unstable.pkgs; [
    go_1_18
  ]);
}
```

I started to use [`nix`](https://nixos.org/download.html) only recently, but it may be possible to codify most setup instructions from `CONTRIBUTING.md` into it, while keeping system-level dependencies to a minimum


.github/workflows/build.yml
```diff
@@ -19,12 +19,10 @@ jobs:
         with:
           go-version: ^1.18
 
-      - name: Set up Rust Nightly
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: nightly-2022-05-23
-          components: rustfmt, clippy
+      - name: Set up Rust
```
kakkoyun, line 22: The nightly is the one that needs to be used. Also, it should be a date pinned one, so that we can preserve the reproducible build promise we have.

See also https://github.com/parca-dev/parca-agent/pull/431/files

Sorry for the confusion here, we just merge the Rust pipeline and not everything is stabilized or documented yet.

.github/workflows/codeql-analysis.yml
```diff
@@ -46,12 +46,10 @@ jobs:
         with:
           go-version: ^1.18
 
-      - name: Set up Rust Nightly
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: nightly-2022-05-23
-          components: rustfmt, clippy
+      - name: Set up Rust
```
kakkoyun, line 49: ditto.

.github/workflows/docs.yml
```diff
@@ -18,12 +18,10 @@ jobs:
         with:
           go-version: ^1.18
 
-      - name: Set up Rust Nightly
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: nightly-2022-05-23
-          components: rustfmt, clippy
+      - name: Set up Rust
```
kakkoyun, line 21: ditto.

.github/workflows/release.yml
```diff
@@ -23,12 +23,10 @@ jobs:
         with:
           go-version: ^1.18
 
-      - name: Set up Rust Nightly
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: nightly-2022-05-23
-          components: rustfmt, clippy
+      - name: Set up Rust
```
kakkoyun, line 26: ditto.

Dockerfile
```diff
@@ -21,13 +21,14 @@ RUN apt-get -o Acquire::Check-Valid-Until="false" update -y && \
...
       ln -s /usr/bin/llc-14 /usr/bin/llc
 
-# Install Rust Nightly
-RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --default-toolchain none -y
-ENV PATH="/root/.cargo/bin:${PATH}"
-RUN rustup toolchain install nightly-2022-05-23 --allow-downgrade --profile=minimal --component clippy
-
 WORKDIR /parca-agent
 
+# Install Rust
```
kakkoyun, line 26: ditto.

kakkoyun (changes requested):
Awesome PR! Thanks for handling these.

The nightly is the one that needs to be used. Also, it should be a date pinned one, so that we can preserve the reproducible build promise we have.

See also https://github.com/parca-dev/parca-agent/pull/431/files

Sorry for the confusion here, we just merge the Rust pipeline and not everything is stabilized or documented yet.

And I'm more than happy to accept the nix file if you're up to maintain that file :D

.github/workflows/build.yml
maxbrunet: > The nightly is the one that needs to be used

Yes, the date-pinned nightly channel is set in the root `rust-toolchain.toml` file, the `rustup` commands will respect it. Sorry if it was not clear, it got lost in all the edits I did.

You can verify in the pipeline:

```console
info: syncing channel updates for 'nightly-2022-05-23-x86_64-unknown-linux-gnu'
info: latest update on 2022-05-23, rust version 1.63.0-nightly (b2eed72a6 2022-05-22)
```

> Sorry for the confusion here, we just merge the Rust pipeline and not everything is stabilized or documented yet.

No worries, I noticed, quite happy you made the jump from C to Rust!

maxbrunet:
> The nightly is the one that needs to be used. Also, it should be a date pinned one, so that we can preserve the reproducible build promise we have.

I agree, it is now set from a single place, the `rust-toolchain.toml` file, it avoids specifying it on every command and makes updates easier.

> And I'm more than happy to accept the nix file if you're up to maintain that file :D

I'll raise another PR and we can discuss there :+1:

kakkoyun:
> I agree, it is now set from a single place, the `rust-toolchain.toml` file, it avoids specifying it on every command and makes updates easier.

Oh, now I see. So sorry, it's my bad. I have overlooked it. :see_no_evil:

kakkoyun (approved):
LGTM :taco:

kakkoyun:
I'm merging this. @v-thakkar You might want to rebase https://github.com/parca-dev/parca-agent/pull/431
I love the idea of setting the rust version in a single place rather than a shotgun surgery.

Issue #420, opened 2022-05-26
symbolizer: In some occasions we are uploading broken debug files


javierhonduco:
Seems like some of the debug files we are uploading to Parca are either empty or wrong. Opening this issue so I don't forget to take a closer look at it. 

> ```level=error name=parca ts=2022-05-26T10:17:06.842259165Z caller=symbolizer.go:59 component=symbolizer msg="symbolization attempt finished with errors" err="2 errors occurred:\n\t* storage symbolization request failed: failed to check for DWARF: failed to open elf: bad magic number '[0 0 0 0]' in record at byte 0x0\n\t* storage symbolization request failed: failed to check for DWARF: failed to open elf: bad magic number '[0 0 0 0]' in record at byte 0x0\n\n"```

kakkoyun:
Probably better to move this issue to the server-side. It seems like we send badly formated files to the server-side.

javierhonduco:
Let me open another issue in the backend, as I don't think we should be sending nor accepting broken debug files

brancz:
Agreed, the server should validate, but the client shouldn't even attempt to upload faulty data.

PR #444, opened 2022-06-08
debuginfo: Fix race when extracting debug info files

javierhonduco:
fixes: https://github.com/parca-dev/parca-agent/issues/420

## Context
When generating profiles, we might need to extract and upload the debug information present in the binaries so we can symbolise the addresses with DWARF. We do this here:

https://github.com/parca-dev/parca-agent/blob/5f8f1525173e34125a101c374191e1bd0c031b16/pkg/profiler/profiler.go#L612-L622

When profiling many processes, we might have a race condition. For example, the same executables (libc) might show up in multiple processes, and we will attempt to upload them all. 

## Why is this a problem?

The explanation of it is in the code comment below. The result is that we end up with corrupted debug files, which this PR fixes!

It's also a performance issue, as we are repeating lots of work (extracting binaries, attempting to send them) that unnecesarily burn CPU cycles both in the server and in the agent.

## Test plan

Before this PR we generated some empty or corrupt files, now it's clean:

```
[javierhonduco@localhost-live parca]$ file  tmp/*/debuginfo | grep ELF | wc -l
336
[javierhonduco@localhost-live parca]$ file  tmp/*/debuginfo | grep -v ELF
[javierhonduco@localhost-live parca]$ echo $?
0
```

Simbolisation still works:

image

### DIFF
--- a/pkg/debuginfo/debuginfo.go
+++ b/pkg/debuginfo/debuginfo.go
@@ -38,6 +38,8 @@ type DebugInfo struct {
 	existsCache        cache.Cache
 	debugInfoFileCache cache.Cache
 
+	uploadingCache cache.Cache
+
 	*Extractor
 	*Uploader
 	*Finder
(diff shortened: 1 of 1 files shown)


javierhonduco:
Linter

pkg/debuginfo/debuginfo.go
```diff
@@ -53,12 +57,49 @@ func New(logger log.Logger, client Client, tmp string) *DebugInfo {
...
+// and once all the writers are done, the debug file looks is an ELF
+// with the correct bytes.
+//
+// However I don't believe there's any guarantees on this, so the
+// files aren't getting corrupted most of the time by sheer luck.
+//
+// These two helpers make sure that we don't try to extract + upload
+// the same buildID concurrently.
+func (di *DebugInfo) alreadyUploading(buildID string) bool {
+	di.uploadingCacheMutex.Lock()
```
brancz, line 85: Do we really need these locking mechanisms? The `cache.Cache` looks to me as if it's safe for concurrent use.

brancz:
Very nice find!

pkg/debuginfo/debuginfo.go
javierhonduco: Good catch! https://github.com/goburrow/cache/issues/22#issuecomment-791784927

kakkoyun (approved):
Really great find! And amazing work to fix it.
Thanks a lot @javierhonduco!

PR #447, opened 2022-06-09
chore: configure entire toolchain in rust-toolchain.toml

maxbrunet:
I figured we can simplify further, the toolchain can be fully configured from `rust-toolchain.toml`:
https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file

And yes, oddly `rustup show` installs the toolchain :upside_down_face:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,9 +20,7 @@ jobs:
           go-version: ^1.18
 
       - name: Set up Rust
-        run: |
-          rustup set profile minimal
-          rustup component add rustfmt clippy
+        run: rustup show
 
       - name: Rust Cache
         uses: Swatinem/rust-cache@v1.4.0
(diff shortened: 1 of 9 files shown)


.github/workflows/build.yml
```diff
@@ -20,9 +20,7 @@ jobs:
           go-version: ^1.18
 
       - name: Set up Rust
-        run: |
-          rustup set profile minimal
-          rustup component add rustfmt clippy
+        run: rustup show
```
kakkoyun, line 23: You explained it in the PR description, however, it would be awesome to add a comment here that explains this installs rust using rust-toolchain.toml.

We are still bunch of newbies in the rust ecosystem :D

kakkoyun (commented):
Great work! Even less redundancy! :rocket: 

I only have one nit, and also you need the rebase the branch.

kakkoyun (approved):
LGTM

PR #448, opened 2022-06-09
chore: add shell.nix file for development

maxbrunet:
This adds a `shell.nix` file to manage the development dependencies, this is my first attempt at using it in project. The idea is to contain as much as possible the dependencies within the repository and avoid spilling on the system or even the user's home.

I added instructions about how to use it in the `CONTRIBUTING.md`, including basics on how to install Nix even though it is out of the scope of `parca-agent`.

It sets a few environment variables to keep things within the repository like Go and Rust binaries, we can add more or remove if that causes troubles.

Once `nix-shell` has been run and with #447 to configure the Rust toolchain, only a couple more commands are needs to build the agent, basically you are ready to work with:

```console
$ nix-shell
[nix-shell:/path/to/parca-agent]$ cargo install bpf-linker # shellHook seems to run before the dependencies like glibc are available
[nix-shell:/path/to/parca-agent]$ make
```

`direnv` can also be use to enter the shell automatically and use your own shell if different from `bash`.

I have only tested the agent local build, not the full parca setup. On NixOS and an Ubuntu VM (both package install and Nix installer script)

Some references:
* https://nixos.wiki/wiki/Development_environment_with_nix-shell
* https://nixos.org/manual/nix/stable/expressions/expression-syntax.html
* https://nixos.org/manual/nixpkgs/stable/#sec-pkgs-mkShell
* https://nixos.wiki/wiki/Rust#Installation_via_rustup
* https://nixos.wiki/wiki/Go#Compile_go_program_with_static_compile_flag

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,31 @@
+{ pkgs ? import <nixpkgs> { } }:
+
+pkgs.mkShell rec {
+  name = "parca-agent";
+
+  packages = with pkgs; [
+    clang
+    gnumake
+    go_1_18
+    kubectl
+    libbpf
+    libelf
+    llvm
+    minikube
+    nodejs-16_x
+    rustup
+    zlib.static
+  ] ++ (lib.optional stdenv.isLinux [ glibc glibc.static ]);
+
+  shellHook = ''
+    export PATH="''${PROJECT_ROOT}/bin:''${CARGO_HOME}/bin:''${PATH}"
+    rustup show >/dev/null
+  '';
+
+  PROJECT_ROOT = builtins.toString ./.;
+  CARGO_HOME = "${PROJECT_ROOT}/tmp/cargo";
+  GOBIN = "${PROJECT_ROOT}/bin";
+  RUSTUP_HOME = "${PROJECT_ROOT}/tmp/rustup";
+  RUST_BACKTRACE = 1;
+}
+
--- a/shell.nix
+++ b/shell.nix
@@ -12,7 +12,6 @@ pkgs.mkShell rec {
     libelf
     llvm
     minikube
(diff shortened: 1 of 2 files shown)


shell.nix
```diff
@@ -0,0 +1,31 @@
...
+  packages = with pkgs; [
+    clang
+    gnumake
+    go_1_18
+    kubectl
+    libbpf
+    libelf
+    llvm
+    minikube
+    nodejs-16_x
```
kakkoyun, line 15: I think we don't need nodejs in this repo. I was that it's mentioned in the contributing.md above, we should also remove it. I think it's a residue from parca repo.

kakkoyun (commented):
Awesome! Thanks for contributing this.
One more reason to try out the nix ecosystem :)

Just one tiny thing to change.

shell.nix
maxbrunet: It is probably because there are steps to build `parca` here:

https://github.com/parca-dev/parca-agent/blob/51aee7a5fb288d9a1db111ebfc395e213c15e4a4/CONTRIBUTING.md#L77-L79

Should we remove everything specific to parca and link its `CONTRIBUTING.md` instead?

kakkoyun: Yes, let's do that. Do you want to handle it? Otherwise, I'd be happy to do it.

maxbrunet: Done!

maxbrunet:
I have removed the Nix instructions, I have not used Nix outside NixOS myself yet, so I prefer avoiding misguiding the users and referring them to the official website or letting them do their own research.

kakkoyun (approved):
LGTM :100:

PR #454, opened 2022-06-09
debuginfo: Add pure Go ELF Writer to extract debug information

kakkoyun:

## Test Plan

- [x] `sudo ./scripts/local-run.sh`
- [x]  `make dev/up && tilt up`
  - Visual check

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -122,16 +122,22 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-go-
 
+      - name: Build BPF
+        run: make bpf
+
+      - name: Build
+        run: make build
+
       - name: Test
-        run: make bpf test vet lint
+        run: make test
 
       - name: Format
         run: |
           make format
           git diff --exit-code
 
-      - name: Build
-        run: make build
+      - name: Vet
+        run: make vet lint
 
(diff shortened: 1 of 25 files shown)


brancz:
This is very very nice! I’m curious, do you think it would be possible to instead of writing the debuginfos to a separate file first to immediately write it to the uploader? We’ve seen so many issues with the agent crashing nodes by over utilizing the empty dir volume in Kubernetes that I’ve been wondering if we can avoid it entirely. Nothing to necessarily do in this PR, but food for thought maybe.

kakkoyun:
> This is very very nice! I’m curious, do you think it would be possible to instead of writing the debuginfos to a separate file first to immediately write it to the uploader? We’ve seen so many issues with the agent crashing nodes by over utilizing the empty dir volume in Kubernetes that I’ve been wondering if we can avoid it entirely. Nothing to necessarily do in this PR, but food for thought maybe.

@brancz This is totally possible. And it's actually something I plan. It just requires some changes in existing architecture, and this PR is already too large. I'll send consecutive PR to remove the need for interim files.

brancz:
Sounds good!

v-thakkar:
Really wonderful work! Having an in-house package will be super helpful in the future too. So I'm excited about it. Looking good :clap:

PR #461, opened 2022-06-14
build(deps): use llvm-14 from Debian snapshot

### LINKED ISSUE #449: chore: Ensure LLVM installs are reproducible
cc @v-thakkar

maxbrunet:
This is a suggestion. apt.llvm.org only keeps the latest build.

This uses the `sid` (`unstable`) channel from the Debian snapshot to get the LLVM 14 toolchain.
`bookworm` (`testing`) should also work depending on how stable or fast fixes are desired.

https://wiki.debian.org/DebianUnstable
https://wiki.debian.org/DebianTesting

It does get hard to tell which channel provides which package due to how they depend on each other, but unstable/testing should get the lowest priority if I understand correctly: https://manpages.debian.org/bullseye/apt/apt_preferences.5.en.html#APT's_Default_Priority_Assignments

Closes #449

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,11 +12,8 @@ deb http://snapshot.debian.org/archive/debian/20220420T025302Z bullseye main\n\
 deb http://snapshot.debian.org/archive/debian/20220420T025302Z bullseye-updates main\n\
 deb http://snapshot.debian.org/archive/debian/20220420T025302Z bullseye-backports main\n\
 deb http://snapshot.debian.org/archive/debian-security/20220420T025302Z bullseye-security main\n\
+deb http://snapshot.debian.org/archive/debian/20220420T025302Z sid main\n\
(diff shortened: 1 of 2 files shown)


brancz:
If I'm not mistaken I understood we actually need the latest llvm14 version, not just any llvm14. @v-thakkar we had this conversation recently I think?

kakkoyun (approved):
This actually looks good to me. Everything works as expected when I have checked the container builds.

If I'm not mistaken, we just need LLVM 14, but an unreleased (if it hasn't already) tip of the master version of aya and aya/bpf-linker.

I'll wait for @v-thakkar's confirmation before merging it, though.

v-thakkar:
Yes, I just checked the container build and it seems to be working for arm64 build. I was unaware of the unstable channels. So this is a much better change. Thanks @maxbrunet. :clap:

PR #465, opened 2022-06-15
build: fix silent container build failure and lint

### LINKED ISSUE #462: chore: Improve container build script
At the moment container build script doesn't fail when arm64 build fails to pass. We should add a check there so that we can continue to have arm images.

maxbrunet:
* Fix `set +euxo pipefail` --> `set -eux pipefail`
* Pin manifest list digest of base images directly in `Dockerfile` (you can use `regctl manifest digest --require-list IMAGE_NAME` to find it), instead of passing individual arch digest via args, this should make updates simpler, and Dependabot may be able to do PRs, if not, I highly recommend [Renovate](https://github.com/apps/renovate) which [can do a lot more](https://docs.renovatebot.com/docker/#digest-pinning)
* Use [Buildkit's `TARGETARCH` ARG](https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope) instead of `ARCH`, it standardize things a little and makes things easier if you happen to build with `docker-buildx`
* Lint with `shellcheck` and `hadolint`
* Sync `Dockerfile` and `Dockerfile.dev` as much as possible, to ensure nothing was missed, it also helps with the cache if you build one after the other

Closes #462

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,53 +1,66 @@
-ARG GOLANG_BASE
-ARG DEBIAN_BASE
-
-FROM ${GOLANG_BASE} as build
+FROM docker.io/golang:1.18.1-bullseye@sha256:3b1a72af045ad0fff9fe8e00736baae76d70ff51325ac5bb814fe4754044b972 AS build
 
 # tag=1.24.3
 ARG RUSTUP_VERSION=ce5817a94ac372804babe32626ba7fd2d5e1b6ac
 
 # For more information about the snapshots, see: https://snapshot.debian.org/
-RUN echo "\
+RUN printf '\
 deb http://snapshot.debian.org/archive/debian/20220420T025302Z bullseye main\n\
 deb http://snapshot.debian.org/archive/debian/20220420T025302Z bullseye-updates main\n\
 deb http://snapshot.debian.org/archive/debian/20220420T025302Z bullseye-backports main\n\
 deb http://snapshot.debian.org/archive/debian-security/20220420T025302Z bullseye-security main\n\
 deb http://snapshot.debian.org/archive/debian/20220420T025302Z sid main\n\
-" > /etc/apt/sources.list
+' > /etc/apt/sources.list
 
-# NOTICE: -o Acquire::Check-Valid-Until="false" added as a mitigation, see https://github.com/parca-dev/parca-agent/issues/10 for further details.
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,53 +1,66 @@
-ARG GOLANG_BASE
-ARG DEBIAN_BASE
-
-FROM ${GOLANG_BASE} as build
+FROM docker.io/golang:1.18.1-bullseye@sha256:3b1a72af045ad0fff9fe8e00736baae76d70ff51325ac5bb814fe4754044b972 AS build
```
kakkoyun, line 1: This base image is different from the one that we used in the script.
I assume you haven't commented on the data as we did before. We want to let dependabot or renovate to upgrade this regularly.

No action here. I just want to put a comment for future reader.

Dockerfile
```diff
@@ -1,53 +1,66 @@
...
+# SHELL is not supported for OCI image format
+# https://github.com/containers/buildah/blob/v1.26.1/config.go#L366-L377
+# hadolint ignore=DL4006
 RUN curl --proto '=https' --tlsv1.2 -sSf "https://raw.githubusercontent.com/rust-lang/rustup/${RUSTUP_VERSION}/rustup-init.sh" \
     | sh -s -- --default-toolchain none -y
 ENV PATH="/root/.cargo/bin:${PATH}"
 RUN rustup show
 
-ARG ARCH
+ARG TARGETARCH=amd64
```
kakkoyun, line 45: Magical :open_mouth: TIL.

For the ones who also didn't know about it https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

And Podman supports these ARGs and flags.

kakkoyun (approved):
Amazing PR! Thanks for improving these @maxbrunet.
I really appreciate it. If you have free cycles, feel free to do similar things in Parca as well :hugs: 
Tag me for the reviews.

We should definitely need those shell checks in our CI.
And I'll check out Renovate!

kakkoyun:
> * Lint with `shellcheck` and `hadolint`

We should definitely add these checks both in parca and parca-agent!

maxbrunet:
> And I'll check out Renovate!

You can find examples of how I use it in my personal repos:
https://github.com/maxbrunet/bitbucket-semantic-pull-requests
https://github.com/maxbrunet/prometheus-elasticache-sd

They maintain themselves

PR #468, opened 2022-06-16
chore(nix-shell): fix llvm

maxbrunet:
Since `bpf-linker` is built with `--features system-llvm`, LLVM 14 needs to be setup properly in `nix-shell` as well. 

~Blocked by NixOS/nixpkgs#148117 (https://nixpk.gs/pr-tracker.html?pr=178007)~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,8 +48,9 @@ OUT_DOCKER ?= ghcr.io/parca-dev/parca-agent
 DOCKER_BUILDER ?= parca-dev/cross-builder
 
 LIBBPF_SRC := 3rdparty/libbpf/src
-LIBBPF_HEADERS := $(OUT_DIR)/libbpf/$(ARCH)/usr/include
-LIBBPF_OBJ := $(OUT_DIR)/libbpf/$(ARCH)/libbpf.a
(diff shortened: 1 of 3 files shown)


maxbrunet:
This is ready for review, upstream has been fixed! (and `bpf-linker` requires `xml2` as well)

It also fixes the `CGO` flags for the `build-dyn` target:

```
runtime/cgo
# runtime/cgo
/nix/store/rq6bh3qfrqnyqwik0w3q6w180zg3w2pa-binutils-2.38/bin/ld: cannot find -lbpf: No such file or directory
clang-14: error: linker command failed with exit code 1 (use -v to see invocation)
```

```
github.com/aquasecurity/libbpfgo
# github.com/aquasecurity/libbpfgo
../../../../.local/share/go/pkg/mod/github.com/aquasecurity/libbpfgo@v0.3.0-libbpf-0.8.0/libbpfgo.go:10:10: fatal error: 'bpf/bpf.h' file not found
#include <bpf/bpf.h>
         ^~~~~~~~~~~
1 error generated.
```

maxbrunet:
Just found out that `cargo install` does not use the crate's lock file by default :hushed:, so I added `--locked`, it will avoid suprises

kakkoyun (approved):

Issue #146, opened 2021-11-29
Explain further what "Show Profile" does


brancz:
This feature originated from a debug mechanism for me to understand profiling data without a storage backend. It turns out people like this to understand what is going on under the hood, but for that, it is still quite raw of a feature and needs some more explanation (also why clicking show profiles can take up to 10 seconds to display anything). Beyond that, the actual showing of the profile could use some explanation of what the user is looking at.

PR #474, opened 2022-06-17
fix(statuspage): improve display of next profile

maxbrunet:
* Use next profile start time instead of last. Last has become less relevant (goes up to 2x duration) and shows 52 years for the first duration
* Change `Show Profile` into `Show Next Profile`
* Add `title` to link (tooltip) informing it may take up to the profiling duration to display
* Trim white-spaces
* Change `Download Pprof` into `Download Next Pprof`, add tooltip as well
* Calculate query timeout based on `--profiling-duration` flag
* Fix `profileListener.NextMatchingProfile()` (it was always returning the first profile)

image

Related to #146

### DIFF
--- a/cmd/parca-agent/main.go
+++ b/cmd/parca-agent/main.go
@@ -295,20 +295,21 @@ func main() {
 				return
 			}
 
-			// We profile every 10 seconds so leaving 1s wiggle room. If after
-			// 11s no profile has matched, then there is very likely no
+			// We profile every ProfilingDuration so leaving 1s wiggle room. If after
+			// ProfilingDuration+1s no profile has matched, then there is very likely no
 			// profiler running that matches the label-set.
-			ctx, cancel := context.WithTimeout(ctx, time.Second*11)
+			timeout := flags.ProfilingDuration + time.Second
+			ctx, cancel := context.WithTimeout(ctx, timeout)
 			defer cancel()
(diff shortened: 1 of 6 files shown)


brancz:
Very nice!

> Use next profile start time instead of last. Last has become less relevant (goes up to 2x duration) and shows 52 years for the first duration

If I'm not mistaken, the start is initialized and will be shown regardless of whether profiling succeeds. This has been a useful tool to identify that we've never successfully finished a profiling loop. If we change this I would like us to have some other indicator that tells us this at first glance.

kakkoyun (commented):
Thanks for showing some love to this page. It's been a while since the last time we touched it.

Happy to accept it when we address Frederic's concerns.

maxbrunet:
> If we change this I would like us to have some other indicator that tells us this at first glance.

`NextStartedAgo` will also show 52y ago if there are issues setting up the BPF program, and it will go beyond `profileDuration` if collecting samples fails

brancz:
Oh ok, I was misremembering the situation then. Awesome! Then I’m happy with this!

leaving merging to @kakkoyun as I think he’s in the process of preparing a new release.

kakkoyun:
I'm happy to merge this. I haven't tagged the new release yet. Still working on a few details.

Issue #331, opened 2022-04-07
Make sure github actions get triggered depending on correct context


kakkoyun:
- jsonnet  => jsonnet action
- go => go related actions and container builds
- no container build action runs on documentation (or the files won't end up in the container)

importhuman:
Hey @kakkoyun, can I pick this up? Also, any suggestions to try this locally instead of pushing commits to ensure it works?

kakkoyun:
Completed by #331

PR #476, opened 2022-06-19
Update Github actions for jsonnet and Go (#331)

importhuman:
Did not understand this point in the original issue:

> no container build action runs on documentation (or the files won't end up in the container)

Is this saying that currently, the `docs.yml` action is not triggered for any file changes, including documentation, so we should add the paths, or something else?

Regarding the new actions in the PR, I based them off of the existing actions in the Parca repo. Removed any Node.js jobs, and changed the path in the workflow from `parca` to `parca-agent` based on the Dockerfile. Please let me know if other changes are required.


kakkoyun (commented):
Definitely, in the right direction. We need to refine things.

kakkoyun:
> Is this saying that currently, the `docs.yml` action is not triggered for any file changes, including documentation, so we should add the paths, or something else?

We can run the docs action whenever something changes in the `cmd/parca-agent/main.go` it is for generating flag documentation on the README.md

kakkoyun:
And if we were to separate go actions we should add the equivalent rust actions.

kakkoyun:
@importhuman Basically, we should start with adding event filters for whatever we have as workflows.

Do you still want to take care of this?

importhuman:
Yes, I will provide an update if I'm unable to work on it today, in case this is urgent.

kakkoyun:
> Yes, I will provide an update if I'm unable to work on it today, in case this is urgent.

Noop, it's not urgent. I have just realized we still have the issue :)

importhuman:
Removed the `go-build-test.yml` file, updated filter paths in docs and build workflows.

maxbrunet (commented):
You can remove `.github/workflows/go-lint.yml` too, linting happens as part of the `build` workflow, and Go is not needed in the `container` workflow, it's setup within container :)

.github/workflows/container.yml
```diff
@@ -3,8 +3,22 @@ name: Container
...
+      - .github/workflows/container.yml
   pull_request:
     branches: [ main ]
+    paths:
+      - '**.go'
+      - 'go.mod'
+      - 'go.sum'
+      - Dockerfile*
+      - .dockerignore
+      - .github/workflows/container.yml
```
maxbrunet, line 21: These paths look good to me, you might need to had:

* the Rust files (.rs, cargo, rust-toolchain.toml) for the BPF program
* the `3rdparty` directory for the libbpf Git submodule
* the Makefile

.github/workflows/container.yml
```diff
@@ -16,7 +30,18 @@ jobs:
       contents: read
     steps:
       - name: Check out the code
-        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # tag=v3.0.2
+        uses: actions/checkout@v3
```
kakkoyun, line 33: Please keep the existing specific version.

.github/workflows/go-lint.yml
```diff
@@ -0,0 +1,41 @@
+name: golangci-lint
```
kakkoyun, line 1: No need to add this workflow. It's redundant since we cover these in the build.
Agent workflows are simpler compared to Parca itself. Let's keep it so for now.

kakkoyun (commented):
Looking great! A couple of more refinements and we're done. Good job :+1:

importhuman:
Not sure about the cargo files, let me know if other changes are required as well.

kakkoyun (approved):
LGTM :100: 

Thanks a lot for handling this.

.github/workflows/build.yml
```diff
@@ -3,8 +3,18 @@ name: Build
...
+      - 'go.mod'
+      - 'go.sum'
+      - .github/workflows/build.yml
   pull_request:
     branches: [ main ]
+    paths:
+      - '**.go'
+      - 'go.mod'
+      - 'go.sum'
+      - .github/workflows/build.yml
```
maxbrunet, line 17: Rust/Cargo/BPF would need to trigger the build workflow as well

kakkoyun: I'll handle it in #510

PR #485, opened 2022-06-20
chore: prepare dependencies for Renovate

maxbrunet:
This basically ensures all dependencies are pinned, Renovate is able to do it in some cases, but I prefer to do it myself to make sure nothing is missed.

This also adds the "renovate" comments for the custom regex

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM docker.io/golang:1.18.3-bullseye@sha256:db42e4bb1a7f32da1ec430906769dbbabe9f1868bd4170751e4923f1b8948a45 AS build
 
-# tag=1.24.3
-ARG RUSTUP_VERSION=ce5817a94ac372804babe32626ba7fd2d5e1b6ac
+# renovate: datasource:github-releases depName=rust-lang/rustup
+ARG RUSTUP_VERSION=1.24.3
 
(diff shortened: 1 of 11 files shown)


Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM docker.io/golang:1.18.3-bullseye@sha256:db42e4bb1a7f32da1ec430906769dbbabe9f1868bd4170751e4923f1b8948a45 AS build
 
-# tag=1.24.3
-ARG RUSTUP_VERSION=ce5817a94ac372804babe32626ba7fd2d5e1b6ac
+# renovate: datasource:github-releases depName=rust-lang/rustup
+ARG RUSTUP_VERSION=1.24.3
```
maxbrunet, line 4: I have stopped using the commit sha for simplicity, we'll see if we work something out to add it back later

kakkoyun: I don't know how nit-picky we should be with these. Git tags (GitHub releases) are mutable. In theory, this could break reproducible builds. We should pin everything in an immutable way if it's possible.

maxbrunet: You are right, I suggest we open an issue, and get back to it asap

kakkoyun: Sounds good to me.

Issue #445, opened 2022-06-08
Add race detector


javierhonduco:
I am not sure if [Golang's race detector](https://go.dev/doc/articles/race_detector) would have caught https://github.com/parca-dev/parca-agent/pull/444, I will try in 2 weeks, but I think we should enable it in dev and test:
- [ ] Add a flag to enable the race detector in developement (similar to https://github.com/parca-dev/parca-agent/pull/370)
- [ ] Run tests and integration tests with the race detector on, always

Related issue: https://github.com/parca-dev/parca-agent/issues/368

Issue #463, opened 2022-06-15
More e2e tests


v-thakkar:
Let's use this issue to keep track of things we want to add in e2e tests once basic support is merged.
- [ ] Add memory sanitizers (see #368)
- [ ] Checks for pinned rust nightly and pinned Aya commit
- [ ] Add golang's race detector (see #445)

PR #497, opened 2022-06-21
build: Add race detector flag

javierhonduco:
Let's add [Golang's race detector](https://go.dev/doc/articles/race_detector) to our Makefile. I have been running it in my test environment, and it showed zero errors, even under a lot of pressure (lots of concurrent debug info uploads, etc).

With the `GORACE="log_path=/tmp/race"` env variable we can pick a path where to write its output, rather than stdout / stderr.

Right now we should run this locally, but once https://github.com/parca-dev/parca-agent/issues/463 lands, this will be run in CI, too! 🎉 

Fixes: https://github.com/parca-dev/parca-agent/issues/445
## Test plan

**Build without sanitisers works**:

```
[javierhonduco@computer parca-agent]$ make build-dyn 
GOOS=linux GOARCH=amd64 CC=clang CGO_CFLAGS="-I /home/javierhonduco/code/parca-agent/dist/libbpf/usr/include" CGO_LDFLAGS="/home/javierhonduco/code/parca-agent/dist/libbpf/libbpf.a" go build  -tags osusergo,netgo -trimpath -v -o dist/parca-agent ./cmd/parca-agent
```

**Build with ASAN works**:

```
[javierhonduco@computer parca-agent]$ make build-dyn ENABLE_ASAN=yes
GOOS=linux GOARCH=amd64 CC=clang CGO_CFLAGS="-I /home/javierhonduco/code/parca-agent/dist/libbpf/usr/include" CGO_LDFLAGS="/home/javierhonduco/code/parca-agent/dist/libbpf/libbpf.a" go build -asan -tags osusergo,netgo -trimpath -v -o dist/parca-agent ./cmd/parca-agent
```

**Build with race detector works**:

```
[javierhonduco@computer parca-agent]$ make build-dyn ENABLE_RACE=yes
GOOS=linux GOARCH=amd64 CC=clang CGO_CFLAGS="-I /home/javierhonduco/code/parca-agent/dist/libbpf/usr/include" CGO_LDFLAGS="/home/javierhonduco/code/parca-agent/dist/libbpf/libbpf.a" go build -race -tags osusergo,netgo -trimpath -v -o dist/parca-agent ./cmd/parca-agent
```

_note_ ASAN + race isn't supported yet. I don't think we should check for this in the Makefile as it will increase its complexity and this is something that might change in the future:

```
[javierhonduco@computer parca-agent]$ make build-dyn ENABLE_ASAN=yes ENABLE_RACE=yes
GOOS=linux GOARCH=amd64 CC=clang CGO_CFLAGS="-I /home/javierhonduco/code/parca-agent/dist/libbpf/usr/include" CGO_LDFLAGS="/home/javierhonduco/code/parca-agent/dist/libbpf/libbpf.a" go build -asan -race -tags osusergo,netgo -trimpath -v -o dist/parca-agent ./cmd/parca-agent
go: may not use -race and -asan simultaneously
```

(`make test ENABLE_RACE=yes` and `make test ENABLE_ASAN=yes` also worked without issues)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -133,9 +133,6 @@ jobs:
         run: make build
 
       - name: Test
-        run: make test
-
-      - name: Test with race detector
         run: make test ENABLE_RACE=yes
 
       - name: Format
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -135,6 +135,9 @@ jobs:
       - name: Test
         run: make test
 
+      - name: Test with race detector
+        run: make test ENABLE_RACE=yes
+
       - name: Format
         run: |
           make format
--- a/Makefile
+++ b/Makefile
@@ -141,12 +141,12 @@ $(OUT_BPF): $(DOCKER_BUILDER) | $(OUT_DIR)
 endif
 
 test/profiler: $(GO_SRC) $(LIBBPF_HEADERS) $(LIBBPF_OBJ) bpf
-	sudo $(go_env) go test -v $(shell go list ./... | grep "pkg/profiler") $(SANITIZERS)
+	sudo $(go_env) go test $(SANITIZERS) -v $(shell go list ./... | grep "pkg/profiler")
 
 .PHONY: test
 ifndef DOCKER
 test: $(GO_SRC) $(LIBBPF_HEADERS) $(LIBBPF_OBJ) bpf test/profiler
-	$(go_env) go test -v $(shell go list ./... | grep -v "internal/pprof" | grep -v "pkg/profiler" | grep -v "e2e")
+	$(go_env) go test $(SANITIZERS) -v $(shell go list ./... | grep -v "internal/pprof" | grep -v "pkg/profiler" | grep -v "e2e")
 else
 test: $(DOCKER_BUILDER)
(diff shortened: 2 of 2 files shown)


kakkoyun (commented):
LGTM.

What does prevent us to utilize this in CI right now?

kakkoyun:
> Nothing that I can think of, we would have to add another job with the right sanitiser enabled and some extra options (such as crashing in case of a race condition instead of continuing, perhaps).
> 
> Adding them in these tests would be a good first step, but enabling them in the integration tests will pay way higher dividends!

Let's do it as part of this PR! @Sylfrena is already working on the e2e CI action. She can make sure necessary make is called properly.

https://github.com/parca-dev/parca-agent/blob/ba3b7410f79a7820c58b9b66c200a5bbe82474b5/Makefile#L259-L262

javierhonduco:
Mm, these are passing locally for me. The only difference I can see is that my patch version is earlier than the one in CI, will try with the same version we run in CI

.github/workflows/build.yml
```diff
@@ -135,6 +135,9 @@ jobs:
       - name: Test
         run: make test
```
kakkoyun, line 136: We can get rid of this one.

PR #499, opened 2022-06-21
chore(deps): use full tag for gh actions

maxbrunet:
Follow up on #491, that pinned a few actions I had missed in #485.

Without the tags used by releases, we would get obscure digest updates, instead of pretty release notes


kakkoyun (approved):
LGTM

PR #500, opened 2022-06-21
chore(renovate): fix gh actions _VERSION regex

maxbrunet:
It does not currently match until the end of line, match only `1` for Go `1.18.3` for example


kakkoyun (approved):
LGTM :100: 

You're really on top of things :+1:

Issue #460, opened 2022-06-14
chore: Reduce the time spent in CI checks


kakkoyun:
Similar to https://github.com/parca-dev/parca/issues/1152, we could reduce the time we spent in CI.

We should cross-build the binaries and copy them to the containers. We have already reduced the external dependencies to zero and statically linked the binaries.

The only challenge I foresee is to be able to cross-compile libbpf with CGO.

Keep in mind that we can't break the reproducible build guarantee.

xref: https://github.com/parca-dev/parca/pull/1153

maxbrunet:
Talked about in https://github.com/parca-dev/parca-agent/pull/465#discussion_r898725845, we can run multi-stage container build with different architecture with Buildkit:
https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

Here is a short example (Docker Buildx is used there): https://github.com/maxbrunet/prometheus-elasticache-sd/blob/v1.0.41/Dockerfile

Podman seems to support the `TARGET*` args too, but not `BUILD*`:
https://github.com/containers/buildah/blob/v1.26.1/imagebuildah/build.go#L349-L381

Not sure if that would help yet, but Podman can be used with Buildkit: https://pythonspeed.com/articles/podman-buildkit/

maxbrunet:
Made a test, all ARGs are available with Podman:

<details>
<summary><code>Dockerfile</code></summary>

```Dockerfile
FROM --platform="${BUILDPLATFORM}" docker.io/busybox

ARG TARGETPLATFORM \
    TARGETOS \
    TARGETARCH \
    TARGETVARIANT \
    BUILDPLATFORM \
    BUILDOS \
    BUILDARCH \
    BUILDVARIANT

RUN printf 'TARGETPLATFORM: %s\n' "${TARGETPLATFORM}" && \
    printf 'TARGETOS: %s\n' "${TARGETOS}" && \
    printf 'TARGETARCH: %s\n' "${TARGETARCH}" && \
    printf 'TARGETVARIANT: %s\n' "${TARGETVARIANT}" && \
    printf 'BUILDPLATFORM: %s\n' "${BUILDPLATFORM}" && \
    printf 'BUILDOS: %s\n' "${BUILDOS}" && \
    printf 'BUILDARCH: %s\n' "${BUILDARCH}" && \
    printf 'BUILDVARIANT: %s\n' "${BUILDVARIANT}"
```

</details>

```console
[...]

PR #510, opened 2022-06-27
chore: Fix syntax error with workflow

kakkoyun:


maxbrunet:
The Build workflow would also need to same paths as the container https://github.com/parca-dev/parca-agent/pull/476/files#r907554072

I do not see the syntax error :confused:, what is it exactly?

kakkoyun:
> The Build workflow would also need to same paths as the container https://github.com/parca-dev/parca-agent/pull/476/files#r907554072
> 
> I do not see the syntax error confused, what is it exactly?

`*/.cargo` is not an allowed syntax. It's not even valid YAML according to my linter. 

But now the failure is something else. I'll check.

kakkoyun:
The container step failed to build with an OOM. We should build the binaries outside the container. https://github.com/parca-dev/parca-agent/issues/460

PR #532, opened 2022-06-30
[WIP]  *: System wide v2

### LINKED ISSUE #260: Allow profiling of arbitrary processes
Up to now you can either use "kubernetes" profiling or "systemd-unit".

For local testing I would like the agent to be more flexible.

I would like agent to attach to a pid (or a list of pids) or all processes with a particular name.

Does this make sense to you?

javierhonduco:
This PR is a work in progress to attach one global perf event, instead of creating one per cgroup. This has several advantages, including:
- Enabling host-wide profiling: Allowing the profiling of the whole machine, without excluding a single process! As a side-effect, we will be able to profile processes that aren't containerised, too (*)
- Reducing resource usage. Both in userland, `NewCgroupProfiler`s instances linearly use memory and CPU, as well as kernel space (fewer data structures that the kernel needs to create, fewer events that have to be delivered, etc)
- Making the architecture more flexible. This design will allow us to incorporate new profilers more easily

## Rough plan

- [x] Attach to a single `perf_event_open`
- [x] Generate profiles
- [ ] Add the right labels for Kubernetes and Docker workloads
- [ ] Ensure that `parca-agent`'s web ui works

(*): In this PR, the goal is to keep the same behaviour we have right now but with a different architecture, that is, we will only show profiles for the cgroups or containers running under kubernetes that are passed via the CLI (note that this is not implemented yet). In a future PR, we will add another flag to either profile the whole machine (`--system-wide`) or even individual PIDs

## Test plan

Showing profiles for the cgroup where the test program runs:
image

Fixes #260


javierhonduco:
Just wanted to give a quick update on the current status. I am currently running Parca Agent and Parca on Minikube via `make dev/up`, which uses Tilt, in order to fetch the metadata we need from Kubernetes, such as the pod CgroupIDs and so on.

The first thing that looked a bit off is that all profiles were tagged with `cgroup_id=1`, which is not what I expected or saw while running Parca Agent in my bare-metal host.

To test whether this is an error on my side, I wanted to run a bpftrace script to fetch the cgroupid of the process running in the host

```dtrace
$ sudo bpftrace -e 'profile:hz:99 { @[comm, cgroup]=count(); }'
```

Installing and running it in Minikube proved to be a bit challenging, as it's a custom Linux build:
```
$ cat /etc/os-release
NAME=Buildroot
VERSION=2021.02.12-1-g14f2929-dirty
ID=buildroot
VERSION_ID=2021.02.12
PRETTY_NAME="Buildroot 2021.02.12"
```

Thankfully, https://github.com/kinvolk/cloud-native-bpf-workshop exists!

```
$ kubectl trace run --imagename "docker.io/albanc/kubectl-trace-bpftrace:e896345e3d8f80aa968422c6199ac5180d688f65" --init-imagename "docker.io/albanc/kubectl-trace-init:e896345e3d8f80aa968422c6199ac5180d688f65" parca-agent  -e "profile:hz:99 { @[comm, cgroup]=count(); }"
``` 
```
$ kubectl trace get kubectl-trace-e3b76215-53ea-476e-b820-f39f5cb50872
NAMESPACE       NODE            NAME                                                    STATUS  AGE
default         parca-agent     kubectl-trace-e3b76215-53ea-476e-b820-f39f5cb50872      Running 0s
```
```
$ kubectl trace attach kubectl-trace-e3b76215-53ea-476e-b820-f39f5cb50872
^C
first SIGINT received, now if your program had maps and did not free them it should print them out

@[kworker/1:4, 1]: 1
@[runc:[2:INIT], 1]: 1
@[runc:[0:PARENT], 1]: 1
@[docker-init, 1]: 1
@[bpftrace, 1]: 1
@[rcu_sched, 1]: 1
@[kworker/0:1, 1]: 1
@[kube-proxy, 1]: 2
@[runc:[1:CHILD], 1]: 2
@[dlv, 1]: 3
@[grpc-health-pro, 1]: 3
@[exe, 1]: 3
@[systemd, 1]: 3
@[5, 1]: 4
@[registry, 1]: 4
@[storage-provisi, 1]: 5
@[containerd-shim, 1]: 10
@[kube-scheduler, 1]: 11
@[coredns, 1]: 12
@[containerd, 1]: 12
@[runc, 1]: 22
@[parca-agent, 1]: 25
@[etcd, 1]: 42
@[kube-controller, 1]: 46
```

That confirms what I saw reported by Parca Agent. BPF's cgroup helper returns 1 for every single process. Will look into the Kernel source code to understand what can be the source of the issue.

A hypothesis is that Minikube seems to be using Cgroups v1, while my bare-metal host uses Cgroups v2, and this might be the cause for these results.

On Minikube:
```
$ grep cgroup /proc/filesystems
nodev   cgroup
nodev   cgroup2
```

We have both cgroup systems, but it seems like we might be using Cgroups v1:

``` 
$ cat /sys/fs/cgroup/cgroup.controllers
cat: /sys/fs/cgroup/cgroup.controllers: No such file or directory
$ grep ^cgroup /etc/mtab
cgroup /sys/fs/cgroup/systemd cgroup rw,nosuid,nodev,noexec,relatime,xattr,release_agent=/usr/lib/systemd/systemd-cgroups-agent,name=systemd 0 0
cgroup /sys/fs/cgroup/cpu,cpuacct cgroup rw,nosuid,nodev,noexec,relatime,cpu,cpuacct 0 0
cgroup /sys/fs/cgroup/cpuset cgroup rw,nosuid,nodev,noexec,relatime,cpuset 0 0
cgroup /sys/fs/cgroup/blkio cgroup rw,nosuid,nodev,noexec,relatime,blkio 0 0
cgroup /sys/fs/cgroup/perf_event cgroup rw,nosuid,nodev,noexec,relatime,perf_event 0 0
cgroup /sys/fs/cgroup/net_cls,net_prio cgroup rw,nosuid,nodev,noexec,relatime,net_cls,net_prio 0 0
cgroup /sys/fs/cgroup/devices cgroup rw,nosuid,nodev,noexec,relatime,devices 0 0
cgroup /sys/fs/cgroup/memory cgroup rw,nosuid,nodev,noexec,relatime,memory 0 0
cgroup /sys/fs/cgroup/freezer cgroup rw,nosuid,nodev,noexec,relatime,freezer 0 0
cgroup /sys/fs/cgroup/hugetlb cgroup rw,nosuid,nodev,noexec,relatime,hugetlb 0 0
cgroup /sys/fs/cgroup/pids cgroup rw,nosuid,nodev,noexec,relatime,pids 0 0
```

Will keep this PR updated once I learn more details

javierhonduco:
Superseded by https://github.com/parca-dev/parca-agent/pull/627

PR #534, opened 2022-06-30
buildid: Fix extraction issues

### LINKED ISSUE #475: Agent fails to read Rust symbols
I was looking into using Parca to profile a Rust binary but it fails to read the symbols from the binary.
I have verified that the binary does include the symbols.

The error I receive in the agent is:
```
level=warn ts=2022-06-18T00:55:37.856812691Z caller=maps.go:106 msg="failed to read object build ID" object=/proc/21830/root/app/src/rust/graphql/src/api/bin.runfiles/monorepo/src/rust/graphql/src/api/bin err="get elf build id: note name too long (5590599 bytes)"
```

Compilation OS: Fedora 35
Agent deployed on k8s cluster running Debian hosts
Rust version: 1.61.0

kakkoyun:

cc @v-thakkar @javierhonduco 

Fixes #475


pkg/buildid/buildid.go
```diff
@@ -56,29 +56,84 @@ func BuildID(path string) (string, error) {
...
+	}
+	defer f.Close()
+
+	s := f.Section(section)
 	if s == nil {
-		return "", fmt.Errorf("failed to find .note.go.buildid section")
+		return nil, fmt.Errorf("failed to find %s section", section)
 	}
+
+	notes, err := elfreader.ParseNotes(s.Open(), int(s.Addralign), f.ByteOrder)
```
javierhonduco, line 120: Do we have to close it?

pkg/elfreader/note.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// ParseNotes returns the notes from a SHT_NOTE section or PT_NOTE segment.
+func ParseNotes(reader io.Reader, alignment int, order binary.ByteOrder) ([]ElfNote, error) {
+	r := bufio.NewReader(reader)
+
+	// padding returns the number of bytes required to pad the given size to an
+	// alignment boundary.
+	padding := func(size int) int {
+		return ((size + (alignment - 1)) &^ (alignment - 1)) - size
```
javierhonduco, line 44: `&^` lol


actually... what does this do? 😄

pkg/buildid/buildid.go
kakkoyun: Noop, we don't. Section is not a Closer.

pkg/elfreader/note.go
javierhonduco: It seems that it's a logical NOT (+ AND)

> &^   bit clear (AND NOT)    integers

in the "Arithmetic operations" section from https://go.dev/ref/spec#Integer_operators

pkg/buildid/buildid.go
javierhonduco: Oops misread and thought it was returning a ReadWriteCloser rather than a ReadSeeker, my bad!

PR #550, opened 2022-07-06
chore: check Cargo.lock into git

maxbrunet:
> If you’re building an end product, which are executable like command-line tool or an application [...], check `Cargo.lock` into `git`. If you're curious about why that is, see ["Why do binaries have `Cargo.lock` in version control, but not libraries?" in the FAQ](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries).

https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html


kakkoyun (approved):
LGTM.
Good catch!

Issue #520, opened 2022-06-28
bpf: Make sure there is no concurrent access issue with CPU profiler


kakkoyun:
By migrating to Rust, we might introduce a regression for atomic counters.
The easier way to fix the issue is to use pre-CPU BPF maps.

xref: https://github.com/aquasecurity/libbpfgo/pull/156/files

maxbrunet:
Would it be as simple as `sed -i 's/HashMap/PerCpuHashMap/g' bpf/cpu-profiler/src/main.rs`?:

```diff
--- a/bpf/cpu-profiler/src/main.rs
+++ b/bpf/cpu-profiler/src/main.rs
@@ -8,7 +8,7 @@ pub static LICENSE: [u8; 4] = *b"GPL\0";
 use aya_bpf::{
     bindings::BPF_F_USER_STACK,
     macros::{map, perf_event},
-    maps::{HashMap, StackTrace},
+    maps::{PerCpuHashMap, StackTrace},
     programs::PerfEventContext,
     BpfContext,
 };
@@ -24,8 +24,8 @@ pub struct StackCountKey {
 }
 
 #[map(name = "counts")]
-pub static mut COUNTS: HashMap<StackCountKey, u64> =
-    HashMap::with_max_entries(MAX_STACK_ADDRESSES, 0);
+pub static mut COUNTS: PerCpuHashMap<StackCountKey, u64> =
+    PerCpuHashMap::with_max_entries(MAX_STACK_ADDRESSES, 0);
 
 #[map(name = "stack_traces")]
 pub static mut STACK_TRACES: StackTrace = StackTrace::with_max_entries(MAX_STACK_DEPTH, 0);
```

The maps are still named `counts`, so I guess the agent would pick them up without change:

https://github.com/parca-dev/parca-agent/blob/1e7edd6629ffd4b2ddf5fdb5ed10bf42c5279121/pkg/profiler/profiler.go#L378

That builds and at first look seem to work

kakkoyun:
@maxbrunet This seems about right. As you suggested we need to test this extensively.

And I'd suggest we have the `libbpf-go` and `libbpf` upgrade in a separate PR. We can also automate upgrades. Previously it was disabled because of the lack of tests. (We now have them https://github.com/parca-dev/parca-agent/blob/main/pkg/profiler/profiler_test.go) We should test our access patterns.

I think this is on the right track. Looking forward to the PR :hugs:

kakkoyun:
[...]

PR #552, opened 2022-07-06
Add atomics

javierhonduco:
We used to increment the stack counts in a data-race-free way by using atomics

https://github.com/parca-dev/parca-agent/blob/f2e1871d845f8fed2a36a667e6fd030cdbbb9cf0/parca-agent.bpf.c#L136-L137

but with the Rust rewrite (https://github.com/parca-dev/parca-agent/pull/377) we stopped using them and started relying on non-atomic instructions which can result in profiles not being accurate

## Test plan

**Generated code**
```
$ sudo bpftool prog dump jited id 3347
```

```
[...]
# sets .bx to 1 
  80:   mov    $0x1,%ebx  
  85:   mov    %rbx,-0x8(%rbp)
  89:   mov    %rbp,%rsi
  8c:   add    $0xffffffffffffffe8,%rsi
  90:   mov    %rbp,%rdx
  93:   add    $0xfffffffffffffff8,%rdx
  97:   movabs $0xffff8a5b0cba1000,%rdi
  a1:   call   0xffffffffcb286a08
  a6:   test   %rax,%rax
  a9:   je     0x00000000000000af
  ab:   add    $0x40,%rax
  af:   test   %rax,%rax
  b2:   je     0x00000000000000b9
# atomically access rax's pointee, and increment it by .bx which is 1
  b4:   lock add %rbx,0x0(%rax) 
  b9:   xor    %eax,%eax
  bb:   pop    %r13
  bd:   pop    %rbx
  be:   leave  
  bf:   ret    
```

Comparing the instructions of the code from my branch replacing the intrinsic with a normal increment:

image

**End-to-end test**

image

cc/ @maxbrunet @kakkoyun @Sylfrena 
Fixes: https://github.com/parca-dev/parca-agent/issues/520


bpf/cpu-profiler/src/main.rs
```diff
@@ -32,16 +32,17 @@ pub static mut STACK_TRACES: StackTrace = StackTrace::with_max_entries(MAX_STACK
...
 #[perf_event]
 pub fn profile_cpu(ctx: PerfEventContext) -> u32 {
-    match unsafe { try_profile_cpu(ctx) } {
-        Ok(ret) => ret,
-        Err(ret) => ret,
+    unsafe {
+        try_profile_cpu(ctx);
     }
+
+    0
```
javierhonduco, line 39: Returning 0 all the time as the other values don't have a lot of meaning here and we can't gather them from userspace anyways.

I believe that returning non-zero might result in the kernel dropping other BPF programs waiting to execute (think of tracepoints, for example)

bpf/cpu-profiler/src/main.rs
```diff
@@ -32,16 +32,17 @@ pub static mut STACK_TRACES: StackTrace = StackTrace::with_max_entries(MAX_STACK
...
 }
 
 #[inline(always)]
-unsafe fn try_profile_cpu(ctx: PerfEventContext) -> Result<u32, u32> {
+unsafe fn try_profile_cpu(ctx: PerfEventContext) {
     if ctx.pid() == 0 {
```
javierhonduco, line 44: Unrelated to this PR, but I noticed it now, we should not have this check as pid=0 is a special kernel process (swapper) and we might want to take it into account

```
$ sudo bpftrace -e 'profile:hz:99 /pid==0/ {@[kstack]=count()}'
```

kakkoyun (approved):
LGTM :tada: 

Awesome work!

PR #560, opened 2022-07-08
ci: fix image manifest push

### LINKED ISSUE #559: chore: Fix container push errors
https://github.com/parca-dev/parca-agent/runs/7252096350?check_suite_focus=true

maxbrunet:
I have not tested pushing, but ~I have a hunch~ Podman's `--tag` behave like the classic `docker build`, rather than changing to Docker Buildx's behavior with `--platform`, so platforms override one after the other and `--manifest` must be used instead

> When more than one platform is specified, the `--manifest` option should be used instead of the `--tag` option.

https://docs.podman.io/en/latest/markdown/podman-build.1.html#platform-os-arch-variant

Fixes #559

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -233,7 +233,7 @@ container: $(OUT_DIR)
 	podman build \
 		--platform linux/amd64,linux/arm64 \
 		--timestamp 0 \
-		--tag $(OUT_DOCKER):$(VERSION) .
+		--manifest $(OUT_DOCKER):$(VERSION) .
 
 .PHONY: container-dev
 container-dev:


Makefile
```diff
@@ -233,7 +233,7 @@ container: $(OUT_DIR)
 	podman build \
 		--platform linux/amd64,linux/arm64 \
 		--timestamp 0 \
-		--tag $(OUT_DOCKER):$(VERSION) .
+		--manifest $(OUT_DOCKER):$(VERSION) .
```
kakkoyun, line 236: Nice catch! I've definitely overlooked this one

kakkoyun (approved):
LGTM! Super fast reaction!

PR #586, opened 2022-07-13
Add API for the kernel config checks

v-thakkar:
Not yet ready for the review as I still need to push a few more stuff:

- [x] Kernel StackUnwinding check
- [x] Tests
- [x] Use of isInContainer API


v-thakkar:
Oops, lol. Messed up my working branch. But will push the clean one with other TODOs.

pkg/profiler/profiler.go
```diff
@@ -303,6 +304,34 @@ func (p *CgroupProfiler) Run(ctx context.Context) error {
...
+		level.Warn(p.logger).Log("msg", "Check container error:", err)
+	}
+
+	if isContainer {
+		level.Info(p.logger).Log("msg", "Parca agent is running in a container. It'll need to access the host kernel config")
+	}
+
+	bpfEnabled, err := kconfig.IsBPFEnabled()
+	if err != nil {
+		level.Warn(p.logger).Log("msg", "failed to determine if eBPF is supported", "err", err)
```
v-thakkar, line 318: Let's start with warnings as I'm sure this doesn't cover all the possible cases where kconfig file would be found. I'll send a follow-up PR for it in the upcoming weeks. We can error out after that.

pkg/kconfig/kconfig.go
```diff
@@ -0,0 +1,190 @@
...
+	"CONFIG_HAVE_EBPF_JIT",
+	"CONFIG_BPF_JIT",
+	"CONFIG_BPF_JIT_ALWAYS_ON",
+	"CONFIG_BPF_EVENTS",
+}
+
+func IsBPFEnabled() (bool, error) {
+	var e error
+
+	KernelConfig, e := getConfig()
```
kakkoyun, line 51:
```suggestion
	kernelConfig, e := getConfig()
```
nit

pkg/kconfig/kconfig.go
```diff
@@ -0,0 +1,190 @@
...
+
+// TODO(vthakkar): Add more locations for scanning the config
+// https://github.com/libbpf/libbpf/blob/9a3a42608dbe3731256a5682a125ac1e23bced8f/src/btf.c#L3114-L3122
+var (
+	configPaths = []string{
+		"/proc/config.gz",
+		"/boot/config",
+		"/boot/config-%s",
+	}
+)
```
kakkoyun, line 37:
```suggestion
var configPaths = []string{
		"/proc/config.gz",
		"/boot/config",
		"/boot/config-%s",
	}
```

nit. Or could this var scope could include the below var

pkg/kconfig/kconfig.go
```diff
@@ -0,0 +1,190 @@
...
+	"CONFIG_BPF",
+	"CONFIG_BPF_SYSCALL",
+	"CONFIG_HAVE_EBPF_JIT",
+	"CONFIG_BPF_JIT",
+	"CONFIG_BPF_JIT_ALWAYS_ON",
+	"CONFIG_BPF_EVENTS",
+}
+
+func IsBPFEnabled() (bool, error) {
+	var e error
```
kakkoyun, line 49:
```suggestion
	var err error
```
More idiomatic.

pkg/kconfig/kconfig.go
```diff
@@ -0,0 +1,190 @@
...
+
+	KernelConfig, e := getConfig()
+	if e != nil {
+		return false, e
+	}
+
+	for _, option := range ebpfCheckOptions {
+		value, found := KernelConfig[option]
+		if !found {
+			return false, fmt.Errorf("kernel Config required for ebpf not found, Config Option:%s", option)
```
kakkoyun, line 59:
```suggestion
			return false, fmt.Errorf("kernel config required for ebpf not found, Config Option:%s", option)
```

pkg/kconfig/kconfig.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+
+		if value != "y" {
+			return false, fmt.Errorf("kernel Config required for ebpf is disabled, Config Option:%s", option)
```
kakkoyun, line 63:
```suggestion
			return false, fmt.Errorf("kernel config required for ebpf is disabled, Config Option:%s", option)
```

pkg/kconfig/kconfig.go
```diff
@@ -0,0 +1,190 @@
...
+	"bufio"
+	"compress/gzip"
+	"fmt"
+	"os"
+	"strings"
+)
+
+const (
+	ContainerCgroupPath = "/proc/1/cgroup"
+	StackUnwinder       = "CONFIG_UNWINDER_FRAME_POINTER"
```
kakkoyun, line 26: How about ORC?

pkg/kconfig/kconfig.go
```diff
@@ -0,0 +1,190 @@
...
+
+		if value != "y" {
+			return false, fmt.Errorf("kernel Config required for ebpf is disabled, Config Option:%s", option)
+		}
+	}
+	return true, nil
+}
+
+func IsStackUnwindingEnabled() (bool, error) {
+	var e error
```
kakkoyun, line 70: The same issues were pointed out above.

pkg/profiler/profiler.go
```diff
@@ -303,6 +304,34 @@ func (p *CgroupProfiler) Run(ctx context.Context) error {
 	ctx, p.cancel = context.WithCancel(ctx)
 	p.mtx.Unlock()
 
+	isContainer, err := kconfig.IsInContainer()
+	if err != nil {
+		level.Warn(p.logger).Log("msg", "Check container error:", err)
```
kakkoyun, line 309:
```suggestion
		level.Warn(p.logger).Log("msg", "check container error:", err)
```

pkg/kconfig/kconfig_test.go
```diff
@@ -0,0 +1,52 @@
...
+		return
+	}
+
+	configPaths = []string{
+		"/testdata/config",
+		"/testdata/config.gz",
+	}
+	KernelConfig, e := getConfig()
+	if e != nil {
+		t.Logf("GetConfig error:%s", e.Error())
```
kakkoyun, line 38: Rather than logging, we should assert and fail. Logging in tests is rarely used in idiomatic Go.

You can check table tests if you have multiple cases with different scenarios.

pkg/kconfig/kconfig.go
```diff
@@ -0,0 +1,190 @@
...
+	if err = parse(s, KernelConfig); err != nil {
+		return KernelConfig, err
+	}
+	return KernelConfig, nil
+}
+
+// Returns true is the process is running in a container
+// TODO: Add a container detection via Sched to cover more scenarios
+// https://man7.org/linux/man-pages/man7/sched.7.html
+func IsInContainer() (bool, error) {
```
kakkoyun, line 163: I don't know if it's possible, but we need to check if it's running in privileged mode. Otherwise, the info about running in a container is not actionable.

kakkoyun (commented):
Great work! Thanks a lot!

I have pointed out a couple of style issues. I'm very surprised that they haven't actually caught up by the CI>

I think we should immediately stop the operation when the necessary checks fail. We can discover every possible scenario we need until we release it.

Also, I think we should check if the user has `CAP_SYS_ADMIN` at the beginning? Or do you want to handle it in a subsequent PR?

pkg/kconfig/kconfig.go
v-thakkar: Yeah, AFAIU it's not easy to check if it's running in a privileged mode. I'll check via sched though. Another idea was that we can try running some command(e.g. iproute2) that's only allowed in the privileged mode to detect that. But I'm not sure if that's worth it as everything is warning at the moment. Although I think it's still a useful Log for the user when the kernel config read fails because they can at least know that this could be a possible reason.

v-thakkar:
> Great work! Thanks a lot!
> 
> I have pointed out a couple of style issues. I'm very surprised that they haven't actually caught up by the CI>

Thanks. Yeah, they weren't caught by lints locally as well. :thinking: I will push the fixed-up commit in some time.

> I think we should immediately stop the operation when the necessary checks fail. We can discover every possible scenario we need until we release it.

Agree. That's my plan. To move to error out before the release. And until then I'll keep discovering the possible scenarios and adding checks for them.  

> Also, I think we should check if the user has `CAP_SYS_ADMIN` at the beginning? Or do you want to handle it in a subsequent PR?

Good point. I'll send a subsequent PR for that.

v-thakkar:
Superseded by #686

PR #587, opened 2022-07-13
elfwriter: Make sure linked section consistency

### LINKED ISSUE #582: elfwriter: Improve handling of links
Repro:

```shell
$ make && sudo dist/debug-info extract /lib64/libc.so.6 
$ readelf -a out/9b7d731d92528db48798ae00fd91905fc0904736.debuginfo | grep "corrupt" 
readelf: out/9b7d731d92528db48798ae00fd91905fc0904736.debuginfo: Warning: [ 4]: Link field (0) should index a string section.
[...]
     0: 0000000000000000     0 NOTYPE  LOCAL  DEFAULT  UND <corrupt>
[...]
   881: 000000000009f120    65 FUNC    GLOBAL DEFAULT bad section index[ 14] <corrupt>
[...]

### PR DESCRIPTION
**Pull request by kakkoyun**
Fixes #582 #506

## Changes 
- Refactor the API and introduce new wrappers

cc @javierhonduco

### CODE REVIEW

File: .github/workflows/build.yml [hunk 1]
```diff
@@ -101,7 +102,9 @@ jobs:
         run: make build
 
       - name: Test
-        run: make test ENABLE_RACE=yes
+        run: |
+          make test ENABLE_RACE=yes
+          # make test/profiler (Needs to be fixed, somehow fails because of generics)
```
javierhonduco, line 107: Let's create an issue for this so we don't forget to re-enable it

Makefile
```diff
@@ -113,7 +112,7 @@ $(OUT_BIN): $(DOCKER_BUILDER) | $(OUT_DIR)
 endif
 
 .PHONY: build-dyn
-build-dyn: libbpf $(OUT_BPF)
+build-dyn: $(OUT_BPF) libbpf
```
javierhonduco, line 115: No idea why this works but thanks for fixing it!

pkg/debuginfo/extract.go
```diff
@@ -69,71 +69,39 @@ func (e *Extractor) Extract(ctx context.Context, dst io.WriteSeeker, src string)
 	default:
 	}
 
-	elfFile, err := open(src)
+	f, err := os.Open(src)
```
javierhonduco, line 72: Thanks for cleaning my subpar approach!! :D

pkg/debuginfo/extract.go
```diff
@@ -142,8 +110,9 @@ var isDwarf = func(s *elf.Section) bool {
 
 var isSymbolTable = func(s *elf.Section) bool {
 	return s.Name == ".symtab" ||
-		s.Name == ".dynsymtab" ||
+		s.Name == ".dynsym" ||
 		s.Name == ".strtab" ||
+		s.Name == ".dynstr" ||
```
javierhonduco, line 115: Nice that these two symbol tables are added! I assume `dynsymtab` is not needed because they will be dynamic libraries and will be relocated? Not sure!

javierhonduco (approved):
LGTM!! Great work!! 

Have done a first review pass, will run some manual tests as we have discussed, and can do another review soon if you would like me to :)

pkg/debuginfo/extract.go
kakkoyun: We also to section type checks so if they're properly typed everything should work.

Also `.dynsymtab` is not a conventional section. I had made a mistake there previously.

PR #627, opened 2022-07-28
Migrate from cgroup profiling to system-wide profiling

### LINKED ISSUE #260: Allow profiling of arbitrary processes

javierhonduco:
This PR reworks how we attach the CPU profiler to create one global perf event, instead of creating one per cgroup. This has several advantages, including:

- Enabling host-wide profiling: Allowing the profiling of the whole machine, without excluding a single process! As a side-effect, we will be able to profile processes that aren't containerised, too (*)
- Reduced resource usage. Both in userland, `NewCgroupProfiler`s instances linearly use memory and CPU, as well as kernel space (fewer data structures that the kernel needs to create, fewer events that have to be delivered, etc)
- Making the architecture more flexible. This design will allow us to incorporate new profilers more easily

## Design

The discovery mechanism (Kubernetes or Systemd) no longer drives the creation cgroup profilers, instead:

1. A new CPU profiler is created when the agent starts, which is not attached to any cgroup;
2. When the kubernetes discovery mechanism detects a new container, it adds some metadata, a mapping of PID to labels, that can be collected on-demand;
3. Once a profiler 'round' finishes, we create the profiles. Instead of creating one per cgroup, we now generate one per PID. We check the mapping of PID to labels and add them to the profile

### Cgroup filtering
One of the initial goals of this work was to keep compatibility with the current filtering that Parca Agent has. For example, adding a new flag to enable system-wide collection (`--system-wide`, for example) and otherwise we would collect profiles for the Kubernetes clusters or Cgroups that are discovered or specified via arguments.

However, this proved to be [trickier than I anticipated](https://github.com/parca-dev/parca-agent/pull/532#issuecomment-1191326226), due to the fact that the `bpf_get_current_cgroup_id()` helper used to retrieve the Cgroup id in BPF does not work under cgroups v1, which is widely used in Kubernetes environments. While a migration to cgroups v2 is on the way, we definitely want to support both cgroup versions.

While this approach would have been the most efficient way to faithfully replicate the current behaviour, as we could do the filtering in BPF and avoid counting these stacks, as well as reading it in userspace, I tried fetching the Cgroup ID in userspace and do the filtering there. However, after trying out this idea, I was not able to read a cgroups v1 ID ([one of the approaches](https://github.com/torvalds/linux/blob/df0cc57e/tools/testing/selftests/bpf/cgroup_helpers.c#L274)).

In the end, I think that while the idea of trying to replicate the current behaviour (and in the future enable system-wide profile collection by default) has some validity, I am not sure if it's worth it. Not only this would be something temporary, but having to do the cgroup ID filtering in userspace will still generate more data and make us read the same data that we would have to do with system-wide enabled by default (minus the profile generation itself). Finding the correct paths for the cgroups can be difficult, too, which might bring bugs and will force us to do way more testing.

### Profile generation
Initially, we proposed to generate a profile per cgroup to maintain the current behaviour and to allow filtering by cgroup (see point above).

In this PR we generate one profile per PID.

## Other significant changes
- The web UI has undergone a lot of changes, we only show the profilers now and downloading the pprof profilers has been removed. We can re-add this in the future if needed, but most of the time Parca Agent is tested with Parca with receives the profiles and will do symbolication, making the profiles easier.
- Removal of the Systemd / Cgroup discovery mechanism, as it's no longer needed. We are losing the unit names, which we can re-add later on.

## Future work
- Remove unneeded code
- Add back the systemd/cgroup metadata, if needed
- Tackle the TODOs 

## Test plan
**Bare metal**

image

**Minikube**

image

**Web UI**
image

Fixes #260


javierhonduco:
Fix web tests

cmd/parca-agent/main.go
```diff
@@ -75,9 +72,10 @@ type flags struct {
 	SamplingRatio      float64           `kong:"help='Sampling ratio to control how many of the discovered targets to profile. Defaults to 1.0, which is all.',default='1.0'"`
 	Kubernetes         bool              `kong:"help='Discover containers running on this node to profile automatically.',default='true'"`
 	PodLabelSelector   string            `kong:"help='Label selector to control which Kubernetes Pods to select.'"`
-	Cgroups            []string          `kong:"help='Cgroups to profile on this node.'"`
-	// SystemdUnits is deprecated and will be eventually removed, please use the Cgroups flag instead.
-	SystemdUnits []string `kong:"help='[deprecated, use --cgroups instead] systemd units to profile on this node.'"`
+	// Cgroups is deprecated and will be eventually removed.
+	Cgroups []string `kong:"help='[deprecated, will automatically profile all cgroups] Cgroups to profile on this node.'"`
```
brancz, line 76: we're pre v1 and this is already unused now, so I would recommend to just remove it instead

brancz (approved):
Woah. Amazing!! 💯

kakkoyun:
> ## Other significant changes
> * The web UI has undergone a lot of changes, we only show the profilers now and downloading the pprof profilers has been removed. We can re-add this in the future if needed, but most of the time Parca Agent is tested with Parca with receives the profiles and will do symbolication, making the profiles easier.
> * Removal of the Systemd / Cgroup discovery mechanism, as it's no longer needed. We are losing the unit names, which we can re-add later on.
> 
> ## Future work
> * Remove unneeded code
> * Add back the systemd/cgroup metadata, if needed
> * Tackle the TODOs

Could you please open issues for all these? 


Note to the releaser: These should be mentioned in the release notes.

cmd/parca-agent/main.go
javierhonduco: Happy to do that! Thought that it could be confusing for users that might not be aware of these changes to see that not only their chosen cgroup is profiled

cmd/parca-agent/main.go
```diff
@@ -390,16 +268,10 @@ func main() {
...
-		ctx, cancel := context.WithCancel(ctx)
-		g.Add(func() error {
-			level.Debug(logger).Log("msg", "starting target manager")
-			return tm.Run(ctx, m.SyncCh())
-		}, func(error) {
-			cancel()
-		})
-	}
+	// Add profilers.
+	_ = pp.AddProfiler(ctx, profiler.NewCPUProfiler, func() map[int]model.LabelSet {
```
kakkoyun, line 272: Let's not just ignore the errors.

pkg/profiler/pool.go
```diff
@@ -30,89 +30,84 @@ import (
...
+		pp.reg,
+		pp.ksymCache,
+		pp.objCache,
+		pp.writeClient,
+		pp.debugInfoClient,
+		labelSet,
+		pp.profilingDuration,
+		allGroups,
+	)
+	go func() {
```
kakkoyun, line 105: Maybe we should move this goroutine to the main and run it inside the rungroup, so we can stop everything else and clean up when a profiler fails. This is more align with our current code base.

pkg/profiler/noop.go
```diff
@@ -0,0 +1,81 @@
+// Copyright (c) 2022 The Parca Authors
```
kakkoyun, line 1: Love this ❤️

pkg/profiler/cpu.go
```diff
@@ -0,0 +1,909 @@
...
+			},
+		},
+		byteOrder: byteorder.GetHostByteOrder(),
+		metrics:   newMetrics(reg, target),
+		pidLabels: pidLabels,
+	}
+}
+
+func (p *CPUProfiler) Name() string {
+	return "bpf-cpu-cycles"
```
kakkoyun, line 266:
```suggestion
	return "bpf-cpu-sampled-stack-trace"
```
?

kakkoyun (commented):
Looking good. I did the first round, and I'll continue to review this later on.

Especially, I haven't checked `pkg/profiler/cpu.go` yet.

Awesome work so far 💯

pkg/profiler/cpu.go
javierhonduco: The reason for the name is that I think it's a good indicator of what kind of sampling we are doing. In the future, we might want to add other perf events, such as cache miss counters and others (e.g.`bpf-cpu-l3-miss`) and I think this naming scheme would be clearer

pkg/profiler/pool.go
javierhonduco: Good shout! Was planning to do this but forgot about it 😅 .

(I will address this and the rest of the comments after the review is complete 😄 )

cmd/parca-agent/main.go
```diff
@@ -390,17 +260,25 @@ func main() {
...
-			cancel()
+			return profiler.Run(ctx)
+		}, func(err error) {
+			profiler.Stop()
+			level.Error(logger).Log("msg", "profiler ended with", "error", err, "profilerName", profiler.Name())
 		})
 	}
 
+	if err != nil {
+		panic("fatal error: could not start profiler")
```
kakkoyun, line 279:
```suggestion
		panic("fatal error: could not start profiler")
```
		level.Error(logger).Log("fatal error: could not start profiler", err)
 		os.Exit(1)

better than panicing

kakkoyun:
Just a couple of nits and then we can handle the rest in subsequent PRs

cmd/parca-agent/main.go
javierhonduco: Ohh this was a bad merge, good eye!!

javierhonduco:
Address feedback

kakkoyun (approved):

PR #643, opened 2022-08-01
profiler: Remove BPF stack helper error logging

javierhonduco:
As they are incredibly verbose and prevent the legibility of the logs overall. It should be pretty safe to do this as the helpers will fail when executed in contexts that don't have either stack and we can't really use this information to have a better understanding of what's failing.


kakkoyun (approved):

kakkoyun:
Just to make sure, do we have a metric to record these? @javierhonduco 
If not, we should add a metric or include these cases in a failure metric.

javierhonduco:
I was originally thinking on adding such metrics, but then realised that there might not be much value in doing so. That being said, I am not against adding metrics, happy to do so if you think it could be useful

kakkoyun:
> I was originally thinking on adding such metrics, but then realised that there might not be much value in doing so. That being said, I am not against adding metrics, happy to do so if you think it could be useful

Metrics are cheap and these might help us to surface some issue so it's better than being blind.

kakkoyun (approved):

brancz:
Maybe it's time to start a Parca Agent dashboard? 😎

PR #648, opened 2022-08-02
Remove profile listener

javierhonduco:
The profile listener was used to intercept profile writes when downloading pprof files from the UI. This functionality has been removed in https://github.com/parca-dev/parca-agent/pull/627, and if we decide to add something like this back, we will probably save the pprof files directly to the FS, just before they are sent to the async client, avoiding the need for this wrapper.


brancz:
It makes sense to remove for now, but I would recommend to still have something HTTP based for debugging purposes. It tends to be much more difficult to have FS access than let's say do a port-forward in Kubernetes.

kakkoyun (changes requested):
I think, as Frederic suggested, we should keep the existing functionality. We're not in a hurry we still have time until we cut the next release. Let's have the UI back.

PR #649, opened 2022-08-02
docs: System-wide profiling related updates

javierhonduco:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,12 +61,10 @@ $ make -C bpf setup
 
 $ make
 
-$ sudo dist/parca-agent --node=test --cgroups=docker.service --log-level=debug --kubernetes=false --insecure
+$ sudo dist/parca-agent --node=test --log-level=debug --kubernetes=false --insecure
 ```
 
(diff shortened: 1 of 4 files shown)


docs/design.md
```diff
@@ -6,35 +6,27 @@ Parca Agent uses BPF CO-RE (Compile Once – Run Everywhere) using [libbpf](http
 
 From a high-level it performs the following steps:
 
-* [Target Discovery](#target-discovery): Discovers cgroups to attach the profiler on the machine the Parca Agent is running on.
+* [Target Discovery](#target-discovery): Discovers Kubernetes clusters, if any, and fetches its metadata.
```
kakkoyun, line 9: I think this should be called something else.
Our targets now are all the processes on the system.

We should add a step to describe how we provide context for the processes using container runtimes, cgroups and systems units.

javierhonduco: "Metadata Discovery"?

javierhonduco:
This PR updates all the docs, except `design.md` which I will update in another PR (cutting an issue for that one)

docs/design.md
kakkoyun: Both sound great. You pick one :)

PR #662, opened 2022-08-03
debuginfo: Increase exists cache size

javierhonduco:
and increase the cache expiration time

Spotted this while running the Agent for several minutes. The rationale for this change is that:
- The increase in memory usage won't be very large: from 2,048 bytes to 6,144 bytes (+ the overhead of the cache data structures)

(Also removed an outdated comment).

Let me know what you think and if I missed anything

### DIFF
--- a/pkg/debuginfo/debuginfo.go
+++ b/pkg/debuginfo/debuginfo.go
@@ -52,8 +52,8 @@ func New(logger log.Logger, client Client) *DebugInfo {
 		logger: logger,
 		client: client,
 		existsCache: cache.New(
-			cache.WithMaximumSize(128),                // Arbitrary cache size.
-			cache.WithExpireAfterWrite(2*time.Minute), // Arbitrary period.
+			cache.WithMaximumSize(512),                 // Arbitrary cache size.
+			cache.WithExpireAfterWrite(15*time.Minute), // Arbitrary period.
 		),
 		debugInfoSrcCache: cache.New(cache.WithMaximumSize(128)), // Arbitrary cache size.
 		// Up to this amount of debug files in flight at once. This number is very large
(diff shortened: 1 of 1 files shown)


brancz:
The rationale is correct, but I think we should still have a time-based eviction policy for now as most Parca servers are ephemeral and after rollouts, all data is gone. I think the time based eviction can be much larger though, 15 minutes would probably still be acceptable.

javierhonduco:
That makes sense! I will wait for @kakkoyun's comments and will update the PR 😄

pkg/debuginfo/debuginfo.go
```diff
@@ -52,8 +51,8 @@ func New(logger log.Logger, client Client) *DebugInfo {
 		logger: logger,
 		client: client,
 		existsCache: cache.New(
-			cache.WithMaximumSize(128),                // Arbitrary cache size.
-			cache.WithExpireAfterWrite(2*time.Minute), // Arbitrary period.
+			cache.WithMaximumSize(512), // Arbitrary cache size.
```
kakkoyun, line 54: :+1:

pkg/debuginfo/debuginfo.go
```diff
@@ -52,8 +51,8 @@ func New(logger log.Logger, client Client) *DebugInfo {
...
+			cache.WithPolicy("lru"),
```
kakkoyun, line 55: We need to keep a time-based eviction strategy. 
As Frederic mentioned, things can change on the Parca server side. Invalid or corrupted information can be deleted on the server-side. 
The single source of truth should be the server. However, we don't want to overwhelm the server with the "exists" queries that's why we added the cache and set an expiration date so that symbolization would eventually be consistent.

kakkoyun (commented):
I'm happy to increase the cache size, but we should keep the time-based expiration.

kakkoyun (approved):

Issue #661, opened 2022-08-03
Be more explicit in service discovery logging


brancz:
I think the log line:

```
level=info ts=2022-08-02T19:28:49.515254599Z caller=podinformer. go:148 discovery=ip-10-149-92-64.eu-central-1.compute.internal msg="starting pod controller"
```

Is useful as it tells us that Kubernetes metadata discovery is starting. Most people won't know or care that that's what "starting the pod controller" means though. I think we should rephrase this log line to be clearer.

Issue #660, opened 2022-08-03
Only log about debuginfo collection if it's turned off


brancz:
By default debuginfo collection is on and that's good, but it can be confusing to users that it is being logged about.

```
level=info ts=2022-08-02T19:28:49.497612717Z caller=main.go:163 msg="debug information collection is enabled"
```

I think we should only log about this when we turn it off as that's the situation we want to inform about.

Namanl2001:
This issue has already been resolved [here](https://github.com/parca-dev/parca-agent/blob/HEAD/cmd/parca-agent/main.go#L651-L657)

brancz:
Thanks for catching that!

Issue #356, opened 2022-04-20
profiler: Remove unique label values from the metrics


kakkoyun:
`parca_agent_profiler_failed_stack_unwinding_attempts_total{target="{__cgroup_path__=\"/sys/fs/cgroup/perf_event/kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-pod2ef4eceb_afe7_4f19_ab76_1458216087eb.slice/docker-ff7933c90e305fe74a5ebc5ef0e6c2e3dd71250d3e022c7af9714a215977e7ea.scope\", addonmanager_kubernetes_io_mode=\"Reconcile\", container=\"storage-provisioner\", containerid=\"docker://ff7933c90e305fe74a5ebc5ef0e6c2e3dd71250d3e022c7af9714a215977e7ea\", integration_test=\"storage-provisioner\", namespace=\"kube-system\", node=\"parca-agent\", pod=\"storage-provisioner\"}",type="kernel"} 6`

We are using `target.String()` as is it causes unique label values to leak profiler metrics, we need to remove them

cc @Sylfrena

palash25:
i am guessing this issue can be closed now since its mentioned in this PR description https://github.com/parca-dev/parca-agent/pull/663  that it fixes this issue? 

I also couldn't find any metric named  `parca_agent_profiler_failed_stack_unwinding_attempts_total` i am guessing it was replaced with `parca_agent_profiler_map_attempts_total` and there is no usage of `target.String()`.

kakkoyun:
This might be deprecated. I'm closing it.

@palash25 Thanks for the bring it to our notice.

PR #663, opened 2022-08-03
Refactor after and cleanup system-wide profiling changes

### LINKED ISSUE #624: ksym: Move prometheus cache stats to the ksym package
see: https://github.com/parca-dev/parca-agent/pull/391

kakkoyun:

After the #627 changes, we changed the control flow of the agent. Previously we were using service discovery to find cgroups, and we were attaching a profiler per cgroup. This has changed metadata resolution drastically. On top of that, we had a bunch of residue because of cgroup assumptions.

This PR addresses a bunch of issues and renames a lot of packages and types. To:
* Clean the tech debt
* Increase readability (rename all the packages)
* Increase testability (interfaces everywhere)
* To add future profilers easily

But mostly, providing the metadata information could be cleaner.

Also adds `systemd` metadata provider.

Fixes #624 #356 #660 #661


pkg/cri/cri.go
```diff
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package containerutils
+package cri
```
brancz, line 14: I don't think `cri` is accurate. The whole point of this package is that each runtime behaves subtly differently and not under the Kubernetes CRI spec.

Something that has less existing meaning would be better I think. Eg. `containerruntimes` or just `container`.

brancz (commented):
I'm not 100% certain the profiler abstraction is quite right, but I don't think it matters as long as we only have a single profiler type so we can still play with it for the time being. As in we don't know whether maybe in the future a ring-buffer-based one might make more sense.

In any case this lgtm.

pkg/cri/cri.go
kakkoyun: Will do! My first attempt was `containerruntime` :+1:

kakkoyun:
> I'm not 100% certain the profiler abstraction is quite right, but I don't think it matters as long as we only have a single profiler type so we can still play with it for the time being.

I'm %100 open to suggestions. Do you have any design in mind?

pkg/discovery/discovery_manager.go
```diff
@@ -97,7 +97,7 @@ func NewManager(logger log.Logger, reg prometheus.Registerer, options ...func(*M
 		metrics:        newMetrics(reg),
 		updatert:       5 * time.Second,
 		triggerSend:    make(chan struct{}, 1),
-		pidLabels:      make(map[int]model.LabelSet),
+		pidLabels:      make(map[uint32]model.LabelSet),
```
javierhonduco, line 100: Process IDs are an `int` (https://ftp.gnu.org/old-gnu/Manuals/glibc-2.2.3/html_node/libc_554.html). I was planning to change all the `uint32`s to become `int`s.

May I suggest that we undo this change and in this, or another PR, change all the PIDs to be 32 bit signed integers?

pkg/debuginfo/processor.go
```diff
@@ -33,7 +33,8 @@ import (
 
 var errNotFound = errors.New("not found")
 
-type DebugInfo struct {
+// Processor is a mechanism for extracting or find the relevant debug information for the discovered executables.
```
javierhonduco, line 36: Nit: typo,
```suggestion
// Processor is a mechanism for extracting or finding the relevant debug information for the discovered executables.
```

pkg/debuginfo/processor.go
```diff
@@ -33,7 +33,8 @@ import (
 
 var errNotFound = errors.New("not found")
 
-type DebugInfo struct {
```
javierhonduco, line 36: I think that `DebugInfo` or `DebugInfo$thing` is a more direct name?

pkg/k8s/k8s.go
```diff
@@ -155,7 +155,7 @@ type ContainerDefinition struct {
 	CgroupV1      string
 	CgroupV2      string
 	MountSources  []string
-	PID           int
+	PID           uint32
```
javierhonduco, line 158: Same as above, `int` should be a better type to represent PIDs

pkg/process/mapping.go
```diff
@@ -40,7 +44,7 @@ func (m *Mapping) PIDAddrMapping(pid uint32, addr uint64) (*profile.Mapping, err
 	maps, ok := m.pidMappings[pid]
 	if !ok {
 		var err error
-		maps, err = m.fileCache.MappingForPID(pid)
+		maps, err = m.cache.MappingForPID(pid)
```
javierhonduco, line 47: +1 to this name!

javierhonduco (commented):
Good stuff! Took an initial glance at the code and left some minor comments and nits so far

pkg/discovery/discovery_manager.go
kakkoyun: Of course, I thought they are `uint32` :see_no_evil: I'll do the change everywhere. Amazing catch!

pkg/debuginfo/processor.go
kakkoyun: The name of the shoudl be consider with the package name. The full name is `debuginfo.Processor`, it's more align with the conventions rather than `debuginfo.DebugInfo`. I'm open to other suggestions though.

javierhonduco: Some ideas:

`debuginfo.Extractor`
`debuginfo.Finder`
`debuginfo.Manager`

brancz:
> I'm %100 open to suggestions. Do you have any design in mind?

Like I said, since we aren't actually adding other profilers just yet I think this is not truly an abstraction yet, so I wouldn't do anything different for now, just something to keep in mind so we don't paint ourselves into a corner.

pkg/discovery/discovery_manager.go
kakkoyun: I'm ok to represent every PID in user-space a `int`. However the PID we get from eBPF helper functions are `uint32` https://github.com/iovisor/bcc/blob/master/docs/reference_guide.md#4-bpf_get_current_pid_tgid

pkg/debuginfo/processor.go
kakkoyun: Manager sounds good. Because it does all of the above.

README.md
```diff
@@ -61,35 +61,36 @@ Flags:
...
-      --store-address=STRING      gRPC address to send profiles and symbols to.
-      --bearer-token=STRING       Bearer token to authenticate with store.
-      --bearer-token-file=STRING
+      --metadata-container-runtime-socket-path=STRING
+                                  The filesystem path to the container runtimes
+                                  socket. Leave this empty to use the defaults.
+      --local-store-directory="./tmp/profiles"
+                                  The local directory to store the profiling
+                                  data.
+      --remote-store-address=STRING
```
javierhonduco, line 77: Such a better name!

pkg/address/normalizer.go
```diff
@@ -0,0 +1,74 @@
...
+
+	objFile, err := n.objCache.ObjectFileForProcess(pid, m)
+	if err != nil {
+		if !(errors.Is(err, objectfile.ErrNoFile) || errors.Is(err, os.ErrNotExist)) {
+			level.Debug(logger).Log("msg", "failed to open object file", "err", err)
+		}
+		return addr
+	}
+
+	// Transform the address by normalizing Kernel memory offsets.
```
javierhonduco, line 66: The understanding I had here was that we were normalising user space addresses to subtract the base process address

pkg/debuginfo/upload.go
```diff
@@ -55,7 +54,7 @@ func (u *Uploader) UploadAll(ctx context.Context, srcDbgInfo map[SourceInfo]io.R
 	var result *multierror.Error
 	for src, r := range srcDbgInfo {
 		if err := u.Upload(ctx, src, r); err != nil {
-			level.Warn(u.logger).Log(
+			level.Debug(u.logger).Log(
```
javierhonduco, line 57: Should we keep it as a warning? If uploading fails, something has gone pretty wrong, no?

pkg/discovery/discovery_manager.go
```diff
@@ -292,25 +300,21 @@ func (m *Manager) AllGroups() map[string][]*target.Group {
...
-				// TODO: handle deletions.
-				m.pidLabels[pid] = group.Labels
+				pidLabels[pid] = group.Labels
 			}
 		}
 	}
-}
+	// Just swap the map with the new one. Considering how frequently we profile,
+	// we shouldn't receive anything when the discoverer reconcile labels.
+	m.pidLabels = pidLabels
```
javierhonduco, line 315: Could this potentially overwrite previous values we care about?

I am thinking of this:
1. The provider starts, and provides the currently running k8s pods
2. Then a new pod starts, my understanding is that we would get an update just for it
3. We will wipe the data stored in `m.pidLabels` for the pids stored in 1)

pkg/discovery/kubernetes/kubernetes.go
```diff
@@ -177,24 +177,6 @@ func (c *ContainerDefinition) Labels() []*profilestorepb.Label {
 	}}
 }
 
-func (c *ContainerDefinition) PerfEventCgroupPath() string {
```
javierhonduco, line 180: Hehe, was going to delete this soon, glad you are removing it here! Makes me quite happy to see this gone 😄

pkg/discovery/systemd.go
```diff
@@ -0,0 +1,145 @@
...
+			return fmt.Errorf("failed to list units: %w", err)
+		}
+		var targetGroups []*Group
+		for _, unit := range units {
+			labelset := model.LabelSet{
+				"systemd_unit": model.LabelValue(unit),
+			}
+			pids, err := c.pids(ctx, unit)
+			if err != nil {
+				return fmt.Errorf("failed to get PIDs for unit %s: %w", unit, err)
```
javierhonduco, line 72: Perhaps let's `continue` so we continue dealing with the other units

pkg/discovery/systemd.go
```diff
@@ -0,0 +1,145 @@
...
+		return nil, err
+	}
+
+	pid, err := strconv.Atoi(strings.TrimSpace(string(output)))
+	if err != nil {
+		return nil, err
+	}
+
+	return []int{pid}, nil
+}
```
javierhonduco, line 145: Really like the idea of having this systemd discovery mechanism 💯

pkg/ksym/ksym.go
```diff
@@ -109,14 +103,13 @@ func (c *Cache) Resolve(addrs map[uint64]struct{}) (map[uint64]string, error) {
...
-		c.Stats.Total += 1
-
 		if !ok {
 			notCached = append(notCached, addr)
+			c.cacheFetch.WithLabelValues("miss").Inc()
 			continue
 		}
 		res[addr] = sym
-		c.Stats.Hits += 1
+		c.cacheFetch.WithLabelValues("hits").Inc()
```
javierhonduco, line 112: Thanks for cleaning this up! 🙇

pkg/objectfile/cache.go
```diff
@@ -19,30 +19,32 @@ import (
...
+			burrow.WithExpireAfterAccess(time.Minute), // Deprecate it if it is not used for 6 profiling cycles (default).
+		),
 	}
 }
 
 // ObjectFileForProcess returns the object file for the given mapping and process id.
 // If object file is already in the cache, it is returned.
 // Otherwise, the object file is loaded from the file system.
-func (c *cache) ObjectFileForProcess(pid uint32, m *profile.Mapping) (*MappedObjectFile, error) {
+func (c *cache) ObjectFileForProcess(pid int, m *profile.Mapping) (*MappedObjectFile, error) {
```
javierhonduco, line 47: Noice!

javierhonduco (approved):
Great work!! Thanks so much for cleaning up the accumulated cruft and the pid uint32 → int32 changes! 

Left some comments, but feel free to merge, this is solid, and the testing you've done is very thorough, so we can always iterate on other PRs

kakkoyun:
@jnsgruk Could you please a look at 51abc90a7a6b97d51fe31551b9641944c4f0217c?

jnsgruk:
> @jnsgruk Could you please a look at [51abc90](https://github.com/parca-dev/parca-agent/commit/51abc90a7a6b97d51fe31551b9641944c4f0217c)?

Looked good apart from one conditional, have left a comment which hopefully hasn't disappeared! Let me know if you can't see it :)

kakkoyun:
> > @jnsgruk Could you please a look at [51abc90](https://github.com/parca-dev/parca-agent/commit/51abc90a7a6b97d51fe31551b9641944c4f0217c)?
> 
> Looked good apart from one conditional, have left a comment which hopefully hasn't disappeared! Let me know if you can't see it :)

I think it got disappeared. I can't see it :/

pkg/address/normalizer.go
kakkoyun: That's what we do. The comment is misleading. I'll fix it.

snap/hooks/configure
```diff
@@ -29,9 +23,9 @@ if [[ -z "$http_address" ]]; then
 fi
 
 # gRPC address to send profiles and symbols to.
-store="$(snapctl get store-address)"
+remote_store="$(snapctl get remote-store-address)"
 if [[ -z "$store" ]]; then
```
jnsgruk, line 27:
```suggestion
if [[ -z "$remote_store" ]]; then
```

pkg/debuginfo/upload.go
kakkoyun: There's an error returned at the end that actually gets logged. We don't need additional warning if we aren't actually debugging.

snap/hooks/configure
jnsgruk: @kakkoyun ok this was a more sensible place to put the comment ;-)

pkg/discovery/discovery_manager.go
kakkoyun: I thought about this and decided not to spend too much time on it. This could happen for a single profile run where the process disappears, and we receive a profile after that. The downside of the previous approach we never delete anything from the map, and that meant we ever only aggregated, basically, it was a leak.

I'm happy to introduce some sort of time-based strategy to clean if the metadata has not been accessed for a time. What do you think?

snap/hooks/configure
kakkoyun: Great catch!

pkg/discovery/discovery_manager.go
javierhonduco: > I'm happy to introduce some sort of time-based strategy to clean if the metadata has not been accessed for a time. What do you think?

Sounds great, it's what I was thinking to do at some point to avoid the unbounded growth of the map as you mentioned 😄 . We can do this in another PR if you prefer

kakkoyun: I'll address it on this one :+1:

jnsgruk:
Hmm, failure on the tests for the snap. Looking at the logs, the failure is sort of legit, in that the agent exits because the actions runner doesn't properly support eBPF, which is probably true :thinking:

jnsgruk:
> Hmm, failure on the tests for the snap. Looking at the logs, the failure is sort of legit, in that the agent exits because the actions runner doesn't properly support eBPF, which is probably true thinking
> 
> [parca-dev/parca-agent/runs/7742823897?check_suite_focus=true#step:8:1](https://github.com/parca-dev/parca-agent/runs/7742823897?check_suite_focus=true#step:8:1)
> 
> I came across this [gist](https://gist.github.com/chadbrewbaker/64c561cf74a07243c762d43f4b67e436), although it's lacking enough in context that it's a little suspect

We'll need to update [this line](https://github.com/kakkoyun/parca-agent/blob/5bdec23d1612a64a09860f72fd7e029aa9e93a93/.github/workflows/snap.yml#L243) anyway to adjust the snap config name :)

pkg/discovery/discovery_manager.go
kakkoyun: @javierhonduco PTAL https://github.com/parca-dev/parca-agent/pull/663/commits/5bdec23d1612a64a09860f72fd7e029aa9e93a93

jnsgruk:
So I'm doing some testing over on my fork, I've got an `upterm` session so I can ssh into the runner, and there is definitely *some* eBPF support. I can, for example run `sudo execsnoop-bpfcc` and things look right.

jnsgruk:
> So I'm doing some testing over on my fork, I've got an `upterm` session so I can ssh into the runner, and there is definitely _some_ eBPF support. I can, for example run `sudo execsnoop-bpfcc` and things look right.
> 
> Interestingly, I get these log lines from the agent
> 
> ```
> level=debug name=parca-agent ts=2022-08-09T10:23:34.334773778Z caller=main.go:148 msg="config not found" path=/proc/config.gz
> level=debug name=parca-agent ts=2022-08-09T10:23:34.334830579Z caller=main.go:148 msg="config not found" path=/boot/config
> level=debug name=parca-agent ts=2022-08-09T10:23:34.334846279Z caller=main.go:148 msg="config not found" path=/boot/config-%s
> level=error name=parca-agent ts=2022-08-09T10:23:34.334856579Z caller=main.go:127 err="host kernel does not support eBPF"
> ```
> 
> Note the `/boot/config-%s` line? There is a kernel config, though:
> 
> ```
> runner@fv-az204-851:~/work/parca-agent/parca-agent$ ls -la /boot/config-$(uname -r)
> -rw-r--r-- 1 root root 176404 Jun 23 19:44 /boot/config-5.15.0-1014-azure
> ```

Looks like [here](https://github.com/parca-dev/parca-agent/pull/663/files#diff-57833d176ed681886b67a7fa2e4dc2f68d5e00cb1db269e7854fb3dafb45f54eR144) you're iterating over `configPaths`, which is defined statically [here](https://github.com/parca-dev/parca-agent/pull/663/files#diff-57833d176ed681886b67a7fa2e4dc2f68d5e00cb1db269e7854fb3dafb45f54eR104), so I'm guessing there is just some string formatting missing there? :) That might solve the snap test problem all together :)

v-thakkar:
@jnsgruk Yeah, it should have been able to find the config via '/boot/config-%s'.  The string is formatted here: https://github.com/parca-dev/parca-agent/blob/main/pkg/kconfig/kconfig.go#L69 . I'll check the PR locally to see what's the reason.

jnsgruk:
> @jnsgruk Yeah, it should have been able to find the config via '/boot/config-%s'.  The string is formatted here: https://github.com/parca-dev/parca-agent/blob/main/pkg/kconfig/kconfig.go#L69 . I'll check the PR locally to see what's the reason.

Yeh, I dug through the list of kernel config options its checking for and they're all configured correctly afaict, so more likely just and error reading the file

jnsgruk:
Ah! @kakkoyun @v-thakkar I know what it is! The snap confinement is probably blocking access to '/boot'. 

I can follow up with another PR after this one has landed to fix that separately?

v-thakkar:
> Ah! @kakkoyun @v-thakkar I know what it is! The snap confinement is probably blocking access to '/boot'.

Ah, that makes sense. 

> I can follow up with another PR after this one has landed to fix that separately?

That'll be nice. Parca-agent should give a more useful error message in the case of an access issue though. I'll take care of it. Thanks.

kakkoyun:
@v-thakkar @jnsgruk I'll merge this soon. We can send the planned changes in subsequent PRs.

PR #686, opened 2022-08-08
Add API for the kernel config checks

### LINKED ISSUE #337: profiler: Add kernel feature checks at the startup
See https://github.com/parca-dev/parca-agent/issues/333#issuecomment-1094997990

### ACs:

* [ ] Check necessary Linux features[BPF, BTF] of the host
* [ ] Check if Kernel stack unwinding enabled or not https://github.com/torvalds/linux/blob/master/arch/x86/Kconfig.debug#L252

v-thakkar:
- Supersedes #586 
- Fixes #337


v-thakkar:
Rebased and reworked on API as well as test

pkg/profiler/cpu.go
```diff
@@ -314,13 +315,45 @@ func (p *CPUProfiler) Labels() model.LabelSet {
...
+}
+
 func (p *CPUProfiler) Run(ctx context.Context) error {
 	level.Debug(p.logger).Log("msg", "starting cpu profiler")
 
 	p.mtx.Lock()
 	ctx, p.cancel = context.WithCancel(ctx)
 	p.mtx.Unlock()
 
+	isContainer, err := kconfig.IsInContainer()
```
kakkoyun, line 331: Let's move these checks to "cmd/parca-agent/main". 
And just stop the program at the start-up time if it's a no go for us.

kakkoyun (commented):
Wow, amazingly well-written PR :100: !

Only one comment on the execution time of the checks. Otherwise LGTM!

pkg/profiler/cpu.go
v-thakkar: Just to be double sure, you mean here right after initializing parca-agent, right? https://github.com/parca-dev/parca-agent/blob/main/cmd/parca-agent/main.go#L124

kakkoyun: Yes, exactly!

kakkoyun (approved):

pkg/kconfig/kconfig.go
```diff
@@ -0,0 +1,160 @@
...
+const (
+	containerCgroupPath = "/proc/1/cgroup"
+)
+
+var ebpfCheckOptions = []string{
+	"CONFIG_BPF",
+	"CONFIG_BPF_SYSCALL",
+	"CONFIG_HAVE_EBPF_JIT",
+	"CONFIG_BPF_JIT",
+	"CONFIG_BPF_JIT_ALWAYS_ON",
```
javierhonduco, line 33: I don't think the BPF JIT is necessary, right? cc @Sylfrena

v-thakkar: Yeah not necessary but I think it'll be quite slow otherwise: https://elixir.bootlin.com/linux/latest/source/kernel/bpf/Kconfig#L40 So I personally think we should keep it as a mandatory check. But we can probably see how many users we come across where it doesn't enabled by default and can move it to 'nice to have' then?

javierhonduco: Sounds good

PR #689, opened 2022-08-08
feat: add snap packaging

### LINKED ISSUE #674: release: Automate releasing snap packages
xref: https://github.com/parca-dev/parca/pull/1488

jnsgruk:
Adds snap packaging to the goreleaser config, along with some CI setup to test/publish the snap on merges to main, or when publishing a release.

Will need some credentials adding before merging for the publishing to work! :)

Similar to https://github.com/parca-dev/parca/pull/1488 but works slightly differently. Because the agent is built in a container, and snapcraft cannot be installed in a container (Snapcraft is only delivered as a snap) I've had to make it a separate build step.

Fixes #674

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -243,5 +243,4 @@ jobs:
           SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.SNAPCRAFT_STORE_CREDENTIALS }}
         run: |
           snapcraft upload dist/*_amd64.snap --release stable
-          # amd64 releases only at the moment
-          # snapcraft upload dist/*_arm64.snap --release stable
+          snapcraft upload dist/*_arm64.snap --release stable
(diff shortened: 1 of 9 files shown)


jnsgruk:
@kakkoyun force push storm should be over :sweat_smile: I think this is good to go.

jnsgruk:
This is slightly different to how it was cracked for Parca, but I'm not convinced it isn't actually better because we're actually building the snap with snapcraft here. I might consider doing the same over in Parca sometime.

maxbrunet (commented):
Just a little detail I overlooked last time, please run the scripts in "strict mode" by adding `set -euo pipefail` at the top for each of them

jnsgruk:
> Just a little detail I overlooked last time, please run the scripts in "strict mode" by adding `set -euo pipefail` at the top for each of them

Done! Thanks, will retrospectively add to the other snap.

kakkoyun (approved):
LGTM

PR #701, opened 2022-08-11
pkg/kconfig: Check for BPF_JIT_DEFAULT_ON

v-thakkar:
Let's check for BPF_JIT_DEFAULT_ON instead of
BPF_JIT_ALWAYS_ON. This is more accurate as
some distros may have ARCH_WANT_DEFAULT_BPF_JIT
enabled and thus doesn't need BPF_JIT_ALWAYS_ON.

cc @Sylfrena @javierhonduco


javierhonduco (approved):
LGTM

Issue #644, opened 2022-08-01
optimization: Reduce the amount of memory the Agent uses


kakkoyun:
We need to recheck what we cache. And the lifecycle of the cached items. 

I will attach profiles from our demo environment to this issue.

javierhonduco:
Should we close this issue? It's based on pretty old profiles, and in general, we should always strive to keep the memory footprint of the agent as low as possible. We can open issues for specific issues that we spot that deserve fixing

Issue #393, opened 2022-05-05
optimization: Optimization after v0.8.0


kakkoyun:
This is how it looks after running the agent for 3 hours. I have added an interactive profile version and possible optimization areas with screenshots.

If anyone is interested, optimization games for parca-agent begins :stuck_out_tongue: 

Interactive version: https://share.polarsignals.com/8836b8d/

* Write Profile:
![Thu May  5 03:47:22 PM CEST 2022](https://user-images.githubusercontent.com/536449/166938412-5116f826-4dd6-4afc-8b65-6c1059bcbb45.png)

![Thu May  5 03:49:33 PM CEST 2022](https://user-images.githubusercontent.com/536449/166938667-3a8319b8-6ef3-4a92-89c3-3efc5c4a6e58.png)

* Build Profile:
![Thu May  5 03:48:05 PM CEST 2022](https://user-images.githubusercontent.com/536449/166938504-76155855-6cec-4110-a6b8-7621af670421.png)

* Ensure Uploaded:
![Thu May  5 03:48:36 PM CEST 2022](https://user-images.githubusercontent.com/536449/166938566-a52e7801-28df-4749-9e5b-ce1d3fbd9d24.png)

kakkoyun:
Related #391 

kysm is not the biggest bottleneck we have right now.

cc @javierhonduco

javierhonduco:
Thanks for doing this!

The unknown frames seem odd, let's investigate what it is and why it can't symbolise
<img width="1461" alt="image" src="https://user-images.githubusercontent.com/959128/166965120-f31feae1-b788-40e6-b28a-8c54907a16d8.png">

kakkoyun:
Deprecated.

PR #708, opened 2022-08-16
Kernel symbolication fixes

javierhonduco:
This PR introduces two changes:
1. Calling the Kernel symbolizer (and the user function symboliser for JITs), which got lost in https://github.com/parca-dev/parca-agent/pull/663
2. Dealing with unsorted `/proc/kallsyms`

More details can be found in the respective commit messages

### Before these changes
image

### After these changes
- Added multiple tests for the Kernel symboliser fixes.

- Run e2e tests
image

image


pkg/ksym/ksym.go
```diff
@@ -140,66 +150,64 @@ func (c *cache) Resolve(addrs map[uint64]struct{}) (map[uint64]string, error) {
 	return res, nil
 }
 
-func unsafeString(b []byte) string {
```
javierhonduco, line 143: Perhaps I am missing something, but I think this wasn't needed

pkg/symbol/symbol.go
```diff
@@ -62,16 +62,16 @@ func (s *Symbolizer) Symbolize(prof *profiler.Profile) error {
...
-	userFunctions, err := s.resolveJITedFunctions(pid, prof.UserLocations)
+	userJITedFunctions, err := s.resolveJITedFunctions(pid, prof.UserLocations)
 	if err != nil {
 		// We expect only a minority of processes to have a JIT and produce the perf map.
 		if !errors.Is(err, perf.ErrNotFound) {
 			level.Debug(s.logger).Log("msg", "failed to obtain perf map for pid", "pid", pid, "err", err)
 			return nil
 		}
-		return fmt.Errorf("failed to resolve user functions: %w", err)
+		return fmt.Errorf("failed to resolve user JITed functions: %w", err)
```
javierhonduco, line 72: Added this to clarify that this is not a 'normal' user address but it's a JIT'ed one

pkg/ksym/ksym.go
kakkoyun: cc @brancz

pkg/profiler/cpu/cpu.go
```diff
@@ -242,6 +242,11 @@ func (p *CPU) Run(ctx context.Context) error {
 		}
 
 		for _, prof := range profiles {
+			err = p.symbolizer.Symbolize(prof)
```
kakkoyun, line 245: 😱 I wonder how did I manage to remove this one??!

pkg/ksym/ksym_test.go
```diff
@@ -54,56 +59,108 @@ ffffffff8f6d1600 b xfrm_state_afinfo
...
 ffffffff8f6d1770 b xfrm_state_gc_list
 ffffffff8f6d1780 b xfrm_napi_dev
+ffffffff8f6d15c4 b not_in_order
 		`),
 			}))
 
-	addr1 := uint64(0xffffffff8f6d14a4) + 1
-	addr2 := uint64(0xffffffff8f6d15e0) + 1
-	addr3 := uint64(0xffffffff8f6d1480) + 1
+	addr1 := uint64(0xffffffff8f6d14a4)
```
kakkoyun, line 66: Why did you remove +1 from here and add it to every call?
An explanation would be awesome. (also  as a comment)

pkg/ksym/ksym.go
```diff
@@ -52,8 +57,12 @@ func NewKsymCache(logger log.Logger, reg prometheus.Registerer, f ...fs.FS) *cac
...
 	}
 	return &cache{
-		logger:         logger,
-		fs:             fs,
+		logger: logger,
+		fs:     fs,
+		// My machine has ~200k loaded symbols. Reserving 50k entries, as there might be
+		// boxes with fewer symbols loaded, and if we need to reallocate, we would have
+		// to do it once or twice, so this value seems like a reasonable middle ground.
+		kernelSymbols:  make([]ksym, 0, 50000),
```
kakkoyun, line 65: What's the amount of memory pressure that this one puts in the worst-case?

kakkoyun (commented):
Definitely a great catch 👍 

I have a couple of questions though.

pkg/ksym/ksym_test.go
javierhonduco: I found it confusing as the addresses themselves were at the offset zero of the functions we wanted to search for. I wanted to cover both intermediate ranges (such as `start_address` + `n` (where `n` is 1 in this case), as well as 'exact' matches when the requested address is the `start_address`.

Hope this makes sense. Will add some comments to the test cases

pkg/ksym/ksym.go
javierhonduco: Each Ksym is 24B in size + the string overhead, a back-of-the-envelop calculation would be something like:

`50000  (number of ksyms) * (24B (size of the address and string metadata) + 38 characters (P90 length symbols in my box) * 8Bytes/character` = ~ 16400000Bytes = ~16MB 

This is some rough calculation, but I think it should be representative.

I've also added a unittest to ensure we don't add extra fields to the ksym structure without noticing: https://github.com/parca-dev/parca-agent/pull/708/files#diff-694cd04c70fb74c102d9e61bc2c373a0d8df849095e5ddc439af8f2cb5372f40R28

pkg/ksym/ksym_test.go
```diff
@@ -24,86 +25,142 @@ import (
...
-		addr1: {},
-		addr2: {},
+		addr1 + 1: {},
+		addr2 + 1: {},
 	})
 	require.NoError(t, err)
 	require.Equal(t, map[uint64]string{
-		addr1: "tcp_sock_id",
-		addr2: "xfrm_km_lock",
+		addr1 + 1: "tcp_sock_id", /////
```
javierhonduco, line 78: Note to self: remove the trailing slashes :/

pkg/ksym/ksym.go
kakkoyun: Let's add this as a comment as well. I'm fine to add this because we spent a lot of time checking the kernel symbols. If I'm not mistaken, it should've been mentioned in #393.

But also we have #644 to reduce the memory. So let's add this as a comment and keep an eye on this one. I'm sure we have many other things to optimise until it comes to this.

javierhonduco: Sounds good. Let's make sure that we have the right data and we can always optimise things down the line.


By the way, not sure if you saw the comment about the kernel symbol cache. I am worried about it growing in size a lot as it's an unbounded map. I am planning to add a bounded LRU cache there in another PR.

kakkoyun: Just go for it. IT would be nice if you can create issues for the things that you discovered.

kakkoyun:
@javierhonduco Let me know when you're done, so that I can approve it.

pkg/ksym/ksym.go
brancz: premature optimization, only do this if we have measurable impact

pkg/ksym/ksym.go
```diff
@@ -140,66 +154,79 @@ func (c *cache) Resolve(addrs map[uint64]struct{}) (map[uint64]string, error) {
...
-
 	s := bufio.NewScanner(fd)
-	lastSym := ""
+	symbol := ""
 
 	for s.Scan() {
-		l := s.Bytes()
+		line := s.Bytes()
 
-		curAddr, err := strconv.ParseUint(unsafeString(l[:16]), 16, 64)
```
kakkoyun, line 165: I think here `unsafeString` is used to prevent additional allocations for the interim string! This might help https://chris124567.github.io/2021-06-21-go-performance/ (I couldn't find a better resource, I need to dig more!)

kakkoyun (commented):
Looks good!

I've added a comment about `unsafeString`. Let's wait for @brancz to validate that claim.

I don't want to drag the improvement, but how about writing a benchmark test using the symbol file on your machine and attaching the results to the commit? 

https://pkg.go.dev/golang.org/x/tools/cmd/benchcmp
https://betterprogramming.pub/comparing-two-benchmarks-in-go-1-18-e6161de3757

It would clarify how much memory we're trading for the CPU time.

pkg/ksym/ksym.go
javierhonduco: Realised it was to avoid allocations, but thought that the unsafe wasn't worth it for the performance improvements. While the improvement is far from minimal, it's trumped by the rest of the code that does perform way more allocations and uses more CPU cycles to operate.

Added some basic benchmarks in the new commit:
```
[javierhonduco@fedora parca-agent]$ go test -v -bench=. -benchmem github.com/parca-dev/parca-agent/pkg/ksym
=== RUN   TestEnsureKsymSizeDoesNotGrow
--- PASS: TestEnsureKsymSizeDoesNotGrow (0.00s)
=== RUN   TestKsym
--- PASS: TestKsym (0.00s)
goos: linux
goarch: amd64
pkg: github.com/parca-dev/parca-agent/pkg/ksym
cpu: Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz
BenchmarkStringParsing
BenchmarkStringParsing-12        2184770               727.6 ns/op           384 B/op         24 allocs/op
BenchmarkUnsafeString
BenchmarkUnsafeString-12        106100638               11.14 ns/op            0 B/op          0 allocs/op
PASS
ok      github.com/parca-dev/parca-agent/pkg/ksym       3.853s
```

kakkoyun (approved):
LGTM.

Let's see this in action, and we can iterate on it if it's necessary.

PR #711, opened 2022-08-16
Ksym cache: Use a bounded cache

javierhonduco:
Rather than an unlimited one, which has the potential of creating too much memory pressure.

When the cache is full, finding an element is cheaper than it used to thanks to https://github.com/parca-dev/parca-agent/pull/708, as:
- We don't have to read the file nor parse it every single time
- And most importantly, we swapped the linear search with a binary search

### DIFF
--- a/pkg/ksym/ksym.go
+++ b/pkg/ksym/ksym.go
@@ -25,11 +25,14 @@ import (
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
+	burrow "github.com/goburrow/cache"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 
 	"github.com/parca-dev/parca-agent/pkg/hash"
 )
+
+const KsymCacheSize = 10_000 // Arbitrary cache size.
 
 type ksym struct {
 	address uint64
@@ -43,7 +46,7 @@ type cache struct {
 	lastHash              uint64
 	lastCacheInvalidation time.Time
 	updateDuration        time.Duration
-	fastCache             map[uint64]string
+	cache                 burrow.Cache
 	mtx                   *sync.RWMutex
 	cacheFetch            *prometheus.CounterVec
 }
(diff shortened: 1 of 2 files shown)


pkg/ksym/ksym_test.go
```diff
@@ -168,11 +168,6 @@ ffffffff8f6d15c4 a not_in_order
 		addr2 + 1: "xfrm_km_lock",
 	}, syms)
 
-	require.Equal(t, map[uint64]string{
-		addr1 + 1: "tcp_sock_id",
-		addr2 + 1: "xfrm_km_lock",
-	}, c.fastCache)
```
javierhonduco, line 174: Removing it and the below, as this is testing an implementation detail. We already make sure that we hit the cache in a test for this purpose

pkg/ksym/ksym.go
```diff
@@ -120,13 +127,14 @@ func (c *cache) Resolve(addrs map[uint64]struct{}) (map[uint64]string, error) {
...
 	for addr := range addrs {
-		sym, ok := c.fastCache[addr]
+		sym, ok := c.cache.GetIfPresent(addr)
 		if !ok {
 			notCached = append(notCached, addr)
 			c.cacheFetch.WithLabelValues("miss").Inc()
 			continue
 		}
-		res[addr] = sym
+		//nolint:forcetypeassert
```
kakkoyun, line 136: nit: We can check if the type assertion is correct and, if not, return an error. This way, it won't panic if we change the type of things we put in the cache. In any case, this is unlikely to happen considering the locality of this, hence this is a nit-pick.

pkg/ksym/ksym_test.go
```diff
@@ -168,11 +168,6 @@ ffffffff8f6d15c4 a not_in_order
 		addr2 + 1: "xfrm_km_lock",
 	}, syms)
 
-	require.Equal(t, map[uint64]string{
```
kakkoyun, line 171: nit: Maybe it won't be exact but we can still have equivalent test cases for this case.

kakkoyun (approved):
LGTM.

I only have some nits.

pkg/ksym/ksym.go
javierhonduco: Good idea!

pkg/ksym/ksym_test.go
javierhonduco: I would rather not re-add these as they are implementation details that should be fully covered by this test

https://github.com/parca-dev/parca-agent/blob/2668fa07c12821f9aeed7e1ec54a01aad765369b/pkg/ksym/ksym_test.go#L232-L243

Happy to augment that test or add more to make sure that the functionality is correct despite of the cache layer, if you think that the current tests don't fully cover it at the moment

kakkoyun: True, your point is valid.

PR #713, opened 2022-08-16
Remove force type asserts linter exclusions

javierhonduco:
And log / error instead in case of errors

cc @kakkoyun


kakkoyun (approved):
LGTM

PR #716, opened 2022-08-17
pkg/profiler/cpu: Lower log level when failing to symbolize

brancz:
This error is not preventing us from shipping data to a storage, so seeing this error in logs is more confusing to users than helpful. When debugging it might be useful though so lowering it to debug level.


v-thakkar (approved):
:100:

kakkoyun (approved):
LGTM 🚀

kakkoyun:
cc @javierhonduco

PR #717, opened 2022-08-17
pkg/perf: Clarify the not found error is about perf-maps

brancz:
I tried to understand what a log line was trying to tell me and I kept thinking that the Parca Agent was failing to find symbols for something, when it was really just not finding a perf-map for that process, which is perfectly fine since most processes don't have jitted symbols.


kakkoyun (approved):

kakkoyun:
cc @javierhonduco

Issue #687, opened 2022-08-08
Add more kernel config checks


v-thakkar:
This is for the follow-up PR after #686 has been merged.
 - [ ] Check for the stack unwinding-related options
 - [ ] Check for the CONFIG_SYS_ADMIN, CAP_PERF_MON (for perf_events)
 - [ ] Check for Cgroup versions

kakkoyun:
For capability checks: https://github.com/aquasecurity/tracee/tree/main/pkg/capabilities

palash25:
i would like to take this up if no one is working on this

PR #719, opened 2022-08-17
Check that the user running the Agent is root

javierhonduco:
Some profilers that also require root, such as rbperf, check for the superuser to ensure that they can load BPF programs, among other things.

```
[javierhonduco@fedora rbperf]$ ./rbperf record -p $RUBY_PID cpu
Error: rbperf requires root to load and run BPF programs
```

Parca Agent doesn't run any checks and the output can be confusing to newcomers, so let's fix this :)

## Current behaviour
```
$ ./parca-agent
[...]
level=info name=parca-agent ts=2022-08-17T10:05:14.304168124Z caller=main.go:203 msg="local profile storage is enabled" dir=./tmp/profiles
name=parca-agent ts=2022-08-17T10:05:14.304188383Z caller=main.go:210 msg=starting... node=test store=
level=warn name=parca-agent ts=2022-08-17T10:05:14.304938858Z caller=discovery_manager.go:221 msg="unable to start provider" provider=systemd/0 error="context canceled"
level=error name=parca-agent ts=2022-08-17T10:05:14.304955172Z caller=main.go:121 err="new bpf module: error setting rlimit: operation not permitted"
```

## With this commit

```
$ ./parca-agent
[...]
level=error name=parca-agent ts=2022-08-17T10:12:11.179985242Z caller=main.go:121 err="superuser (root) is required to run Parca Agent"
```

### DIFF
--- a/cmd/parca-agent/main.go
+++ b/cmd/parca-agent/main.go
@@ -122,7 +122,15 @@ func main() {
 	}
 }
 
+func isRoot() bool {
+	return os.Geteuid() == 0
+}
+
 func run(logger log.Logger, reg *prometheus.Registry, flags flags) error {
+	if !isRoot() {
+		return errors.New("superuser (root) is required to run Parca Agent")
+	}
+
 	isContainer, err := kconfig.IsInContainer()
 	if err != nil {
 		level.Warn(logger).Log("msg", "failed to check if running in container", "err", err)


kakkoyun (changes requested):
I remember discussing this issue with @v-thakkar.
We were thinking of adding some capability checks rather than checking for the root user.
Because users can use those APIs if they have the right capability, and the user doesn't need to be the root.

Correct me if I'm wrong @v-thakkar ?

kakkoyun:
ref https://github.com/parca-dev/parca-agent/issues/687

v-thakkar:
Yes, that's correct. So instead of root, the plan was to check for CAP_SYS_ADMIN and CAP_BPF(since 5.8). I was planning to send to include that check with other remaining checks but feel free to send a separate PR for it if you would like @javierhonduco :)

javierhonduco:
Good shout, I just verified that Parca Agent works well with the 3 necessary capabilities:

```
sudo setcap cap_bpf,cap_sys_resource,cap_sys_admin=+ep ./dist/parca-agent 
```

As you mentioned, if we are only supporting Kernels that have or know how to handle these capabilities, that's great. Otherwise, this PR could be a reasonable way to approach it. I can imagine most people would run Parca Agent from a superuser anyways.


All yours @v-thakkar 😄

v-thakkar:
Ah, checking CAP_SYS_RESOURCE will be nice too. Noted :+1:

PR #721, opened 2022-08-18
Add time expiration on debuginfo cache

Pryz:
Fixes https://github.com/parca-dev/parca/issues/1543

This PR an expiration time of 5min on the debuginfo upload cache as well as two counters to observe the upload behavior.


kakkoyun (commented):
I'm ok to adding an expiration limit for the cache but we miss it to remove the uploading state in the failure scenario.

Let's add it as well.

BTW, amazing touch with the metrics 🤗

javierhonduco:
Thanks for taking a look! I can see how this alleviates the issue, but I think we should fix the root problem. For context, this mechanism was introduced to prevent races whilst processing debuginfos, as we can potentially have race conditions. We used a cache because it was convenient and thread-safe, but we are just using it as locking mechanism.

`removeAsUploading()` should be called at the end of `EnsureUploaded()` and it will remove the buildID from `uploadingCache`, no? If so, why isn't the debug info being re-uploaded if everything works ok?

https://github.com/parca-dev/parca-agent/blob/51a1008e3873d4d14150b300939468c176b067b2/pkg/debuginfo/manager.go#L139

https://github.com/parca-dev/parca-agent/blob/51a1008e3873d4d14150b300939468c176b067b2/pkg/debuginfo/manager.go#L93-L95


What might be happening is this:
1. We add an element to the 'uploading' cache
2. Perhaps we hit one of the conditions that cause the loop iteration to be skipped
3. `removeAsUploading` is never called
4. Parca server is re-started
5. We haven't invalidated the 'uploading' cache, so we see there's some in-flight upload, but in reality there is none, so we don't re-upload the file


I think a better fix will be to ensure that we call `removeAsUploading()`at the end of each iteration of the loop even if we have an early `continue`

What do you think?

Pryz:
mmm looks to me that `removeAsUploading` will never be called in the `buildID` is in the cache. See: https://github.com/parca-dev/parca-agent/blob/51a1008e3873d4d14150b300939468c176b067b2/pkg/debuginfo/manager.go#L103-L105.

We could ensure that we remove the `buildID` from the cache at the end of each iteration if that is the behavior we are expecting. The cleanest will probably to defer the work done inside the loop in another function and call `defer di.removeAsUploading(buildID)`.

Let me know if that is the behavior we are expecting :)

pkg/debuginfo/manager.go
```diff
@@ -97,47 +104,56 @@ func (di *Manager) removeAsUploading(buildID string) {
...
-		}
+	logger := log.With(di.logger, "buildid", buildID, "path", path)
 
-		buf := flexbuf.New()
-		if err := di.Extract(ctx, buf, src); err != nil {
-			level.Debug(di.logger).Log("msg", "failed to extract debug information", "err", err, "buildID", buildID, "path", src)
-			continue
-		}
+	// removing the buildID from the cache to ensure a re-upload at the next interation.
+	defer di.removeAsUploading(buildID)
```
kakkoyun, line 118: 👍

pkg/debuginfo/manager.go
```diff
@@ -97,47 +104,56 @@ func (di *Manager) removeAsUploading(buildID string) {
...
 
-		buf.SeekStart()
-		if err := validate(buf); err != nil {
-			level.Debug(logger).Log("msg", "failed to validate debug information", "err", err, "buildID", buildID, "path", src)
-			continue
-		}
+	if di.alreadyUploading(buildID) {
+		return
```
kakkoyun, line 121: Above we defer `removeAsUploading` in this case we will mark this build id as finished uploading prematurely.

kakkoyun: I think defer should move after this condition.

pkg/debuginfo/manager.go
```diff
@@ -97,47 +104,56 @@ func (di *Manager) removeAsUploading(buildID string) {
...
+	}
+	di.markAsUploading(buildID)
```
kakkoyun, line 123:
```suggestion
	di.markAsUploading(buildID)
	// removing the buildID from the cache to ensure a re-upload at the next interation.
 	defer di.removeAsUploading(buildID)
```

kakkoyun (changes requested):
Looking really good. I pointed a possible issue.

pkg/debuginfo/manager.go
Pryz: Thanks for the review ! I'm actually wondering if the condition makes sense anymore.

When a new buildid is discovered, we will go through the entire upload path, the buildid will be added to the cache and due to the defer removed immediately. During the next iteration, `di.alreadyUploading` will be false and will always be false now. Right ? Or am I missing something ?

kakkoyun: This process is async that's the reason we added this cache-based lock in the first place (burrow cache is thread-safe). So we want to prevent re-uploading if there's already an upload in progress. So we should only remove the uploading state if we marked that build id as uploading.

Pryz: Ha yes sorry just saw we are calling this function from a go routine ! Thanks :)

kakkoyun (approved):

javierhonduco:
Thanks for working on this fix! :)

PR #734, opened 2022-08-22
fix: ability to specify alt kernel config options

jnsgruk:
Currently, the kernel config is checked for a static
list of kernel config options pertaining to eBPF.

This change allows the developer to specify
alternatives to each option in cases where there are
valid fallbacks.


pkg/kconfig/kconfig.go
```diff
@@ -63,22 +69,51 @@ func IsBPFEnabled() (bool, error) {
...
+					if altFound {
+						break
+					}
+				}
+				if !altFound {
+					// If we reach this point, we were unable to verify the presence of any alternatives
+					alts := strings.Join(option.alternatives, ", ")
+					return false, fmt.Errorf("%w; alternatives checked:%s", err, alts)
+				}
+			} else {
```
kakkoyun, line 110: I would instead negate the condition, return early with this case, and eliminate the need for `else`.
It would make it easier to read. We're already in the 3rd level. (for, if, if). 

Similarly, we can first check for `found` and return even from the function.

kakkoyun (commented):
LGTM.

I've seen only one thing related to style to improve readbility.

pkg/kconfig/kconfig.go
jnsgruk: Ah yeh, fair enough. Will amend shortly.

jnsgruk:
> LGTM.
> 
> I've seen only one thing related to style to improve readbility.

Alright, I think the latest change gets it a bit closer. I don't think we can bail any earlier than we do, because we're checking for the existence of *any* of the alternatives for each option, not *all* if that makes sense?

pkg/kconfig/kconfig.go
```diff
@@ -25,13 +25,19 @@ const (
...
+	// Used for specifying synonymous kernel options
+	alternatives []string
+}
+
+var ebpfOptions = []ebpfOption{
+	{name: "CONFIG_BPF"},
+	{name: "CONFIG_BPF_SYSCALL"},
+	{name: "CONFIG_HAVE_EBPF_JIT"},
+	{name: "CONFIG_BPF_JIT"},
+	{name: "CONFIG_BPF_JIT_ALWAYS_ON", alternatives: []string{"CONFIG_BPF_JIT_DEFAULT_ON"}},
```
v-thakkar, line 39: Actually, we should just have ARCH_WANT_DEFAULT_BPF_JIT instead of CONFIG_BPF_JIT_DEFAULT_ON as an alternative. That will handle configs in other archs and older kernels too. So, while you're at it, can you please change this? Thanks.

v-thakkar:
Thanks for handling this. Just have one comment, looking good otherwise. :)

pkg/kconfig/kconfig.go
jnsgruk: Done!

v-thakkar:
:tada:

PR #740, opened 2022-08-23
chore: parametrize DOCKER_SOCK in Makefile

maxbrunet:
I use Podman in rootless mode, so `DOCKER_SOCK` is `/run/user/1000/podman/podman.sock`.

(I have not been able to build these targets tho, which is required to build a local container, goreleaser+Makefile has given me a headache, giving up for now, we might want to use `goreleaser build --rm-dist --single-target --snapshot --debug` in the `make build`)


kakkoyun:
> (I have not been able to build these targets tho, which is required to build a local container, goreleaser+Makefile has given me a headache, giving up for now, we might want to use `goreleaser build --rm-dist --single-target --snapshot --debug` in the `make build`)

It would be nice to unify how we build locally and for the releases. Ultimately it is a developer experience concern. As far as I observe, most of the contributors don't bother to install goreleaser (or any additional tooling) if they're working on something. We need to find the right balance and don't make it harder than necessary to contribute. 

We have created https://github.com/parca-dev/cross-builder, and we should have moved builder-related mechanisms to that repo and built a pipeline to release these builder images automatically. We haven't had time to dedicate to it. Feel free to take the initiative if you've free spins ;)

kakkoyun (approved):
LGTM 🚀

PR #747, opened 2022-08-24
.eh_frame: Add Parca Demo's test cpp binary and output

javierhonduco:
As discussed in the meeting, the first milestone for the .eh_frame unwinding project is to be able to unwind the cpp binary we use for testing.

I am adding it here to make testing more straightforward, as well as the generated tables. They don't include the offset information yet, but it's still valuable to ensure that the DWARF parsing is correct as we continue developing this feature.

Later on, we can add more test binaries, with their respective tables, generate them in CI and ensure that they are the same that we have committed (this technique is called snapshot/approval tests)


v-thakkar:
Thanks :+1:

PR #754, opened 2022-08-25
Use executeDwarfProgram

v-thakkar:


internal/dwarf/frame/table.go
```diff
@@ -219,7 +219,10 @@ func executeCIEInstructions(cie *CommonInformationEntry) *FrameContext {
 		instructions: frames,
 		buf:          bytes.NewBuffer(initialInstructions),
 	}
-	//frame.executeDwarfProgram()
+	//TODO: Uncommenting this as this gets us the correct first row
```
kakkoyun, line 222: @v-thakkar I think this should be there regardless of changes in `executeDWARFProgram` since this is to parse the CIE.

Issue #750, opened 2022-08-25
perf-map not found from latest parca-agent version


Pryz:
I wanted to test the debuginfo caching fix but after deploying the agent from the `main` branch I'm hitting the following error:
```
failed to resolve user JITed functions: perf-map not found
``` 

From https://github.com/parca-dev/parca-agent/blob/main/pkg/perf/perf.go#L141 looks like we are expecting a perf map file but I'm not sure where this file is supposed to come from. I don't see any `.map` file on the instances I'm operating.

Also looks like we are going to return perf-map not found error as well if the procfs status file doesn't exist which might be confusing during troubleshooting.

maxbrunet:
The runtime needs to be configured to create it. Some examples can be found here: https://github.com/parca-dev/parca-demo.

E.g for NodeJS, it is the `--perf-basic-prof` flag.

Pryz:
Was reading the code perf map are actually not going to be used on my setup since I'm mostly profiling Go binaries. So mostly a red herring. I think my issue is due to the target discovery instead.

brancz:
What log level was this at? Either way, I think we should make it clear in the log line that most processes probably won't have perf-maps so it's safe to ignore. But we should also make sure this is a debug level log.

maxbrunet:
I think the condition is simply inverted here, we do not want to return error if not found, but log it instead:

https://github.com/parca-dev/parca-agent/blob/4b7e8082dcc77bcaa0aa32373ff089df828114c2/pkg/symbol/symbol.go#L67-L72

I am fixing it in #755

PR #755, opened 2022-08-25
fix: add more context to profile log messages

maxbrunet:
* Add `PID` (and `addr` if available) to profile log messages, this will help to investigate processes the agent is having trouble with
* Add an empty target to systemd target groups, this will look less odd in some places like the `parca_agent_sd_discovered_targets{config="all"}` metric reporting `0` targets
* ~~Fix ignoring perf map not found errors (#750)~~

### DIFF
--- a/pkg/symbol/symbol.go
+++ b/pkg/symbol/symbol.go
@@ -65,7 +65,7 @@ func (s *Symbolizer) Symbolize(prof *profiler.Profile) error {
 	userJITedFunctions, err := s.resolveJITedFunctions(pid, prof.UserLocations)
 	if err != nil {
 		// We expect only a minority of processes to have a JIT and produce the perf map.
-		if !errors.Is(err, perf.ErrNotFound) {
+		if errors.Is(err, perf.ErrNotFound) {
 			level.Debug(s.logger).Log("msg", "failed to obtain perf map for pid", "pid", pid, "err", err)
 			return nil
 		}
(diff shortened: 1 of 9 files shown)


pkg/symbol/symbol.go
```diff
@@ -90,10 +90,7 @@ func (s *Symbolizer) resolveJITedFunctions(pid profiler.PID, locations []*profil
 		if !ok {
 			sym, err := perfMap.Lookup(loc.Address)
 			if err != nil {
-				if !errors.Is(err, perf.ErrNotFound) {
```
maxbrunet, line 93: `Lookup()` only only returns an `ErrNoSymbolFound` error

kakkoyun: I think the `ErrNoSymbolFound` sentinel error only gets checked here. Let's remove the error if we decide not to check against it.

kakkoyun (changes requested):
Good catch!

kakkoyun:
Let's reconsider this since #756 also fixes the same problem.

maxbrunet:
> Let's reconsider this since #756 also fixes the same problem.

@kakkoyun I've removed the overlapping changes

pkg/symbol/symbol.go
maxbrunet: For the record, I think it is a best practice to declare all errors, so leaving `ErrNoSymbolFound` even if nothing catches it at the moment is good

I usually enable the [go-err113](https://github.com/Djarvur/go-err113) linter which disallow dynamic errors.

kakkoyun (approved):

pkg/symbol/symbol.go
kakkoyun: We can declare it, but we don't need to export it. If the identifier is exported, it gives me the sense that it's used outside the package.

PR #756, opened 2022-08-26
perf: split perf-map and proc status not found errors

Pryz:
As I was debugging https://github.com/parca-dev/parca-agent/issues/750 I realized
that we are returning the same error when checking if the map file is present
but also the proc status file which might be confusing during troubleshooting.

### DIFF
--- a/pkg/symbol/symbol.go
+++ b/pkg/symbol/symbol.go
@@ -65,7 +65,7 @@ func (s *Symbolizer) Symbolize(prof *profiler.Profile) error {
 	userJITedFunctions, err := s.resolveJITedFunctions(pid, prof.UserLocations)
 	if err != nil {
 		// We expect only a minority of processes to have a JIT and produce the perf map.
-		if !errors.Is(err, perf.ErrNotFound) {
+		if !errors.Is(err, perf.ErrPerfMapNotFound) {
 			level.Debug(s.logger).Log("msg", "failed to obtain perf map for pid", "pid", pid, "err", err)
 			return nil
 		}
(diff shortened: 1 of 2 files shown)


pkg/symbol/symbol.go
```diff
@@ -65,7 +65,7 @@ func (s *Symbolizer) Symbolize(prof *profiler.Profile) error {
 	userJITedFunctions, err := s.resolveJITedFunctions(pid, prof.UserLocations)
 	if err != nil {
 		// We expect only a minority of processes to have a JIT and produce the perf map.
-		if !errors.Is(err, perf.ErrNotFound) {
+		if !errors.Is(err, perf.ErrPerfMapNotFound) {
```
maxbrunet, line 68: I think the negation here is a mistake
```suggestion
		if errors.Is(err, perf.ErrPerfMapNotFound) {
```

pkg/symbol/symbol.go
```diff
@@ -90,7 +90,7 @@ func (s *Symbolizer) resolveJITedFunctions(pid profiler.PID, locations []*profil
 		if !ok {
 			sym, err := perfMap.Lookup(loc.Address)
 			if err != nil {
-				if !errors.Is(err, perf.ErrNotFound) {
+				if !errors.Is(err, perf.ErrPerfMapNotFound) {
 					continue
 				}
```
maxbrunet, line 95: This condition is not necessary, `Lookup()` only returns `ErrNoSymbolFound`

```suggestion
```

maxbrunet (commented):
I deal with some of that in #755, but more granular errors is a good idea

pkg/symbol/symbol.go
Pryz: Good catch !

pkg/symbol/symbol.go
Pryz: Good catch ! ^ 2

kakkoyun (approved):
LGTM. Thanks for fixing my refactoring mistakes :see_no_evil:

PR #766, opened 2022-08-31
.eh_frame: Add $rbp

javierhonduco:
We need the `rbp` offset to CFA to be able to unwind the stack using .eh_frame. I've tried to use the same names as readelf does, so we print `u` when the offset is zero, otherwise, we print the offset (both relative to the CFA).

## Test plan
More testing has to be done (will detail this in an issue soon) but made sure that the output for our CPP test binary is correct. Created to commits so the changes are more easily distinguishable


javierhonduco:
Copy registers properly rather than just one

internal/dwarf/frame/table.go
```diff
@@ -9,6 +9,9 @@ import (
 	"github.com/parca-dev/parca-agent/internal/dwarf/util"
 )
 
+// TODO(javierhonduco): Add source
+const RBPRegister = 6
```
kakkoyun, line 13: Better to use https://github.com/go-delve/delve/tree/master/pkg/dwarf/regnum

pkg/stack/unwind/plan_table.go
```diff
@@ -243,24 +254,46 @@ func buildTableRows(fde *frame.DescriptionEntry, start uint64) []PlanTableRow {
...
+					row.RA = Instruction{Op: OpUnimplemented}
+				}
+			} else {
 				row.RA = Instruction{Op: OpUnimplemented}
 			}
-		} else {
-			row.RA = Instruction{Op: OpUnimplemented}
+		}
+
+		// Deal with $rbp
```
kakkoyun, line 281: :+1: :heart:

kakkoyun (approved):
LGTM.

I've added a link for subsequent PRs.

javierhonduco:
Resolve feedback

internal/dwarf/frame/table.go
javierhonduco: Thanks so much! Did not look in the right places

PR #767, opened 2022-08-31
feat(ui): bring back target list

maxbrunet:
Work in progress

image

### DIFF
--- a/cmd/parca-agent/main.go
+++ b/cmd/parca-agent/main.go
@@ -22,7 +22,9 @@ import (
 	"net"
 	"net/http"
 	"net/http/pprof"
+	"net/url"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -38,6 +40,8 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/collectors"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/promql/parser"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
@@ -98,6 +102,7 @@ type Profiler interface {
 	Run(ctx context.Context) error
 
 	LastProfileStartedAt() time.Time
+	LastProcessErrors() map[int]error
 	LastError() error
 }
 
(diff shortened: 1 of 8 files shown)


kakkoyun (commented):
Looking good :+1: 
Thanks for picking this up!

cmd/parca-agent/main.go
```diff
@@ -296,24 +306,170 @@ func run(logger log.Logger, reg *prometheus.Registry, flags flags) error {
...
-					Interval:       flags.ProfilingDuration,
 					NextStartedAgo: time.Since(profiler.LastProfileStartedAt()),
 					Error:          profiler.LastError(),
 				})
+
+				processReports[profiler.Name()] = profiler.ProcessReports()
 			}
-			err := template.StatusPageTemplate.Execute(w, statusPage)
+
+			procDirEntries, err := os.ReadDir("/proc")
```
kakkoyun, line 326: Even though it does the same thing under the hood, I would love to unify things and use them https://pkg.go.dev/github.com/keybase/go-ps?utm_source=gopls#Processes to discover processes.

cmd/parca-agent/main.go
```diff
@@ -169,29 +175,38 @@ func run(logger log.Logger, reg *prometheus.Registry, flags flags) error {
...
+
+		g                okrun.Group
+		batchWriteClient = agent.NewBatchWriteClient(logger, profileStoreClient, flags.RemoteStoreBatchWriteInterval)
+		profileListener  = agent.NewMatchingProfileListener(logger, batchWriteClient)
+		profileWriter    profiler.ProfileWriter
+	)
 
-		batchWriteClient := agent.NewBatchWriteClient(logger, profileStoreClient, flags.RemoteStoreBatchWriteInterval)
-		profileWriter = profiler.NewRemoteProfileWriter(agent.NewMatchingProfileListener(logger, batchWriteClient))
+	if flags.LocalStoreDirectory != "" {
```
kakkoyun, line 205: I think we should prioritize the remote writer over the local one. Or even better, if both are specified, we should have a writer to write both of them. We don't necessarily need to handle it in this PR, however, it would be great.

cmd/parca-agent/main.go
```diff
@@ -296,24 +306,170 @@ func run(logger log.Logger, reg *prometheus.Registry, flags flags) error {
 		}
 		if r.URL.Path == "/" {
 			w.Header().Set("Content-Type", "text/html; charset=utf-8")
-			statusPage := template.StatusPage{}
+			statusPage := template.StatusPage{
+				ProfilingInterval:   flags.ProfilingDuration,
+				ProfileLinksEnabled: flags.LocalStoreDirectory == "",
```
kakkoyun, line 311: Let's introduce a variable for this state. `profileLinksEnabled` or `localStorageEnabled` and use this variable for the rest of the code rather than checking the flag equivance to an empty string. It would be easier to make sense of.

kakkoyun (commented):
Amazing PR! I have just noted a couple of nitpicks.

pkg/profiler/cpu/cpu.go
```diff
@@ -86,6 +86,7 @@ type CPU struct {
 
 	// Reporting.
 	lastError                      error
+	processReports                 map[int]error
```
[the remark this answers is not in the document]
maxbrunet: I initially had in mind to use a `struct` and to allow including other things, but we can rename again if the need ever arises

kakkoyun (approved):

PR #785, opened 2022-09-06
.eh_frame: Carry over the initial and previous registers

javierhonduco:
Addresses a good portion of the issues in table generation that affected our test libc. Also, pass the length hint to the map to avoid reallocations + unnecessary copying on map growth.


kakkoyun (approved):
LGTM

PR #787, opened 2022-09-06
dwarf/frame: Add missing instructions

kakkoyun:
= Adds missing implementations for DWARF frame instructions.

```
DW_CFA_MIPS_advance_loc8		MIPS-specific. Implementation is provided regardless.
DW_CFA_GNU_window_save			SPARC-specific. Unsupported architecture. Read and skip.
DW_CFA_GNU_args_size			This instruction specifies the total of the size of the arguments which have been pushed onto the stack.
DW_CFA_GNU_negative_offset_extended	PowerPC-specific. Deprecated. Read and skip.
```

- Fixrd failing test
- Renames `FrameContext` to `Context` to avoid stutter.

### DIFF
--- a/internal/dwarf/frame/table.go
+++ b/internal/dwarf/frame/table.go
@@ -9,6 +9,9 @@ import (
 	"github.com/parca-dev/parca-agent/internal/dwarf/util"
 )
 
+// TODO(kakkoyun): Only export functions that are used outside of this package.
+// TODO(kakkoyun): Remove print statements.
+
 // DWRule wrapper of rule defined for register values.
 type DWRule struct {
 	Rule Rule
@@ -26,11 +29,11 @@ type DWRule struct {
 // DWARF opcodes.
 //
 // Notes:
-// - probs many of these fields should be (re)moved.
-// - in another step we would remove redundant entries, e.g.:
-// 		0000000000401121 rsp+16   c-16  c-8
-// 		0000000000401124 rbp+16   c-16  c-8
-// 	 by grouping them together.
+//   - probs many of these fields should be (re)moved.
+//   - in another step we would remove redundant entries, e.g.:
+//     0000000000401121 rsp+16   c-16  c-8
+//     0000000000401124 rbp+16   c-16  c-8
+//     by grouping them together.
 type InstructionContext struct {
 	loc     uint64
 	address uint64
@@ -44,14 +47,15 @@ type InstructionContext struct {
 	RetAddrReg    uint64
 	codeAlignment uint64
 	dataAlignment int64
+	ArgsSize      uint64
 }
 
 func (instructionContext *InstructionContext) Loc() uint64 {
 	return instructionContext.loc
 }
 
-// Represents a whole native function.
-type FrameContext struct {
+// Context represents a whole native function.
+type Context struct {
 	// An entry for each machine code instruction (not DWARF instructions)
 	instructions []InstructionContext
 	// The buffer where we store the .eh_frame entries to be parsed for this function.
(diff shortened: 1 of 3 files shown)


internal/dwarf/frame/table.go
```diff
@@ -531,135 +537,150 @@ func restoreextended(frameContext *FrameContext) {
...
+	ctx.buf.Next(1)
 }
 
-func hiuser(frameContext *FrameContext) {
-	frameContext.buf.Next(1)
+func hiuser(ctx *Context) {
+	ctx.buf.Next(1)
 }
 
-func gnuargsize(frameContext *FrameContext) {
```
kakkoyun, line 659: cc @javierhonduco

internal/dwarf/frame/table.go
```diff
@@ -531,135 +537,150 @@ func restoreextended(frameContext *FrameContext) {
...
 
+func gnuargsize(ctx *Context) {
 	// The DW_CFA_GNU_args_size instruction takes an unsigned LEB128 operand representing an argument size.
-	// Just read and do nothing.
-	// TODO(kakkoyun): Implement this.
-	_, _ = util.DecodeSLEB128(frameContext.buf)
+	// This instruction specifies the total of the size of the arguments which have been pushed onto the stack.
+	frame := ctx.getCurrentInstruction()
+	size, _ := util.DecodeULEB128(ctx.buf)
+	frame.ArgsSize = size
```
kakkoyun, line 676: This doesn't change anything for stack unwinding. I decided to store it just in case if we want to push changes to the upstream.

internal/dwarf/frame/table.go
javierhonduco: This is what confused me, thanks! I see the implementation doesn't change for this opcode, as we were already reading the unsigned LEB128, which makes sense 😄

internal/dwarf/frame/table.go
```diff
@@ -531,135 +537,150 @@ func restoreextended(frameContext *FrameContext) {
...
+}
+
+// PowerPC-specific. Deprecated.
```
javierhonduco, line 679: PowerPC is not an architecture we intend on supporting. Should we not add this opcode?

internal/dwarf/frame/table.go
```diff
@@ -44,65 +47,64 @@ type InstructionContext struct {
...
 }
 
 func (instructionContext *InstructionContext) Loc() uint64 {
 	return instructionContext.loc
 }
 
-// Represents a whole native function.
-type FrameContext struct {
+// Context represents a whole native function.
+type Context struct {
```
javierhonduco, line 58: nit: I find the new name a bit less descriptive

javierhonduco (approved):
LGTM. It should be alright but will run some tests once this lands. Will add some docs on how to test it for forthcoming PRs!

Focusing on the added instructions was a bit hard, would you mind splitting the name changes and additions in different commits next time? 😃

internal/dwarf/frame/table.go
kakkoyun: I understand that/ We need to think it in the context. It is an idiomatic Go thing; the name of the package is already `frame`.

kakkoyun:
> LGTM. It should be alright but will run some tests once this lands. Will add some docs on how to test it for forthcoming PRs!
> 
> Focusing on the added instructions was a bit hard, would you mind splitting the name changes and additions in different commits next time? smiley

Sorry for this. I'll at least separate these type of changes into a different commit.

internal/dwarf/frame/table.go
kakkoyun: For the sake of completeness of the specs, I'd prefer to have these in place with no-ops. This is what I observed from other implementation as well.

javierhonduco:
> Sorry for this. I'll at least separate these type of changes into a different commit.

No worries, feel free to land this one as-is though 😄

PR #794, opened 2022-09-07
.eh_frame: unwind table map improvements

javierhonduco:
- increase table size by 30k items
- read symbols from the dynamic string table, too
- add initial code to compact the unwind table by removing redundant rows (disabled by default, will add tests in the near future, but tested with `parca-demo-cpp-no-fp` and `parca-demo-cpp` and they both worked)

### DIFF
--- a/pkg/profiler/cpu/cpu.go
+++ b/pkg/profiler/cpu/cpu.go
@@ -397,12 +397,14 @@ func (p *CPU) ensureUnwindTables(pid int) error {
 			return fmt.Errorf("failed to open elf: %w", err)
 		}
 
-		syms, err := e.Symbols()
+		syms, symsErr := e.Symbols()
+		dynSyms, dynSymsErr := e.DynamicSymbols()
 
-		if err != nil {
+		if symsErr != nil && dynSymsErr != nil {
 			return fmt.Errorf("failed to read symbols: %w", err)
 
 		}
+		syms = append(syms, dynSyms...)
 		fmt.Println("symbol count", len(syms))
 
 		for _, sym := range syms {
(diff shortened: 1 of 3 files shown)


pkg/profiler/cpu/cpu.go
```diff
@@ -415,6 +417,25 @@ func (p *CPU) ensureUnwindTables(pid int) error {
 			}
 		}
 
+		if compact {
```
kakkoyun, line 420: :+1:

pkg/profiler/cpu/cpu.go
```diff
@@ -415,6 +417,25 @@ func (p *CPU) ensureUnwindTables(pid int) error {
...
+			compact := unwind.PlanTable{}
+			prevCFARegister := uint64(0)
+			prevCFAOffset := int64(0)
+			prevRBPRegisterOffset := int64(0)
+
+			for _, row := range pt {
+				newEntry := row.CFA.Reg != prevCFARegister || row.CFA.Offset != prevCFAOffset || row.RBP.Offset != prevRBPRegisterOffset
```
kakkoyun, line 427:
```suggestion
				notSameAsPrev := row.CFA.Reg != prevCFARegister || row.CFA.Offset != prevCFAOffset || row.RBP.Offset != prevRBPRegisterOffset
```
I think this is semantically more correct, no?

kakkoyun: This has changed after I've commented above :see_no_evil:

kakkoyun (approved):
LGTM

pkg/profiler/cpu/cpu.go
javierhonduco: Was just changing it :D Thanks for noticing, the previous name was outright wrong!

Issue #760, opened 2022-08-30
metadata: Use go-systemd instead of depending on systemctl


kakkoyun:
It is more resilient and generic rather than using `systemctl` https://github.com/coreos/go-systemd

palash25:
@kakkoyun can i take this up if no one is working on this?

maxbrunet:
Hey @palash25, thank you for your interest, I have some work ready, I'll open the PR on Monday. Sorry, I should have assigned myself

kakkoyun:
@palash25 We have other good first issues. Please let us know if we can help you find things to start 🤗  Also feel free to join our bi-weekly office hours on Thursdays.

palash25:
yeah no problem i would look for some other issues, thanks for the prompt response

PR #834, opened 2022-09-23
fix(discovery): ensure systemd units which have disappeared are accounted for

### LINKED ISSUE #806: discovery never forgets exited processes
The discovery seems to never forgets a process even after it has exited. This is at least true for systemd. This could result in a slight memory leak on high churn nodes.

We want to keep the labels available until the current profiling loop finishes, beyond that there should be a clean up (mark for deletion with timestamp?).

The label cache keys also never expires (maybe use the [`WithExpireAfterWrite`](https://pkg.go.dev/github.com/goburrow/cache#WithExpireAfterWrite) option with 2x profiling interval):

https://github.com/parca-dev/parca-agent/blob/78837cd2d067b25c6b76c8e7dfc1b564784b6f98/pkg/discovery/discovery_manager.go#L129

maxbrunet:
This adds empty `TargetGroup`s (with only `Source` set) to notify the discovery manager when systemd units have exited. Kubernetes has a delete channel which allow do that nicely, but systemd would just leave them dangling until now. It may be possible to use DBus signals instead when doing #760

It also corrects the PID labels cache configuration (I had initially not seen it was configured with an option, and thought it was not configured at all):

* Expire after write as some keys may never be accessed
* Set expiration in relation to profiling duration as the cache is accessed in the profile loop, not during remote batch write

Fixes #806

### DIFF
--- a/cmd/parca-agent/main.go
+++ b/cmd/parca-agent/main.go
@@ -230,7 +230,7 @@ func run(logger log.Logger, reg *prometheus.Registry, flags flags) error {
 		}
 		m = discovery.NewManager(logger, reg,
 			discovery.WithProcessLabelCache(cache.New(
-				cache.WithExpireAfterAccess(flags.RemoteStoreBatchWriteInterval*2),
+				cache.WithExpireAfterWrite(flags.ProfilingDuration*2),
 			)),
 		)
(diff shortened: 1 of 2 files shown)


kakkoyun (approved):
LGTM 🚀

Issue #203, opened 2022-01-10
grpc: received message larger than max


kakkoyun:
Opening this in here but this could be solved on the agent side as well. This is probably an overlooked side effect of Write Request batching on the agent side. https://github.com/parca-dev/parca-agent/pull/116 We probably need to communicate the max message size to the agent (with a flag or an exposed discovery API?)

I see another message now, a lot of them.
``` 
level=error ts=2022-01-10T13:20:04.693514836Z caller=write_client.go:83 msg="Writeclient failed to send profiles" err="rpc error: code = ResourceExhausted desc = grpc: received message larger than max (174243321 vs. 4194304)"

```

_Originally posted by @korjavin in https://github.com/parca-dev/parca/issues/514#issuecomment-1008880961_

cc @Sylfrena @brancz

brancz:
Ultimately though, I think we should stream the data, but that we should only do once we truly implement our own protocol/format.

PR #836, opened 2022-09-24
pkg/agent: trigger batch before grpc message size overflow

heylongdacoder:
Fixes: #203

Signed-off-by: heylongdacoder <heylongdacoder@gmail.com>


heylongdacoder:
Hello, I don't have much experience on writing concurrent code, might make some mistake. Thanks for guidance in advance. :D

pkg/agent/batch_remote_write_client.go
```diff
@@ -32,21 +34,31 @@ type BatchWriteClient struct {
 	writeClient   profilestorepb.ProfileStoreServiceClient
 	writeInterval time.Duration
 
-	mtx    *sync.RWMutex
-	series []*profilestorepb.RawProfileSeries
+	triggerBatch chan struct{}
+	batchDone    chan struct{}
```
kakkoyun, line 38: I think we can get rid of `batchDone` in here.

pkg/agent/batch_remote_write_client.go
```diff
@@ -67,45 +79,48 @@ func (b *BatchWriteClient) Run(ctx context.Context) error {
...
 		b.report(time.Now(), b.batch(ctx))
+
+		select {
+		case b.batchDone <- struct{}{}:
+		default:
+		}
 	}
 }
 
 func (b *BatchWriteClient) batch(ctx context.Context) error {
```
kakkoyun, line 94:
```suggestion
func batch(ctx context.Contextm, r *profilestorepb.WriteRawRequest) error {
```

This method can receive the request and by this way we can get rid of the lock in this method if I'm not mistaken. This method could also omit the write client as a receiver if it's the case.

kakkoyun (changes requested):
Looking great!

I think we can simplify this code and make it more maintainable. We can get rid of one of the mutex locking/unlocking and remove one of the channels if we can make `batch` method to receive a request (probably we should rename it to just `send` after this change). With this flow will be easier and less error prone. 

What do you think?

heylongdacoder:
@kakkoyun thanks for the review and guidance! yes I need some help. I am not sure how would it works if we removed the `batchDone` channel. If we removed `batchDone`, I think there is a chance where `WriteRaw` function will continue to append data to the overflowed `writeRawRequest`(https://github.com/parca-dev/parca-agent/pull/836/files#diff-7276397f96515c11ca49f4f72de2237d63936ab985888a5a4fe07ccf8c371d42R166) before the `batch` function could send it. 

Currently I got another idea to avoid the creation of these `triggerBatch` and `batchDone` channels. Instead of a `writeRawRequest`, what about we create a slices of `writeRawRequest` and a `writeRawRequestIndex`? `WriteRaw` function will write to `writeRawRequest[writeRawRequestIndex]` and when the current one is going to be overflowed, then we write to `writeRawRequest[writeRawRequestIndex+1]`. Then the `batch` function either loop through the slices of `writeRawRequest` to fire gRPC request sequentially or create goroutine for each element to fire gRPC request. Do you think this would work and is this a better approach?

brancz:
Outdated since #2958 

We now use a totally new protocol that is apache-arrow based.

PR #858, opened 2022-10-03
Add system metadata

javierhonduco:
This PR adds the kernel release and the Parca Agent revision. The former is very useful to drill down in performance investigations when the only difference between two sets of machines is the kernel version. The latter will help us spot differences in correctness when we release new changes to the Agent, as we can now compare the before and after.

In both cases having this data in the profiles themselves is great to help debug issues with the Agent.

## Test Plan

image


pkg/metadata/system.go
```diff
@@ -0,0 +1,73 @@
...
+
+	return int8SliceToString(uname.Release[:]), nil
+}
+
+// System provides metadata for the current system.
+func System() *Provider {
+	return &Provider{"system", func(_ int) (model.LabelSet, error) {
+		return model.LabelSet{
+			"kernel_release":     model.LabelValue(release),
+			"agent_vcs_revision": model.LabelValue(revision),
```
maxbrunet, line 54: I wonder if `_version` would be more expected, especially `agent_version`

maxbrunet (commented):
LGTM. Just some minor key/value comments

And needs formatting as spotted by pre-commit/Go build

pkg/metadata/system.go
```diff
@@ -0,0 +1,73 @@
...
+func init() {
+	r, err := kernelRelease()
+	if err != nil {
+		release = "<failed to get kernel release>"
+	}
+	release = r
+
+	b, err := buildinfo.FetchBuildInfo()
+	if err != nil {
+		revision = "<failed to get build revision>"
```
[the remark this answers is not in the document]
javierhonduco: Good idea!

pkg/metadata/system.go
javierhonduco: Perhaps `agent_revision`, so it's clear it's a VCS revision rather than semver?

javierhonduco:
Addressed PR feedback, fixed linter, and creating the whole LabelSet just once

javierhonduco:
More lint fixes

javierhonduco:
Definitely, my dev env is not correct... I don't see these lints locally. Will investigate why

pkg/metadata/system.go
```diff
@@ -0,0 +1,77 @@
...
+	return &Provider{"system", func(_ int) (model.LabelSet, error) {
+		return labels, nil
+	}}
+}
+
+// Call the system metadata getters just once as they will not
+// change while the Agent is running.
+//
+//nolint:gochecknoinits
+func init() {
```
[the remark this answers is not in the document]
javierhonduco: Ohh, TIL, sure! It would take a few more cycles but not a biggie. We can always roll back if needed, but I don't anticipate this will even show up in our profiles 😛

pkg/metadata/system.go
```diff
@@ -0,0 +1,77 @@
...
+var labels model.LabelSet
+
+func int8SliceToString(arr []int8) string {
+	b := make([]byte, 0, len(arr))
+	for _, v := range arr {
+		// NUL byte, as it's a C string.
+		if v == 0 {
+			break
+		}
+		b = append(b, byte(v))
```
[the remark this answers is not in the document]
javierhonduco: Thanks for showing me the way to Go!

javierhonduco:
Addressed feedback + re-ran test plan

pkg/metadata/system.go
```diff
@@ -0,0 +1,79 @@
...
+	if err := syscall.Uname(&uname); err != nil {
+		return "", err
+	}
+
+	return int8SliceToString(uname.Release[:]), nil
+}
+
+// System provides metadata for the current system.
+func System() *Provider {
+	var once sync.Once
```
kakkoyun, line 52: I think this is unnecessary now that I have re-read the code. `System` only gets called once,j anyway.

javierhonduco: While this is true, and it will only be called once, it's because we only have one profiler right now. By memoizing the value here we ensure the value is never recomputed. I am thinking for example once we add multiple profilers or if the way we call the metadata providers and we start calling `System()` more than once

What do you think?

kakkoyun: If it's the intended design, then we need to move the sync.Once to package level.

kakkoyun (approved):

PR #875, opened 2022-10-06
main: Do not exit if feature detection fails

javierhonduco:


kakkoyun (commented):
LGTM.

We need to introduce more sound start-up checks for eBPF support. We'll handle it before the release.

PR #876, opened 2022-10-06
@nocommit WIP pkg/metadata, pkg/buildid perf improvements

javierhonduco:
WIP:

TL;DR: 40% of the Parca Agent CPU time is spent reading all the elf symbols for each binary to see whether they are stripped or not!

image

(8% in memory allocations, 22% reading data)

The changes will be split in two next week:
- caching
- and (maybe) passing elf handles rather than paths


javierhonduco:
Closing as https://github.com/parca-dev/parca-agent/pull/902 contains the most important bit, the caching. Might add the refactor in another PR

Issue #690, opened 2022-08-08
Add ability to specify targets to profile


kakkoyun:
~~This issue is about bringing back the functionality specified by ` --pod-label-selector` and `--systemd-units` flags we had from the last major release.~~

```
      --pod-label-selector=STRING
                                  Label selector to control which Kubernetes
                                  Pods to select.
      --cgroups=CGROUPS,...       Cgroups to profile on this node.
      --systemd-units=SYSTEMD-UNITS,...
                                  [deprecated, use --cgroups instead] systemd
                                  units to profile on this node.
```

It's removed as part of #627, and the fully featured filtering will be added eventually.

# Tasks

- [x] As an initial implementation, let profiler to capture profiles but just do not send them to the server
- [x] Specify pods to profile
- [x] Specify systemd units to profile
- [x] Specify any target to profile
- [ ] Reject the events in eBPF level if the PID is not as part of a desired target

Related #7

https://github.com/parca-dev/parca-agent/pull/887 covers almost all of the intended behavior, so we don't need half-measures for now. We need to update the documentation and provide samples of configuration to demonstrate the functionality is equivalent to the flags.

kakkoyun:
Closing for the sake of https://github.com/parca-dev/parca-agent/pull/887

PR #887, opened 2022-10-11
feat: support relabeling

### LINKED ISSUE #7: Customizing labeling
We currently apply certain non-customizable labeling of data, but that may not be identical to how users label their other observability data. We should allow customizing target labels, in the Prometheus ecosystem, this is done via [relabel configs](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config), however, often relabeling is confusing especially to new-comers, so unless we conclude we need all of the functionality that relabeling provides, I'd be ok with investigating alternative paths as well.

maxbrunet:
This adds supports for `relabel_configs: []` via an optional configuration file.

There are some awkward back-and-forths between `model.LabelSet` and `labels.Labels` types, I tried to minimize them as much as possible.

On the long-run, how and where labels/metadata are managed will need to be rethought.

I had to re-arrange the status page once again to ensure relabeling works with the full set of labels for each profile types.

Following this PR, I have in mind:

* add back k8s/cgroup/systemd flags for best-effort backward compatibility and keep them for 1 or 2 minor releases (#690), this only require appending a relabel config for each case
* update the agent documentation
* reload configuration on-change

Closes #7 and #690

### DIFF
--- a/cmd/parca-agent/main.go
+++ b/cmd/parca-agent/main.go
@@ -25,7 +25,6 @@ import (
 	"net/url"
 	"os"
 	runtimepprof "runtime/pprof"
-	"strconv"
 	"strings"
 	"time"
 
@@ -42,6 +41,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/collectors"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/labels"
 	"github.com/prometheus/prometheus/promql/parser"
 	"google.golang.org/grpc"
@@ -50,6 +50,7 @@ import (
 
 	"github.com/parca-dev/parca-agent/pkg/agent"
 	"github.com/parca-dev/parca-agent/pkg/buildinfo"
+	"github.com/parca-dev/parca-agent/pkg/config"
 	"github.com/parca-dev/parca-agent/pkg/debuginfo"
 	"github.com/parca-dev/parca-agent/pkg/discovery"
 	"github.com/parca-dev/parca-agent/pkg/kconfig"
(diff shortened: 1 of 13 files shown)


kakkoyun (commented):
Wow, this was fast 🎉Amazing work :) I just skimmed it; I'll have another look at it soon.

pkg/config/config_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+func TestLoad(t *testing.T) {
+	t.Parallel()
+
+	_, err := config.Load(`relabel_configs:
+- source_labels: [systemd_unit]
+  regex: ^$
+  action: drop
+`)
+	require.Nil(t, err)
```
brancz, line 32: Nit: I think typically in this case `require.NoError` should be used.

kakkoyun:
> * add back k8s/cgroup/systemd flags for best-effort backward compatibility and keep them for 1 or 2 minor releases ([Add ability to specify targets to profile #690](https://github.com/parca-dev/parca-agent/issues/690)), this only require appending a relabel config for each case

I favor getting rid of the flags since we have the config file now. The system-wide profiling changes a lot of things anyway. We still have this liberty (still v0.x.x); we can break backward compatibility for good reasons. We provide similar functionality with different UX. 

Also we should update jsonnet files for the new flag.

cmd/parca-agent/main.go
```diff
@@ -106,6 +109,8 @@ type Profiler interface {
 	Name() string
 	Run(ctx context.Context) error
 
+	Labels(pid profiler.PID) model.LabelSet
+	Relabel(model.LabelSet) labels.Labels
```
kakkoyun, line 113: Actually, we don't need to expose the `Relabel` method at all. We can handle all of these inside the `Label` method. We should comment that we apply relabeling method inside this function and if it returns nil, it means the process needs to be dropped.

This would also prevent unnecessary conversion between label set and labels.

kakkoyun (changes requested):
Awesome work! ❤️ I have a suggestion for simplifying things.

cmd/parca-agent/main.go
maxbrunet: This is done this way because `Labels()` and `Relabel()` are handle subtlety differently between the profiler and the UI to avoid unnecessary conversions, imo it is the most efficient as it is.

I plan to do some refactoring when I get to caching and reloading, hopefully it will be cleaner

cmd/parca-agent/main.go
```diff
@@ -133,6 +138,15 @@ func main() {
 }
 
 func run(logger log.Logger, reg *prometheus.Registry, flags flags) error {
+	cfg := &config.Config{}
+	if _, err := os.Stat(flags.ConfigPath); !errors.Is(err, os.ErrNotExist) {
```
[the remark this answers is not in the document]
maxbrunet: Not sure what you mean, but I realized that a `Stat` was not necessary here and handling the error from `config.LoadFile` is enough

kakkoyun (approved):

PR #892, opened 2022-10-11
kconfig: Mount required directories for kubernetes config

### LINKED ISSUE #758: Current `main` doesn't start on Scaleway
It would be great to be able to run the latest version of Parca Agent on our [demo.parca.dev](https://demo.parca.dev) instance. 

Right now the agent doesn't start on Scaleway Kubernetes nodes logging the following:
level=warn name=parca-agent ts=2022-08-29T12:22:59.162801832Z caller=main.go:139 msg="failed to determine if eBPF is supported" err="kernel config not found"
level=error name=parca-agent ts=2022-08-29T12:22:59.162868702Z caller=main.go:121 err="host kernel does not support eBPF"
```

The pod is running image `ghcr.io/parca-dev/parca-agent:main-5920bd0f` with arguments:
```
      /bin/parca-agent
      --log-level=info
      --node=$(NODE_NAME)
      --remote-store-address=parca.parca.svc.cluster.local:7070
      --remote-store-insecure
      --remote-store-insecure-skip-verify
```

When `kubectl exec` into a pod (Parca pod since the agent has no shell inside) it seems that there is no `/boot` directory available, nor any `/proc/config*` files.

### PR DESCRIPTION
**Pull request by kakkoyun**
Add checks for the required directories

Fixes #758

## TODO

- [x] Test

### DIFF
--- a/cmd/parca-agent/main.go
+++ b/cmd/parca-agent/main.go
@@ -144,13 +144,12 @@ func run(logger log.Logger, reg *prometheus.Registry, flags flags) error {
 		)
 	}
 
-	bpfEnabled, err := kconfig.IsBPFEnabled()
-	if err == nil {
-		if !bpfEnabled {
-			level.Warn(logger).Log("msg", "host kernel might not support eBPF")
-		}
+	if err := kconfig.CheckBPFEnabled(); err != nil {
+		// TODO(kakkoyun): Add a more definitive test for the cases kconfig fails.
+		// - https://github.com/libbpf/libbpf/blob/1714037104da56308ddb539ae0a362a9936121ff/src/libbpf.c#L4396-L4429
+		level.Warn(logger).Log("msg", "failed to determine if eBPF is supported, host kernel might not support eBPF", "err", err)
 	} else {
-		level.Warn(logger).Log("msg", "failed to determine if eBPF is supported", "err", err)
+		level.Info(logger).Log("msg", "eBPF is supported and enabled by the host kernel")
 	}
 
 	// Fetch build info such as the git revision we are based off
--- a/deploy/lib/parca-agent/parca-agent.libsonnet
+++ b/deploy/lib/parca-agent/parca-agent.libsonnet
@@ -241,6 +241,16 @@ function(params) {
           mountPath: '/run',
         },
         {
+          name: 'boot',
+          mountPath: '/boot',
+          readOnly: true,
+        },
+        {
+          name: 'proc',
+          mountPath: '/proc',
+          readOnly: true,
+        },
+        {
           name: 'modules',
           mountPath: '/lib/modules',
         },
@@ -303,12 +313,28 @@ function(params) {
                 name: 'tmp',
                 emptyDir: {},
               },
+              // Needed for reading the container runtime metadata.
               {
                 name: 'run',
                 hostPath: {
                   path: '/run',
                 },
               },
+              // Needed for reading kernel configuration.
+              {
+                name: 'boot',
+                hostPath: {
+                  path: '/boot',
+                },
+              },
+              // Needed for reading process information.
+              {
+                name: 'proc',
+                hostPath: {
+                  path: '/proc',
+                },
+              },
+              // Deprecated by v0.10.0 release. Remove in a couple of releases.
               {
                 name: 'cgroup',
                 hostPath: {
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: cmd/parca-agent/main.go [hunk 1]
```diff
@@ -144,13 +144,12 @@ func run(logger log.Logger, reg *prometheus.Registry, flags flags) error {
...
-		if !bpfEnabled {
-			level.Warn(logger).Log("msg", "host kernel might not support eBPF")
-		}
+	if err := kconfig.CheckBPFEnabled(); err != nil {
+		// TODO(kakkoyun): Add a more definitive test for the cases kconfig fails.
+		// - https://github.com/libbpf/libbpf/blob/1714037104da56308ddb539ae0a362a9936121ff/src/libbpf.c#L4396-L4429
+		level.Warn(logger).Log("msg", "failed to determine if eBPF is supported, host kernel might not support eBPF", "err", err)
 	} else {
-		level.Warn(logger).Log("msg", "failed to determine if eBPF is supported", "err", err)
+		level.Info(logger).Log("msg", "eBPF is support and enabled by the host Kernel")
```
v-thakkar, line 152: Nit: s/support/supported

v-thakkar (commented):
LGTM :tada:

deploy/lib/parca-agent/parca-agent.libsonnet
```diff
@@ -240,6 +240,14 @@ function(params) {
           name: 'run',
           mountPath: '/run',
         },
+        {
+          name: 'boot',
+          mountPath: '/boot',
+        },
+        {
+          name: 'proc',
+          mountPath: '/proc',
```
maxbrunet, line 249: It might not be a good idea to override the container's `/proc`, we may need to use something like `/host/proc`

deploy/lib/parca-agent/parca-agent.libsonnet
```diff
@@ -303,12 +311,28 @@ function(params) {
...
                 name: 'run',
                 hostPath: {
                   path: '/run',
                 },
               },
+              // Needed for reading kernel configuration.
+              {
+                name: 'boot',
+                hostPath: {
+                  path: '/boot',
```
maxbrunet, line 325: It slightly bothers me that Kubernetes does not have way to have optional "if exists" host volumes. Here, if `/boot` does not exist on the node, it will be created which can be considered as left-over garbage if the agent is uninstalled from the cluster, so it may not be a questionable default

maxbrunet (commented):
We should use `readOnly: true` on volume mounts that do not require write permissions

deploy/lib/parca-agent/parca-agent.libsonnet
maxbrunet: And I did a test on my local machine, `/proc/config.gz` is available in the containers and the options look good (even without `hostPID: true`)

kakkoyun: Which configuration exactly have you tested?

maxbrunet: I checked in a K3d cluster backed by Podman in rootless mode and in Minikube:

```console
$ kubectl run --attach --rm --image=docker.io/busybox --restart=Never "busybox-${USER}" -- \
    gunzip --stdout /proc/config.gz \
    | grep -E '^(CONFIG_(BPF(|_SYSCALL|_JIT(|_ALWAYS_ON)|_EVENTS)|HAVE_EBPF_JIT|ARCH_WANT_DEFAULT_BPF_JIT))='
CONFIG_BPF=y
CONFIG_HAVE_EBPF_JIT=y
CONFIG_ARCH_WANT_DEFAULT_BPF_JIT=y
CONFIG_BPF_SYSCALL=y
CONFIG_BPF_JIT=y
CONFIG_BPF_EVENTS=y
```

The Scaleway cluster does not seem to have the file.

deploy/lib/parca-agent/parca-agent.libsonnet
maxbrunet: No, you're right, not a big issue

deploy/lib/parca-agent/parca-agent.libsonnet
maxbrunet: I also found a new setting that would be more appropriate, `procMount: Unmasked` in the container `securityContext`:
https://snyk.io/blog/10-kubernetes-security-context-settings-you-should-understand/#procmount

(I have not found any good official documentation about it yet though)

maxbrunet: For `procMount`? No, it does not seem necessary at the moment, and its alpha state makes it inconvenient

Found its design proposal, linking it here for easy reference: https://github.com/kubernetes/design-proposals-archive/blob/main/auth/proc-mount-type.md

PR #902, opened 2022-10-13
pkg/metadata: Cache compiler metadata output

javierhonduco:
**TL;DR:** >40% of the Parca Agent CPU cycles are spent reading all the elf symbols for each binary to see whether they are stripped or not. Some extra cycles come from added pressure on the garbage collector

The compiler metadata, specifically, the stripped check, consumes a large amount of CPU as it parses and traverses the whole symbol table for each process running on the system.

This first commit caches the compiler metadata using the BuildID as the caching key.

I noticed that while checking CPU profiles in our demo environment:

image

## Test Plan

I Ran Parca and Parca Agent side by side, and everything looked good, the compiler metadata uses way fewer cycles

image

## Future work

The `string => elf.File` refactor in https://github.com/parca-dev/parca-agent/pull/876 was mostly to improve the API, performance-wise the impact of that is not very big. Will tackle this in another PR!

### DIFF
--- a/pkg/metadata/compiler.go
+++ b/pkg/metadata/compiler.go
@@ -17,14 +17,29 @@ package metadata
 import (
 	"debug/elf"
 	"fmt"
+	"sync"
 
+	burrow "github.com/goburrow/cache"
 	"github.com/keybase/go-ps"
 	"github.com/prometheus/common/model"
 	"github.com/xyproto/ainur"
+
+	"github.com/parca-dev/parca-agent/pkg/buildid"
 )
+
+var (
+	c            burrow.Cache
+	onceCompiler sync.Once
+)
+
+func initialiseCache() {
+	c = burrow.New(burrow.WithMaximumSize(128))
+}
 
 // Compiler provides metadata for determined compiler.
 func Compiler() *Provider {
+	onceCompiler.Do(initialiseCache)
+
 	return &Provider{"compiler", func(pid int) (model.LabelSet, error) {
 		process, err := ps.FindProcess(pid)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/metadata/compiler.go
```diff
@@ -38,17 +53,34 @@ func Compiler() *Provider {
...
+		buildID, err := buildid.BuildID(path)
+		if err != nil {
+			return nil, fmt.Errorf("buildID failed")
+		}
+
+		value, ok := c.GetIfPresent(buildID)
+		if ok {
+			cachedLabels, ok := value.(model.LabelSet)
+			if !ok {
+				panic("The buildID cache contained the wrong type. This should never happen")
```
kakkoyun, line 71: Let's not panic. It would be better to return an error and delete the wrongly cached items.

kakkoyun (commented):
LGTM. Awesome results 🎉 Just one comment, then I'm happy to merge.

pkg/metadata/compiler.go
javierhonduco: For conditions that should absolutely never happen, I like panicking as it can ensure that both readers of the code understand that this is a "hard" assertion, and also ensure that if something changes in the code, later on, we have more chances of catching it. 

What do you think?

v-thakkar: I agree with Javier. I think it's good to panic here instead of returning an error.

v-thakkar:
LGTM :tada:

brancz (approved):

pkg/metadata/compiler.go
kakkoyun: Let's agree to disagree how we approach to panics :) It's fine for here.

PR #918, opened 2022-10-17
Adjust snap package to use classic confinement

jnsgruk:
After some thought, I think the way `parca-agent` accesses the underlying system makes it a more compelling candidate for `classic` confinement, rather than `strict`.

:warning:  This PR shouldn't be merged until the store have enabled classic confinement for this snap - I'll post back when that's done and take care of that getting sorted. :warning:

### DIFF
--- a/.github/workflows/snap.yml
+++ b/.github/workflows/snap.yml
@@ -188,14 +188,7 @@ jobs:
 
       - name: Install snap & invoke Parca Agent
         run: |
-          # TODO(jnsgruk): Undo this once a new snapd is released
-          sudo snap refresh snapd --channel=latest/edge
-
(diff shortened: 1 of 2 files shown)


javierhonduco (approved):
LGTM! I assume these changes are mostly caused or exacerbated by the system-wide changes, no?

jnsgruk:
That's right, there is a little more info here: https://forum.snapcraft.io/t/classic-confinement-for-parca-agent/32076

I'll come back and comment once we're cleared to release a `classic` snap so we don't end up with red CI :)

javierhonduco:
Oops, I just realised that I was meant to wait until the classic snap release. My bad 😓

PR #744, opened 2022-08-24
[WIP] *: .eh_frame-based stack unwinding (using eBPF) 

kakkoyun:
Work In Progress: Very early stages

Supersedes #346 #600
Fixes #293 

Implementation based on https://hal.inria.fr/hal-02297690/document

# TODO

- [ ] Remove Rust toolchain (for now)
- [ ] Remove vmlinux.h
- [x] Change examples with leaner binaries
- [ ] Remove large binaries (make sure they don't appear in the history

PR #922, opened 2022-10-18
.eh_frame: Bubble up DWARF expressions and value registers

javierhonduco:
And print them accordingly. This implementation mostly comes from this commit: https://github.com/parca-dev/parca-agent/pull/744/commits/e1ac8c62bc6a2715a170b095ae5e198f0e104a74, but also cleans up the code to make it less confusing and adds some assertions in the form of panics to ensure that things are working the way we expect. We might change this in the future once this code is more battle-tested.

The tables have been regenerated and tested in https://github.com/parca-dev/testdata/commit/483300c38df358af9aca8938b648953723c3f75c


v-thakkar:
LGTM :tada:

PR #923, opened 2022-10-18
ci: fix curl failing

Sylfrena:


javierhonduco:
Seems like this could be some flakiness as [sometimes it works](https://github.com/parca-dev/parca-agent/actions/runs/3272243509/jobs/5383860736) and sometimes [it doesn't](https://github.com/parca-dev/parca-agent/actions/runs/3266846445/jobs/5379223738). As a stop-gap, perhaps we could add some retries? A like `curl` has support for this natively with the `--retry*` options 😄

e2e/ci-e2e.sh
```diff
@@ -58,7 +58,7 @@ function minikube_down() {
 # Configure clusters to run latest commit in Parca agent
 function deploy() {
     echo "fetching parca binary"
-    SERVER_LATEST_VERSION=$(curl -sSf https://api.github.com/repos/parca-dev/parca/releases/latest | jq -r .tag_name)
+    SERVER_LATEST_VERSION=$(curl -s https://api.github.com/repos/parca-dev/parca/releases/latest | jq -r .tag_name)
```
maxbrunet, line 61: Ignoring and hiding errors :see_no_evil:
Hosted runners are in a shared environment and likely share the same public IP addresses. They must exhausted their anonymous requests quota easily, authenticating with `secrets.GITHUB_TOKEN` could fix that

But what about using `git ls-remote` instead of the API? Git operations have no rate limit that I know of

```
git -c 'versionsort.suffix=-' ls-remote --tags --refs --sort='v:refname' https://github.com/parca-dev/parca.git 'v*.*.*' | tail -1 | cut -d/ -f3
```

Detailed explanation here: https://stackoverflow.com/a/12704727

Sylfrena: Fully agreed! This was only supposed to be a temporary fix till I could verify that it was rate limits causing this.🙈 
While I think the authentication way is the more straightforward way to fix this, I absolutely love how ingenious the 'git ls-remote' approach is so will go with that, thanks so much!

javierhonduco (approved):
Looks great!!

maxbrunet (approved):
LGTM! :sparkles:

Issue #702, opened 2022-08-11
Add startup checks about bpf program and map type support


v-thakkar:
This will be particularly useful once we have a larger bpf program after the stack unwinding feature is merged. We can potentially leverage libbpfgo API for the same. https://github.com/v-thakkar/libbpfgo/blob/main/libbpfgo.go#L1877

PR #930, opened 2022-10-20
pkg/profiler/cpu: check if bpf program and map type are supported

heylongdacoder:
Fixes: #702

Signed-off-by: heylongdacoder <heylongdacoder@gmail.com>


v-thakkar (approved):
Thanks for handling it. LGTM :tada:

Issue #934, opened 2022-10-20
ci: e2e tests not working on PR from forks


maxbrunet:
`secrets` and ghcr.io are not available to forks.

https://github.com/parca-dev/parca-agent/blob/4ab2341857de53c935191775a9bf8a8214b00ba8/.github/workflows/container.yml#L187-L209

e2e tests should use a local registry instead. It should be possible to export/import the built image and pass it between steps with the upload/download actions.

kakkoyun:
There are two ways to fix it either using a local registry with https://github.com/docker/build-push-action/blob/master/docs/advanced/local-registry.md or exporting image as tar and uploading it as artifact https://stackoverflow.com/a/65369271/1113525

The downside of the first object is, it should be within the job, you can't expose services to other jobs. And there's the challenge of accessing that image from minikube. The latter approach could be easier to tackle.

maxbrunet:
> There are two ways to fix it either using a local registry with [docker/build-push-action@`master`/docs/advanced/local-registry.md](https://github.com/docker/build-push-action/blob/master/docs/advanced/local-registry.md?rgh-link-date=2022-11-02T14%3A36%3A01Z) or to export image as tar and uploading it as an artifact [stackoverflow.com/a/65369271/1113525](https://stackoverflow.com/a/65369271/1113525)

My suggestion was to combine both. With the Minikube's `none`, importing the image into Docker should be enough (I guess it is enough in any case as we could use `minikube docker-env`)

PR #960, opened 2022-10-27
feat(metadata): cache label sets

### LINKED ISSUE #890: relabelling: cache the result of relabeling functions
I think this can totally be a follow up, but relabel configs are pure functions meaning as long as relabel configs don’t change the produced labelset doesn’t change. Therefore for any given process we need to only relabel once and can from there cache the result.

_Originally posted by @brancz in https://github.com/parca-dev/parca-agent/pull/887#pullrequestreview-1137015174_

cc @maxbrunet

maxbrunet:
* 1 cache per profiler (key: `pid`, value: `labelSet`)
* lazy caching happens when profiler requests labels
* no caching happens when request comes from UI as:
  * this does not happen often
  * this would require convention from `label.Labels` to `model.LabelSet`
  * UI requests labels for all processes on the system, they are not all profiled if idled, which would unnecessarily inflate the cache
* all caches are invalidated on config reload
* cache keys for individual PIDs are invalidated on updates from service discovery (via a hook function)
* otherwise, cache keys expire if not accessed for more than 10 minutes

Closes #890


javierhonduco:
Hi! Thanks for the PR! Will take a look later today

Sylfrena (approved):

kakkoyun:
I have been checking why actually the PR is failing, but I don't have any reasonable explanations. We haven't changed anything concerning secrets. Yet the container action fails.

maxbrunet:
@kakkoyun please #934 for the CI, we'll ignore that until fixed

@javierhonduco do you still want to add a review?

Something came to my mind, I need to add a read lock when calling the hooks

kakkoyun:
@maxbrunet Thanks for the context. I was about to go nuts 😅 

I'm happy to merge this if others are okay as well.

kakkoyun (approved):

kakkoyun:
> LGTM! Will let it bake for 1 day in the [demo environment](https://demo.parca.dev/)

Have you deployed it? Or shall we merge and deploy a version?

javierhonduco:
@kakkoyun Meant to do this once we merge it :)

javierhonduco:
One small comment about something we need to revisit in the Agent as a whole is that caching PIDs can be problematic due to them being reused.

We don't have to tackle this in this PR but it's something that we should be aware of

maxbrunet:
> One small comment about something we need to revisit in the Agent as a whole is that caching PIDs can be problematic due to them being reused.

We should be able to add the `starttime` field of `/proc/[pid]/stat` to the cache key: `awk '{print $23}' "/proc/${pid}/stat"`

kakkoyun:
> > One small comment about something we need to revisit in the Agent as a whole is that caching PIDs can be problematic due to them being reused.
> 
> We should be able to add the `starttime` field of `/proc/[pid]/stat` to the cache key: `awk '{print $23}' "/proc/${pid}/stat"`
> 
> https://pkg.go.dev/github.com/prometheus/procfs#ProcStat.StartTime

It's a good idea!

Issue #961, opened 2022-10-27
Add readiness/liveness HTTP endpoints


maxbrunet:
This will allow to create probes for the Kubernetes DaemonSet to ensure rollouts are successful

Right now we hope it is running when this command returns:

https://github.com/parca-dev/parca-agent/blob/0e4ef533698b40c09d1295fb05dc1bd64ba04162/e2e/ci-e2e.sh#L73

PR #478, opened 2022-06-20
ci: Enable e2e testing in ci

Sylfrena:
Modify actions and tests to allow e2e testing in parca agent.

For every PR, a test image of parca agent is created(hosted at parca-agent-test)
which is then tested alongwith the latest release of parca server
within a minikube setup in CI.

Add support for debugging opaque failures in CI using upterm shell.

All scripts and tests can be found in the `e2e` directory in parca-agent

Co-developed-by: Frederic Branczyk
Signed-off-by: Sumera Priyadarsini <sylphrenadin@gmail.com>

PR #962, opened 2022-10-27
ci(e2e): run on Linux

maxbrunet:
Since the runner is already a Linux VM, we can use it directly instead of nested virtualization on MacOS, this cuts the duration of e2e tests by more than half (>20min :arrow_right: <10min)

Half of what is left is a 5-minute fixed wait ~~for the deployment to be stable, I opened #961 to optimize that~~ nvm, I just put things together, it is profiling the system for the duration

This also adds a Bash `trap` to output pods statuses and Parca logs, no matter if the setup was successful or not, extremely helpful for debugging

Finally, the latest Parca manifest is broken, so I hard-coded the previous release


e2e/ci-e2e.sh
```diff
@@ -75,8 +90,6 @@ function deploy() {
     echo "Connecting to Parca and Parca agent"
 
     sleep 300
-
-    kubectl get all -A
```
javierhonduco, line 79: I'm assuming this was for debugging, do we still want it @Sylfrena?

javierhonduco (approved):
> Finally, the latest Parca manifest is broken, so I hard-coded the previous release

Thanks for doing this! Do you mind open an issue to not forget to bump the version once the manifest is fixed?

LGTM! If it's ok, let's wait for @Sylfrena's review as she's the maintainer of the e2e tests 😄

e2e/ci-e2e.sh
```diff
@@ -58,11 +68,16 @@ function minikube_down() {
...
-    SERVER_LATEST_VERSION=$(git -c 'versionsort.suffix=-' ls-remote --tags --refs --sort='v:refname' https://github.com/parca-dev/parca.git 'v*.*.*' | tail -1 | cut -d/ -f3)
+    ### FIXME: Until release of 0.13.1, latest Parca manifest is broken
+    ### https://github.com/parca-dev/parca/pull/1989
+    # SERVER_LATEST_VERSION=$(git -c 'versionsort.suffix=-' ls-remote --tags --refs --sort='v:refname' https://github.com/parca-dev/parca.git 'v*.*.*' | tail -1 | cut -d/ -f3)
+    SERVER_LATEST_VERSION=v0.12.1
     echo "Server version: $SERVER_LATEST_VERSION"
 
     #AGENT_LATEST_VERSION=$(curl -sSf https://api.github.com/repos/parca-dev/parca-agent/releases/latest | jq -r .tag_name)
 
+    trap dump_deploy_status EXIT
```
Sylfrena, line 79: Absolutely love this bit, thank you!

Sylfrena:
Thank you for doing this!

> Since the runner is already a Linux VM, we can use it directly instead of nested virtualization on MacOS, this cuts the duration of e2e tests by more than half (>20min ➡️ <10min)

However, I am afraid this will not work as we need a real VM to test parca-agent- so we need to use `qemu/kvm2/virtualbox` as the minikube driver instead of `docker`. Unfortunately this is not supported for Linux runners on GitHub actions so we had to use the macOS runners instead (ref: https://github.com/parca-dev/parca-agent/pull/478#discussion_r901321686)

We probably should have documented this somewhere so it was clearer, sorry about that :/

> This also adds a Bash trap to output pods statuses and Parca logs, no matter if the setup was successful or not, extremely helpful for debugging

That said, I love this so please let's keep this change!

maxbrunet:
> However, I am afraid this will not work as we need a real VM to test parca-agent

Sorry, I missed this comment, @Sylfrena you were right back then! 

Runners are **real**, Azure, VMs. The `docker` driver simply means we run the k8s components on this host directly (theorocally, the `none` driver should work too, although trickier to get going), the containers still use the host kernel which needs to support eBPF.

The eBPF support check does not pass, but it must be a red heirring, otherwise we need to figure out why the e2e tests pass?

cc @brancz @kakkoyun

javierhonduco:
> The eBPF support check does not pass, but it must be a red heirring, otherwise we need to figure out why the e2e tests pass?


This is a bug that we need to fix, the detection was wrong (remediation in https://github.com/parca-dev/parca-agent/pull/875, further fixes https://github.com/parca-dev/parca-agent/pull/892) cc @v-thakkar 


> The eBPF support check does not pass, but it must be a red heirring, otherwise we need to figure out why the e2e tests pass?

I think it's reasonable that the tests passed with the docker runtime, did you expect otherwise?

> Runners are real, Azure, VMs. The docker driver simply means we run the k8s components on this host directly, but containerized (theorocally, the none driver should work too, although trickier to get going), the containers still use the host kernel which needs to support eBPF.

As Sumera mentioned, we would like to use a full-fledged VM with a specific kernel to ensure that all the features we add to the Agent work with it. This way we also ensure that all the Agent developers use the same Kernel version locally in our Minikube setup, both for testing and development. 

Hope this makes sense!

brancz:
> The eBPF support check does not pass, but it must be a red heirring, otherwise we need to figure out why the e2e tests pass?

I think this is something we need to double check. After all the e2e tests are there to increase our confidence, and if we can’t be sure about it then what’s the point … 

> This way we also ensure that all the Agent developers use the same Kernel version locally in our Minikube setup, both for testing and development.

I can see what you’re saying in regards to reproducibility, but at the same time, this project is there to ensure the widest possible support is there (we can also choose a specific image used to run builds which also ensures kernel versions). It may also just be a question of, are we willing to trade off the way faster builds for this?

kakkoyun:
Even though eBPF config checks seem to be failing, it looks like Agent is actually running using the docker driver on these machines. Our previous attempts failed because minikube driver was using containers as node hosts but it doesn't seem to be the case with these Linux machines. I double-checked the documentation and apparently this driver directly using the underlying machine as a host. Has it changed recently? 

All in all, if it works, it works. I'm happy to move in this direction.

javierhonduco:
> I can see what you’re saying in regards to reproducibility, but at the same time, this project is there to ensure the widest possible support is there (we can also choose a specific image used to run builds which also ensures kernel versions). It may also just be a question of, are we willing to trade off the way faster builds for this?


That's a good point. Perhaps something that would be the best of both worlds would be to:

- run the e2e test with every PR using the Docker runtime, so we can get more confidence on the changes, while not making the CI times much worse;
- have a periodic GH action that runs on `main` every X hours/days running with a full-fledged VM to ensure that the Agent runs in some specific kernel version we all agree on. That way we can make sure that we don't introduce regressions that can only be noticed in older kernels;

The downside would be to have these two setups to maintain, which might be too much work, but I think it's something that we should consider

maxbrunet:
> I can see what you’re saying in regards to reproducibility, but at the same time, this project is there to ensure the widest possible support is there (we can also choose a specific image used to run builds which also ensures kernel versions). It may also just be a question of, are we willing to trade off the way faster builds for this?

:heavy_plus_sign: 

> Our previous attempts failed because minikube driver was using containers as node hosts but it doesn't seem to be the case with these Linux machines. I double-checked the documentation and apparently this driver directly using the underlying machine as a host. Has it changed recently?

I do not think so, it is Docker-in-Docker, Minikube creates a `minikube` container which runs its own Docker daemon. Containers cannot run their own kernel, so the structure is flatten at this level. That also mean for the host's `/boot` to be mounted in the agent container, it first needs to be mounted in the node container, I just pushed a commit for that.

If Docker-in-Docker is still an issue, we can give a shot at the `none` driver, and it would be K8s running on a bare-bone VM. The technicality is some dependencies need to be installed (`conntrack`, `socat`...), but [medyagh/setup-minikube](https://github.com/medyagh/setup-minikube) has support for all that

maxbrunet:
> Do you mind open an issue to not forget to bump the version once the manifest is fixed?

Actually, we should use Renovate here and group it with updates for the Go and Jsonnet modules, that way we get a PR and it is tested before getting into `main`

Sylfrena:
> > I can see what you’re saying in regards to reproducibility, but at the same time, this project is there to ensure the widest possible support is there (we can also choose a specific image used to run builds which also ensures kernel versions). It may also just be a question of, are we willing to trade off the way faster builds for this?
> 
> ➕
> 
> > Our previous attempts failed because minikube driver was using containers as node hosts but it doesn't seem to be the case with these Linux machines. I double-checked the documentation and apparently this driver directly using the underlying machine as a host. Has it changed recently?
> 
> I do not think so, it is Docker-in-Docker, Minikube creates a `minikube` container which runs its own Docker daemon. Containers cannot run their own kernel, so the structure is flatten at this level. That also mean for the host's `/boot` to be mounted in the agent container, it first needs to be mounted in the node container, I just pushed a commit for that.
> 

If this is indeed the case, I am super happy to have been proven wrong here. 🎉💃🏻 

> If Docker-in-Docker is still an issue, we can give a shot at the `none` driver, and it would be K8s running on a bare-bone VM. The technicality is some dependencies need to be installed (`conntrack`, `socat`...), but [medyagh/setup-minikube](https://github.com/medyagh/setup-minikube) has support for all that

As far as choosing between `docker` and `none` is concerned, I'd like to try both on a local setup before confidently choosing either, but no strong preferences as of now. I do think Javier's idea to have two setups is very doable and would give us more confidence with the e2e tests.

kakkoyun:
> If Docker-in-Docker is still an issue, we can give a shot at the `none` driver, and it would be K8s running on a bare-bone VM. The technicality is some dependencies need to be installed (`conntrack`, `socat`...), but [medyagh/setup-minikube](https://github.com/medyagh/setup-minikube) has support for all that

`none` should work perfectly fine. If there's a way to set it up painlessly (from the repo, I understand there is), I'm all in for that.

kakkoyun (approved):

kakkoyun:
> Also something to note about the `docker` driver, the node container does not use the host PID namespace and it is not configurable, so the agent cannot see the runner's processes, but it can see everything running in Kubernetes, for the e2e tests I do not think it is a big deal tho.
> 
> imo we can merge this PR to fix and speed up the CI, and then try the none driver (it also requires some clean up of the script, some flags are not supported (nothing we are too attached too) like profiles or anything that would do a downgrade)

For sure. This looks good. And we should create a note for the downside of using minikube somewhere in our docs for the people that would like to use it.

I'll merge this in the day.

javierhonduco (approved):

Sylfrena (approved):

kakkoyun (changes requested):
I think this is not really working as expected.

This also shows our e2e tests fail to catch certain things. That's something we have been expecting. We'll write more accurate e2e tests.

I'll add more soon.

kakkoyun:
You can find the result of running the agent on minikube with the docker driver. It seems like it's working, but it is not entirely working. Queries somewhat work, and we receive results. That's why e2e tests are passed. But the address we don't receive the actual address. This could be an easy-to-solve issue, but we need to investigate. 
  
Query output:
CleanShot 2022-11-02 at 15 01 35@2x

Logs:
```
level=debug name=parca-agent ts=2022-11-02T13:59:45.228424281Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=280129 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.229107577Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=254951 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.229765043Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=1774 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.230218467Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=257702 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.230866994Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=257429 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.231062027Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=276316 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.23123375Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=257388 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.231492761Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=211883 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.231617797Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=218303 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.231749292Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=280100 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.231921366Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=254122 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.23208026Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=220547 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.232190646Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=258926 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.23234918Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=2011 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.232611741Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=280115 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.232871382Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=260367 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.233167961Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=279360 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.233584906Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=280099 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.234140896Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=276400 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.234652848Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=1764 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.235198968Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=1952 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.235477018Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=257381 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.235660971Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=210770 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.235769457Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=280070 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.236130225Z caller=symbol.go:73 component=symbolizer msg="failed to obtain perf map for pid" pid=279398 err="process status not found"
level=debug name=parca-agent ts=2022-11-02T13:59:45.241398765Z caller=batch_remote_write_client.go:139 msg="batch write client sent profiles" count=23
```

cc @javierhonduco @Sylfrena

maxbrunet:
Shall we merge anyway? We'll only have a half working pipeline after, and we can iterate with the `none` driver and improved tests

kakkoyun:
> Shall we merge anyway? We'll only have a half working pipeline after, and we can iterate with the `none` driver and improved tests

I would instead disable the e2e tests and make the pipeline work (and then work on functional e2e tests or fix https://github.com/parca-dev/parca-agent/issues/934). We know the agent won't work in docker-in-docker so we shouldn't have an action would suggest otherwise.

kakkoyun:
@maxbrunet shall I close this or do you want to keep it open for `node` driver changes? (I'm not suggesting, you shouldn't be the one introduce those changes, just asking for your plans)

maxbrunet:
> shall I close this or do you want to keep it open for `node` driver changes?

I'm going to keep it open for tracking. It'll take more time for the `none` driver, it does not "just work" unfortunately

kakkoyun:
Superseded by #1000

PR #963, opened 2022-10-27
feat: add simple health HTTP endpoints

### LINKED ISSUE #961: Add readiness/liveness HTTP endpoints

maxbrunet:
Add simple health endpoints just to get started

Closes #961


cmd/parca-agent/main.go
```diff
@@ -320,7 +320,7 @@ func run(logger log.Logger, reg *prometheus.Registry, flags flags) error {
 		),
 	}
 	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
-		if r.URL.Path == "/favicon.ico" {
+		if r.URL.Path == "/healthy" || r.URL.Path == "/ready" || r.URL.Path == "/favicon.ico" {
```
javierhonduco, line 323: nit: What do you think about`/status`? I think it is a more common health check path, no?

javierhonduco (approved):
LGTM!

cmd/parca-agent/main.go
maxbrunet: `ready` and `healty` is a common naming for the Kubernetes-era, Prometheus, Thanos, and Kubernetes components use it for example. Later, we can imagine more in-depth checks under `/ready` and keep things light for continuous `/healthy` checks

javierhonduco: Sounds good! Guess I am not up to date on the Cloud Native ™️  lingo!

kakkoyun (approved):
It's good for a start. We have to do something more sophisticated in the long run. We have this prober pattern we use in Parca (or in other projects), where you can make sure if the app is actually healthy my changing the state of prober. So for us it should be only healthy if we have a running profiler, otherwise it should fail.

PR #970, opened 2022-10-31
actions: Fix failing action caused by quay login

kakkoyun:

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -2,9 +2,17 @@ name: Container
 
 on:
   push:
-    branches: [ main ]
+    branches:
+    - main
+    - release-*
   pull_request:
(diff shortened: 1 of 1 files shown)


maxbrunet:
@kakkoyun if simply doing that, you would need to skip the e2e jobs too (which isn't worse than ignoring it as do now)

kakkoyun:
> @kakkoyun if simply doing that, you would need to skip the e2e jobs too (which isn't worse than ignoring it as do now)
> 
> Ref #934

Yes, the problem is explained clearly in the issue you referenced. We should use a local registry if possible and not expose these images.

One thing I realized, we haven't pushed anything to the quay for the last six months. That's another issue we need to look at.

kakkoyun:
Closing this one.

PR #971, opened 2022-10-31
refactor(discovery): use systemd D-Bus API

### LINKED ISSUE #760: metadata: Use go-systemd instead of depending on systemctl

maxbrunet:
This replaces the invocations of `systemdctl` CLI by the systemd D-Bus API.

This is still a polling approach, the module conveniently remembers the previous set of units and returns the changes only, allowing to trigger minimal discovery updates.

The D-Bus signals that would allow for this to be event-driven are [unreliable](https://github.com/coreos/go-systemd/blob/v22.4.0/dbus/subscription.go#L160-L168), and although there is [a workaround](https://github.com/coreos/go-systemd/blob/v22.4.0/dbus/subscription.go#L246-L258), I have observe many duplicated updates and easily fall into infinite loops of updates.

Using the D-Bus API permits to mount the D-Bus socket into the agent container to make it aware of the host systemd units. (note: I have also removed the `/proc` mount which is not needed, `hostPID: true` is enough)

Closes #760


javierhonduco:
Nice work, will review it later today 🙇 .

> (note: I have also removed the /proc mount which is not needed, hostPID: true is enough)

I think we still need it for the metadata providers, no?

pkg/discovery/systemd.go
```diff
@@ -50,111 +45,66 @@ func (c *SystemdConfig) NewDiscoverer(d DiscovererOptions) (Discoverer, error) {
...
-			strings.HasPrefix(line, "SUB") {
-			continue
-		}
-		fields := strings.Fields(line)
-		if len(fields) < 1 {
-			continue
-		}
-		units = append(units, fields[0])
+	isService := func(name string) bool {
+		return !strings.HasSuffix(name, ".service")
```
kakkoyun, line 59:
```suggestion
		return !strings.HasSuffix(name, ".service")
```
Shouldn't it be like this, as the name suggests?

pkg/discovery/systemd.go
```diff
@@ -50,111 +45,66 @@ func (c *SystemdConfig) NewDiscoverer(d DiscovererOptions) (Discoverer, error) {
...
-	cmd := exec.CommandContext(ctx, "systemctl", "show", "--property", "MainPID", "--value", unit)
-	output, err := cmd.Output()
-	if err != nil {
-		return nil, err
-	}
-
-	pid, err := strconv.Atoi(strings.TrimSpace(string(output)))
-	if err != nil {
-		return nil, err
+	for {
```
kakkoyun, line 64: Why do we need the outer loop? It seems to me we could fold into one.
What I am missing?

deploy/lib/parca-agent/parca-agent.libsonnet
```diff
@@ -22,6 +22,9 @@ local defaults = {
   debugInfoUploadDisable: false,
   socketPath: '',
 
+  hostDbusSystem: false,
```
kakkoyun, line 25: So it will be disabled by default.

It was the best effort previously, maybe we can do the same and enable it always and let it silently fail if can't communicate with systemd dbus. What do you think?

kakkoyun (commented):
It always awesome to have one less external dependency 👍

pkg/discovery/systemd.go
maxbrunet: I think you forgot to suggest a change :grin:

I know this works as intended, do you mean the function should be called `isNotService`?
I meant "I want the services only", but I was surprised that it needs a negative filter function

pkg/discovery/systemd.go
maxbrunet: Nothing, I hadn't noticed that, thank you for catching it!

deploy/lib/parca-agent/parca-agent.libsonnet
maxbrunet: We can give it a try, I am always worried about what happens when the source does not exist (it's probably edge cases tho like the socket does not exist, the mount happens and creates a directory, D-Bus tries to start and wants to create its socket... I guess in most cases D-Bus should have started long before or it means it is not there)

maxbrunet:
> > (note: I have also removed the /proc mount which is not needed, hostPID: true is enough)
> 
> I think we still need it for the metadata providers, no?

It was done to expose the kernel config, but after some testing, it has been proven unecessary: https://github.com/parca-dev/parca-agent/pull/892#discussion_r993866042

`hostPID: true` tells to run the pod in the host's PID namespace (same as `--pid=host` with `docker run`), so all information about the host's processes should be available to the agent. 

From what I have documented [here](https://github.com/maxbrunet/parca-dev-docs/blob/feat/agent/labelling/docs/parca-agent-labelling.md), everything gathered in `/proc` is under a PID, but we can check, which metadata provider are you worried?

pkg/discovery/systemd.go
kakkoyun: Yeah, exactly that (I've also fixed the suggestion). Semantically name and what it does not resonate for me 😅

kakkoyun (approved):

kakkoyun:
Let's merge and test it further.

pkg/discovery/systemd.go
```diff
@@ -50,111 +45,65 @@ func (c *SystemdConfig) NewDiscoverer(d DiscovererOptions) (Discoverer, error) {
...
-	pid, err := strconv.Atoi(strings.TrimSpace(string(output)))
-	if err != nil {
-		return nil, err
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case up <- groups:
+			}
+		case err := <-errCh:
+			level.Error(c.logger).Log("msg", "received error from systemd D-Bus API", "err", err)
```
kakkoyun, line 104:
```suggestion
			level.Warn(c.logger).Log("msg", "received error from systemd D-Bus API", "err", err)
```

nit: Let's lower the level. It's not an error that stops operations.

kakkoyun: Same goes other error levels.

pkg/discovery/systemd.go
```diff
@@ -50,111 +45,65 @@ func (c *SystemdConfig) NewDiscoverer(d DiscovererOptions) (Discoverer, error) {
...
-	cmd := exec.CommandContext(ctx, "systemctl", "list-units", "--type=service", "--state=running")
-	stdout, err := cmd.StdoutPipe()
+	conn, err := systemd.NewWithContext(ctx)
 	if err != nil {
-		return nil, err
-	}
-	if err := cmd.Start(); err != nil {
-		return nil, err
+		level.Error(c.logger).Log("msg", "failed to connect to systemd D-Bus API", "err", err)
+		return err
```
kakkoyun, line 51:
```suggestion
		return fmt.Errorf("failed to connect to systemd D-Bus API, %w", err)
```
Errors should be handled once and logging is error handling, so let's propagate the error to the caller and let it handle the error.

pkg/discovery/systemd.go
```diff
@@ -50,111 +45,65 @@ func (c *SystemdConfig) NewDiscoverer(d DiscovererOptions) (Discoverer, error) {
...
+
+			for unit, status := range update {
+				if status == nil || status.SubState != "running" {
+					groups = append(groups, &Group{Source: unit})
+					continue
+				}
+
+				mainPIDProperty, err := conn.GetServicePropertyContext(ctx, unit, "MainPID")
+				if err != nil {
+					level.Error(c.logger).Log("msg", "failed to get MainPID property for service", "err", err, "unit", unit)
```
maxbrunet, line 78:
```suggestion
					level.Warn(c.logger).Log("msg", "failed to get MainPID property for service", "err", err, "unit", unit)
```

pkg/discovery/systemd.go
```diff
@@ -50,111 +45,65 @@ func (c *SystemdConfig) NewDiscoverer(d DiscovererOptions) (Discoverer, error) {
...
+					continue
+				}
+
+				pid, ok := mainPIDProperty.Value.Value().(uint32)
+				if !ok {
+					level.Error(c.logger).Log("msg", "failed to assert type of PID", "unit", unit)
```
maxbrunet, line 84:
```suggestion
					level.Warn(c.logger).Log("msg", "failed to assert type of PID", "unit", unit)
```

javierhonduco:
FYI, this feature uses ~15% of the CPU cycles as of `99d1c7927379775e328c9db0222dd6169db6a116` . Not sure how much we spent before spawning to `systemctl` but wanted to surface the current cost

image

https://pprof.me/8b20774

kakkoyun:
Shall we open an issue for this?

javierhonduco:
Might be worth it, as I am not sure of the previous performance implications so this might be a win, perf-wise, but I am worried about the extra GC pressure this is adding

PR #972, opened 2022-11-01
feat: set default node name

maxbrunet:
Maybe you've discuss it or thought about it some time ago?

This defaults `--node` to the host name provided by the kernel, which is correct if not running in a container. Making all `parca-agent` flags optional.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -294,7 +294,7 @@ internal/pprof:
 
 # other artifacts:
 $(OUT_DIR)/help.txt: $(OUT_BIN)
-	$(OUT_BIN) --help > $@
+	$(OUT_BIN) --help | sed 's/--node=".*" */--node="hostname"           /' >$@
 
 DOC_VERSION := "latest"
 .PHONY: deploy/manifests
--- a/Makefile
+++ b/Makefile
@@ -294,6 +294,8 @@ internal/pprof:
 
 # other artifacts:
 $(OUT_DIR)/help.txt: $(OUT_BIN)
+	# The default value of --node is dynamic and depends on the current host's name
+	# so we replace it with something static.
 	$(OUT_BIN) --help | sed 's/--node=".*" */--node="hostname"           /' >$@
 
 DOC_VERSION := "latest"
--- a/cmd/parca-agent/main.go
+++ b/cmd/parca-agent/main.go
@@ -125,6 +125,7 @@ func main() {
 
 	if flags.Node == "" && hostnameErr != nil {
 		level.Error(logger).Log("msg", "failed to get host name. Please set it with the --node flag", "err", hostnameErr)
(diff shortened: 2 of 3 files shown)


cmd/parca-agent/main.go
```diff
@@ -114,11 +114,19 @@ type Profiler interface {
...
 	flags := flags{}
-	kong.Parse(&flags)
+	kong.Parse(&flags, kong.Vars{
+		"hostname": hostname,
+	})
 
 	logger := logger.NewLogger(flags.LogLevel, logger.LogFormatLogfmt, "parca-agent")
 
+	if flags.Node == "" && hostnameErr != nil {
+		level.Error(logger).Log("msg", "failed to get host name. Please set it with the --node flag", "err", hostnameErr)
```
kakkoyun, line 127:
```suggestion
		level.Error(logger).Log("msg", "failed to get host name. Please set it with the --node flag", "err", hostnameErr)
		os.Exit(1)
```

We should fail early in this case. Or we can move this check into the `run` and just return an error.

cmd/parca-agent/main.go
```diff
@@ -114,11 +114,19 @@ type Profiler interface {
 }
 
 func main() {
+	hostname, hostnameErr := os.Hostname()
```
kakkoyun, line 117: I wonder if the returned value matches the `nodeName`? https://github.com/parca-dev/parca-agent/blob/5ed98c6e3563fe15078bf5b167bdfd949a574c82/deploy/lib/parca-agent/parca-agent.libsonnet#L303-L312

Makefile
```diff
@@ -294,7 +294,7 @@ internal/pprof:
 
 # other artifacts:
 $(OUT_DIR)/help.txt: $(OUT_BIN)
-	$(OUT_BIN) --help > $@
+	$(OUT_BIN) --help | sed 's/--node=".*" */--node="hostname"           /' >$@
```
kakkoyun, line 297: It would be nice to have a comment on why we do the transformation.

kakkoyun (commented):
I don't remember if we have discussed this however this looks good.

cmd/parca-agent/main.go
maxbrunet: No, it does not, it is the name of the pod. This can be checked in a container with `uname -n`. So passing `nodeName` with the downward API is still needed

cmd/parca-agent/main.go
maxbrunet: Good catch! I intended to fail here

kakkoyun (approved):

brancz:
I don't feel strongly, but I think I actually like that people have to be very explicit in choosing this. But if people think this is annoying as is, I'm happy to go this way.

PR #978, opened 2022-11-02
Native stack walking with DWARF v2

javierhonduco:
This PR implements the v2 of the DWARF stack unwinder:

- The first commit reduces the unwind table size by half, allowing for more entries to fit in the table;
- The second commit removes `bpf_loop`, reducing the minimum required kernel version;
- The third commit reenables the CPU tests;

The details for each commit can be found in their commit message.

## Test Plan

Ensured, visually, that the stacks are correct:

**`testdata/out/basic-cpp-no-fp`**:

<details>
image
</details>

**CRuby interpreter**

<details>
image
[...]
image
</details>

### DIFF
--- a/bpf/cpu/cpu.bpf.c
+++ b/bpf/cpu/cpu.bpf.c
@@ -18,26 +18,36 @@
 
 /*================================ CONSTANTS =================================*/
 
-#define MAX_STACK_DEPTH 127 // Max depth of each stack trace to track
-
-#define MAX_STACK_ADDRESSES 1024 // Num of unique stack traces.
-#define MAX_ENTRIES                                                            \
-  10240 // Num entries for the `stack_counts` map. Unused atm.
-        // TODO(javierhonduco): use this later on.
-#define UNWIND_TABLES_MAX_STACK_DEPTH                                          \
-  100 // Max depth of each stack trace to track. TODO(javierhonduco): just to
-      // debug. Set to a larger number.
-#define MAX_PID_MAP_SIZE                                                       \
-  256 // Size of the `<PID, unwind_table>` mapping. Determines how many
-      // processes we can unwind.
-#define MAX_BINARY_SEARCH_DEPTH                                                \
-  20 // Binary search iterations. 2ˆ20 can bisect ~1_048_576 entries.
-#define MAX_UNWIND_TABLE_SIZE 250 * 1000 // Size of the unwind_table.
+// Number of frames to walk per tail call iteration.
+#define MAX_STACK_DEPTH_PER_PROGRAM 70
+// Number of frames to walk in total.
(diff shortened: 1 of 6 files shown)


pkg/profiler/cpu/maps.go
```diff
@@ -167,37 +181,46 @@ func (m *bpfMaps) setUnwindTable(pid int, ut unwind.UnwindTable) error {
...
+			return fmt.Errorf("write CFA register bytes: %w", err)
+		}
+
 		// Write CFA.
 		switch row.CFA.Rule {
 		case frame.RuleCFA:
 			// Write CFA register.
-			if err := binary.Write(buf, m.byteOrder, row.CFA.Reg); err != nil {
+			var CFARegister uint16
+			if row.CFA.Reg == 6 {
```
kakkoyun, line 194: nit: Let's use some constants instead of magic numbers here.

bpf/cpu/cpu.bpf.c
```diff
@@ -252,117 +268,102 @@ bpf_map_lookup_or_try_init(void *map, const void *key, const void *init) {
...
-  u64 table_len = unwind_table->table_len;
-  // Just for debugging to ensure that the data we are reading
-  // matches what we wrote.
-  bpf_printk("- unwind table has %d items", table_len);
 
-  // Invariant check.
-  if (table_len >= MAX_UNWIND_TABLE_SIZE) {
-    bpf_printk("should never happen");
-    BUMP_UNWIND_SHOULD_NEVER_HAPPEN_ERROR();
+SEC("perf_event")
```
kakkoyun, line 326: Why is this needed? I always thought it was for hooking to the event. Does it also use for exporting functions?

kakkoyun (commented):
Awesome job! 
v2 is here even before we released the v1!

I've some minor comments.

pkg/profiler/cpu/maps.go
javierhonduco: Good shout Was planning to do it but it seems I forgot about it

bpf/cpu/cpu.bpf.c
javierhonduco: Yes, all the other functions are inlined and one single program is executed. The compiler has replaced every call site with its body and they don't really exist as functions or individual programs anymore. 

As we are doing a tail call to a function, this has to be present and has to be the same BPF type as the BPF program that invokes it

kakkoyun (approved):
LGTM

zdyj3170101136:
is cfa_offset int16 enough?

it represent the stack size, int32 is recommend?

brancz:
Note that this code is entirely replaced starting v0.32

PR #989, opened 2022-11-03
ui: Round time duration in Next Profile Started

### LINKED ISSUE #628: web ui: Round the decimals for "next profile started"
## Context
Parca Agent's UI exposes a web UI on port 7071 by default. We use this interface to troubleshoot issues, as well as to expose some internal Golang runtime information. 

It looks like this:

image

As you can see, the "Next Profile Started" header shows a lot of decimals. Ideally, we would like to show one or maybe two decimals

## Repro
The [contributing doc](https://github.com/parca-dev/parca-agent/blob/main/CONTRIBUTING.md) contains all the information to get you up and running, if that's not the case that's a bug and we would really appreciate if you could open an issue with as much information as you can gather!

Then you can run something like
> ```make && sudo dist/parca-agent --node=test --log-level=debug --kubernetes=false --insecure```

which will start Parca Agent, including the server you will need to fix :)

Parca Agent's entry point can be found [here](https://github.com/parca-dev/parca-agent/blob/main/cmd/parca-agent/main.go)

Sylfrena:
This is how it looks after the changes.
image

Fixes #628

### DIFF
--- a/cmd/parca-agent/main.go
+++ b/cmd/parca-agent/main.go
@@ -350,7 +350,7 @@ func run(logger log.Logger, reg *prometheus.Registry, flags flags) error {
 			for _, profiler := range profilers {
 				statusPage.ActiveProfilers = append(statusPage.ActiveProfilers, template.ActiveProfiler{
 					Name:           profiler.Name(),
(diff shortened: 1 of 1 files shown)


kakkoyun (approved):
LGTM

PR #990, opened 2022-11-03
*: Add ability to set memlock rlimit

kakkoyun:

cc @javierhonduco

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,8 @@ Flags:
       --config-path="parca-agent.yaml"
                                   Path to config file.
       --memlock-rlimit=4294967296
-                                  The value for the RLimit for memlock. Used
+                                  The value for the maximum number of bytes of
+                                  memory that may be locked into RAM. It is used
                                   to ensure the agent can lock memory for eBPF
                                   maps. 0 means no limit.
       --profiling-duration=10s    The agent profiling duration to use. Leave
--- a/README.md
+++ b/README.md
@@ -66,6 +66,10 @@ Flags:
                                   the Kubernetes node name.
       --config-path="parca-agent.yaml"
                                   Path to config file.
+      --memlock-rlimit=4294967296
+                                  The value for the RLimit for memlock. Used
+                                  to ensure the agent can lock memory for eBPF
+                                  maps. 0 means no limit.
       --profiling-duration=10s    The agent profiling duration to use. Leave
                                   this empty to use the defaults.
       --metadata-external-labels=KEY=VALUE;...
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -66,6 +66,10 @@ Flags:
                                   the Kubernetes node name.
       --config-path="parca-agent.yaml"
                                   Path to config file.
+      --memlock-rlimit=4294967296
+                                  The value for the RLimit for memlock. Used
```
javierhonduco, line 70: nit: Should we specify the unit?

cmd/parca-agent/main.go
```diff
@@ -72,13 +73,17 @@ var (
...
 type flags struct {
 	LogLevel    string `kong:"enum='error,warn,info,debug',help='Log level.',default='info'"`
 	HTTPAddress string `kong:"help='Address to bind HTTP server to.',default=':7071'"`
 
-	Node string `kong:"help='The name of the node that the process is running on. If on Kubernetes, this must match the Kubernetes node name.',default='${hostname}'"`
-
-	ConfigPath string `default:"parca-agent.yaml" help:"Path to config file."`
+	Node          string `kong:"help='The name of the node that the process is running on. If on Kubernetes, this must match the Kubernetes node name.',default='${hostname}'"`
+	ConfigPath    string `default:"parca-agent.yaml" help:"Path to config file."`
+	MemlockRlimit uint64 `default:"${default_memlock_rlimit}" help:"The value for the RLimit for memlock. Used to ensure the agent can lock memory for eBPF maps. 0 means no limit."`
```
javierhonduco, line 86: > ${default_memlock_rlimit}"

What! TIL! :D

pkg/profiler/rlimit.go
```diff
@@ -16,22 +16,35 @@ package profiler
...
+	}
+	if max == 0 {
+		max = unix.RLIM_INFINITY
+	}
 	rLimit := syscall.Rlimit{
 		Cur: cur,
 		Max: max,
 	}
+
+	rlimitMu.Lock()
```
javierhonduco, line 41: Why is this needed?

We don't call it concurrently no? Either way, it seems like it's thread-safe:

>         │getrlimit(), setrlimit(), prlimit()   │ Thread safety │ MT-Safe │

https://man7.org/linux/man-pages/man2/getrlimit.2.html

javierhonduco (approved):
Looks great!! 💯

README.md
kakkoyun: Good catch!

pkg/profiler/rlimit.go
kakkoyun: It's primarily future envy. We will have more than one profiler eventually, and this API is designed to be called from the profilers. Happy to get rid of it.

PR #997, opened 2022-11-08
.eh_frame: Initial ASLR support

javierhonduco:
Right now, processes that are loaded at a randomized address won't work
with the current dwarf-based stack unwinding code. The reason for this
is that the program counters aren't adjusted accordingly.

This commit adds a new package under `pkg/executable` that will host
code specific to executables. The `IsASLRElegible` is documented inline,
it's a pretty basic function and doesn't handle some edge cases, but
this should do for most of the processes I've tried in my host (see test
plan of this PR and in the upcoming sharded unwind table work).

In the future, this will be refactored in a more cohesive module that
will do more groundwork necessary for other parts of the profiler.
Caching will be added later on, too.

The first commit updates the testdata repository which adds the PIE executables.

## Test Plan

**Added unittest**
```
[javierhonduco@fedora parca-agent]$ go test -v github.com/parca-dev/parca-agent/pkg/executable
=== RUN   TestASLRDisabled
--- PASS: TestASLRDisabled (0.00s)
=== RUN   TestASLREnabled
--- PASS: TestASLREnabled (0.00s)
PASS
ok      github.com/parca-dev/parca-agent/pkg/executable (cached)
```

**e2e test**

``` 
./testdata/out/basic-cpp-plt-pie
``` 

``` 
make build-dyn ENABLE_ASAN=yes &&  sudo dist/parca-agent-dyn --node=test --remote-store-insecure --remote-store-address=localhost:7070 --experimental-dwarf-unwinding-pids=`pidof basic-cpp-plt-pie`
``` 

image

### DIFF
--- a/pkg/stack/unwind/unwind_table.go
+++ b/pkg/stack/unwind/unwind_table.go
@@ -28,6 +28,7 @@ import (
 	"github.com/prometheus/procfs"
 
 	"github.com/parca-dev/parca-agent/internal/dwarf/frame"
+	"github.com/parca-dev/parca-agent/pkg/executable"
 )
 
 // UnwindTableBuilder helps to build UnwindTable for a given PID.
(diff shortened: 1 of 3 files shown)


pkg/stack/unwind/unwind_table.go
```diff
@@ -130,16 +131,23 @@ func (ptb *UnwindTableBuilder) UnwindTableForPid(pid int) (UnwindTable, error) {
 		rows := buildUnwindTable(fdes)
 		level.Info(ptb.logger).Log("msg", "adding tables for mapped executable", "path", executablePath, "rows", len(rows), "low pc", fmt.Sprintf("%x", rows[0].Loc), "high pc", fmt.Sprintf("%x", rows[len(rows)-1].Loc))
 
-		// TODO(javierhonduco): Revisit this logic with PIE executables as it's only correct
-		// for non-PIE ones.
+		ASLRElegible, err := executable.IsASLRElegible(executablePath)
```
kakkoyun, line 134:
```suggestion
		aslrElegible, err := executable.IsASLRElegible(executablePath)
```

pkg/stack/unwind/unwind_table.go
```diff
@@ -130,16 +131,23 @@ func (ptb *UnwindTableBuilder) UnwindTableForPid(pid int) (UnwindTable, error) {
...
+		if err != nil {
+			return nil, fmt.Errorf("ASLR check failed with with: %w", err)
+		}
+
 		if strings.Contains(executablePath, mainExec) {
-			ut = append(ut, rows...)
+			if ASLRElegible {
```
kakkoyun, line 140: I wonder, do we need this check? Can't we add the mapping start to the locations? If it's not dynamically mapped `m.StartAddr` is probably zero.

javierhonduco: Yes, we need it, even if ASLR is not enabled, `StartAddr` will be non-zero, usually around `0x1000` (4k, to avoid mapping the first page, so accessing addresses `[0..firstpage_address)` will cause a SIGSGV), which is the default base address. This is already accounted for in the DWARF unwind tables, which is why we don't want to add it in this case.

javierhonduco:
Oops, sorry, thought auto merge would not only wait for CI, but also for at least 1x approval

PR #1000, opened 2022-11-08
ci(e2e): simplify setup

maxbrunet:
* Use Minikube's `none` driver to run Kubernetes on GH runner directly
* Only take care of the deployment in `ci-e2e.sh`, wrapping the `minikube` CLI does not add a lot of value and is complex depending on the driver used and host machine, everything is taken care by the setup action here, otherwise it's a simple `minikube start` (a simple comment on how to do it locally has been added with a mandatory Kube context parameter as a safe guard from unintentional deployments)
* Use Minikube's `stable` Kubernetes version (v1.25 currently), this should improve compatibility/stability. It will be bumped with each Minikube updates
* Hide PSP resources from Jsonnet output (removed in K8s v1.25) 
* Remove Minikube's `--feature-gates=EphemeralContainers=true`, now stable in v1.25
* ~~Remove Minikube's `--docker-opt default-ulimit=memlock=`, the bump should be taken care of by the agent~~
* Use a single Jsonnet command for the full deployment, increases flexibility
* Pin Parca server version
* Pipe Jsonnet output directly into `kubectl apply` (avoid going by YAML to disk in between)
* The port 7071 is already by `docker-proxy` on the host (this is one took me the most time to figure out as the error message was obscure and I had to create a local reproduction which once done was pretty quick to spot the issue). I removed the port-forward to the agent as it is unused at moment, another port like 7072 could be used in the future
* Use default config loading chain (`KUBECONFIG` env var, `~/.kube/config`) in e2e and add k8s context flag 

Closes #962

### DIFF
--- a/deploy/e2e.jsonnet
+++ b/deploy/e2e.jsonnet
@@ -1,4 +1,4 @@
-function(version='v0.0.1-alpha.3')
+function(version='v0.0.1-alpha.3', serverVersion='v0.0.3-alpha.2')
   local ns = {
     apiVersion: 'v1',
     kind: 'Namespace',
@@ -12,11 +12,24 @@ function(version='v0.0.1-alpha.3')
     },
   };
 
+  local server = (import 'github.com/parca-dev/parca/deploy/lib/parca/parca.libsonnet')({
+    name: 'parca',
+    namespace: ns.metadata.name,
+    image: 'ghcr.io/parca-dev/parca:' + self.version,
+    version: serverVersion,
+    replicas: 1,
+    corsAllowedOrigins: '*',
+  }) {
+    // Hide PSP: Removed in K8s 1.25
+    // TODO: Remove from lib or add option to to toggle it
+    podSecurityPolicy:: {},
+  };
+
   local agent = (import 'parca-agent/parca-agent.libsonnet')({
     name: 'parca-agent',
     namespace: ns.metadata.name,
     version: version,
-    image: 'ghcr.io/parca-dev/parca-agent-test:' + version,
+    image: 'ghcr.io/parca-dev/parca-agent-test:' + self.version,
     // This assumes there's a running parca in the cluster.
     stores: ['parca.parca.svc.cluster.local:7070'],
     insecure: true,
@@ -31,12 +44,26 @@ function(version='v0.0.1-alpha.3')
     //   Docs for usage of Label Selector
     //   https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors
     //   podLabelSelector: 'app=my-web-app,version=v1',
-  });
+  }) {
+    // Hide PSP: Removed in K8s 1.25
+    // TODO: Remove from lib or add option to to toggle it
+    podSecurityPolicy:: {},
+  };
+
 
   {
-    '0namespace': ns,
-  } + {
-    ['parca-agent-' + name]: agent[name]
-    for name in std.objectFields(agent)
-    if agent[name] != null
+    kind: 'List',
+    apiVersion: 'v1',
+    items:
+      [
+        ns,
+      ] + [
+        server[name]
+        for name in std.objectFields(server)
+        if server[name] != null
+      ] + [
+        agent[name]
+        for name in std.objectFields(agent)
+        if agent[name] != null
+      ],
   }
--- a/deploy/jsonnetfile.json
+++ b/deploy/jsonnetfile.json
@@ -3,20 +3,20 @@
   "dependencies": [
     {
       "source": {
-        "local": {
-          "directory": "lib/parca-agent"
-        }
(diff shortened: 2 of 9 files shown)


kakkoyun:
> * Remove Minikube's `--docker-opt default-ulimit=memlock=`, the bump should be taken care of by the agent

Even though the agent is bumping the limit, I remember adding this because it wasn't enough. But if you already tested this locally and it works, it's great!

e2e/ci-e2e.sh
```diff
@@ -13,79 +13,54 @@
...
-################################################################################
-
 set -euox pipefail
 
-function run() {
-    # Driver set to virtualbox by default if not specified
-    DRIVER=${1-virtualbox}
-    VERSION=$2
+# renovate: datasource=github-releases depName=parca-dev/parca
+SERVER_VERSION='v0.12.1'
```
kakkoyun, line 19:
```suggestion
SERVER_VERSION='v0.13.0'
```
ditto

e2e/e2e_test.go
```diff
@@ -87,11 +87,12 @@ func TestGRPCIntegration(t *testing.T) {
 	}
 	defer serverCloser()
 
-	agentCloser, err := StartPortForward(ctx, cfg, "https", parcaAgent, ns, "7071")
-	if err != nil {
-		require.NoError(t, err, "failed to start port forwarding Parca Agent: %v", err)
-	}
-	defer agentCloser()
+	// // Port 7071 may already be in use by docker-proxy on the host (e.g. Minikube with none driver)
```
kakkoyun, line 90: If we didn't this won't be used, let's delete the code.

scripts/local-dev.sh
```diff
@@ -65,12 +65,9 @@ function up() {
         mk start \
             --driver="${DRIVER}" \
             --nodes="${NODE_COUNT}" \
-            --kubernetes-version=v1.23.3 \
             --cpus=2 \
             --memory=8gb \
-            --disk-size=20gb \
-            --docker-opt dns=8.8.8.8 \
```
kakkoyun, line 72: Have we checked whether this config runs smoothly with Tilt locally? Like collecting metrics and visualizing them with Parca Server? All these are added for a reason, maybe things has changed, so let's make sure. 

This could also be because of using different drivers. Let's check with KVM and VirtualBox and make sure everything works. I can do these as well.

deploy/e2e.jsonnet
```diff
@@ -12,11 +12,24 @@ function(version='v0.0.1-alpha.3')
...
+    podSecurityPolicy:: {},
+  };
+
   local agent = (import 'parca-agent/parca-agent.libsonnet')({
     name: 'parca-agent',
     namespace: ns.metadata.name,
     version: version,
-    image: 'ghcr.io/parca-dev/parca-agent-test:' + version,
+    image: 'ghcr.io/parca-dev/parca-agent-test:' + self.version,
     // This assumes there's a running parca in the cluster.
```
Sylfrena, line 33: Not required for this PR but if we want to use image artifacts here instead to fix #934 then we can do away with this bit as well as we won't need `ghcr.io/parca-dev/parca-agent-test` any longer.

deploy/jsonnetfile.json
```diff
@@ -3,20 +3,20 @@
...
       "source": {
-        "local": {
-          "directory": "lib/parca-agent"
+        "git": {
+          "remote": "https://github.com/parca-dev/parca.git",
+          "subdir": "deploy/lib/parca"
         }
       },
-      "version": ""
+      "version": "v0.12.1"
```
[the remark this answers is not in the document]
kakkoyun: We will release v0.14.0 today and it will have all the needed fixes.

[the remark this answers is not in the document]
maxbrunet: If this PR is ready before the release, I am of the opinion of going ahead with 0.12 and let Renovate to the follow up, otherwise no problem

e2e/e2e_test.go
maxbrunet: It was not straightforward to troubleshoot, the comment might save someone's time in a not so far future?

kakkoyun: We can leave the comment. Do we need to code?

scripts/local-dev.sh
maxbrunet: I've never tried Tilt, but I've used half of the drivers and never faced DNS issues with Minikube. I'm ok with reverting these changes to shorten the discussion since this script is a little out of the scope of the PR, just trying keep it close to the CI

kakkoyun (approved):
LGTM

kakkoyun:
@maxbrunet There should be a special reward for this PR #1000 🎉 🌮 🥳

PR #1001, opened 2022-11-09
Reduce elf.Open() calls

### LINKED ISSUE #964: Reduce ELF `Open` calls
## Background

Currently, we are spending ~10% of the Agent's CPU cycles opening ELF files. We need to do this for our profiler to work, for example, the `build_id` is stored in an ELF section, but the current APIs result in excessive calls to `Open()` as the flamegraph below shows, as they mostly accept a string with the path to the elf file rather than an `*elf.File`.

## Potential impact 
While the savings I think this would get us are not very big (perhaps ~3%) I think that the APIs to deal with ELF files would be clearer. Something that we should measure once these changes land is how memory allocations and system calls are affected, as they should be reduced, too.

## CPU (cycles) profile
[link to the profile](https://demo.parca.dev/?currentProfileView=icicle&expression_a=parca_agent_cpu%3Asamples%3Acount%3Acpu%3Ananoseconds%3Adelta%7Bapp_kubernetes_io_name%3D%22parca-agent%22%7D&from_a=1666850084029&to_a=1666853684029&merge_a=true&time_selection_a=relative:hour|1)
image

Each of the highlighter frames on the left is ~3% of the total Agent CPU. The smaller ones on the right, are around 0.3% CPU each

Sylfrena:
#### Before

CPU usage for `elf.Open()` = **39.73 %** (13.63 + 13.42 + 12.68)

Screenshot from 2022-11-24 14-12-30

#### After

CPU usage for `elf.Open()` = **22.89 %**

Screenshot from 2022-11-24 14-13-12

Fixes #964

### DIFF
--- a/pkg/buildid/buildid.go
+++ b/pkg/buildid/buildid.go
@@ -28,11 +28,13 @@ import (
 	"github.com/parca-dev/parca-agent/pkg/elfreader"
 )
 
-func BuildID(path string) (string, error) {
-	f, err := elf.Open(path)
+func BuildID(f *elf.File, path string) (string, error) {
+	/*f, err := elf.Open(path)
 	if err != nil {
 		return "", fmt.Errorf("failed to open elf: %w", err)
 	}
+	defer f.Close()
+	*/
 
 	hasGoBuildIDSection := false
 	for _, s := range f.Sections {
@@ -42,9 +44,9 @@ func BuildID(path string) (string, error) {
 	}
 
 	if hasGoBuildIDSection {
-		f.Close()
+		//f.Close()
 
-		if id, err := fastGoBuildID(path); err == nil && len(id) > 0 {
+		if id, err := fastGoBuildID(f); err == nil && len(id) > 0 {
 			return hex.EncodeToString(id), nil
 		}
 
@@ -55,16 +57,16 @@ func BuildID(path string) (string, error) {
 
 		return hex.EncodeToString([]byte(id)), nil
 	}
-	f.Close()
+	//f.Close()
 
-	if id, err := fastGNUBuildID(path); err == nil && len(id) > 0 {
+	if id, err := fastGNUBuildID(f); err == nil && len(id) > 0 {
 		return hex.EncodeToString(id), nil
 	}
 
 	return elfBuildID(path)
 }
 
-func fastGoBuildID(path string) ([]byte, error) {
+func fastGoBuildID(f *elf.File) ([]byte, error) {
 	findBuildID := func(notes []elfreader.ElfNote) ([]byte, error) {
 		var buildID []byte
 		for _, note := range notes {
@@ -78,14 +80,14 @@ func fastGoBuildID(path string) ([]byte, error) {
 		}
 		return buildID, nil
 	}
-	data, err := extractNote(path, ".note.go.buildid", findBuildID)
+	data, err := extractNote(f, ".note.go.buildid", findBuildID)
 	if err != nil {
 		return nil, err
 	}
 	return data, nil
 }
 
-func fastGNUBuildID(path string) ([]byte, error) {
+func fastGNUBuildID(f *elf.File) ([]byte, error) {
 	findBuildID := func(notes []elfreader.ElfNote) ([]byte, error) {
 		var buildID []byte
 		for _, note := range notes {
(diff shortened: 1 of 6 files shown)


pkg/buildid/buildid.go
```diff
@@ -28,11 +28,13 @@ import (
 	"github.com/parca-dev/parca-agent/pkg/elfreader"
 )
 
-func BuildID(path string) (string, error) {
-	f, err := elf.Open(path)
+func BuildID(f *elf.File, path string) (string, error) {
```
kakkoyun, line 31:
```suggestion
func BuildID(f *elf.File) (string, error) {
```

Better to eliminate the `path` to unify. Even if it's passed to a helper function we should refactor it as well.

Sylfrena: Made the [changes](https://github.com/parca-dev/parca-agent/pull/1001/commits/bdf32954e9e29e137ea4f8e1ef790bf93f340077)- but they seem a bit overkill to me(although cleaner). Let me know if there might be a better approach I am missing out on.

pkg/buildid/buildid_test.go
```diff
@@ -62,7 +63,11 @@ func TestBuildID(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			got, err := BuildID(tt.args.path)
+			file, err := elf.Open(tt.args.path)
+			require.NoError(t, err)
+			elfFile := &ElfFile{Path: tt.args.path, File: file}
+
+			got, err := BuildID(elfFile)
```
Sylfrena, line 70:
```suggestion
			got, err := BuildID(&ElfFile{Path: tt.args.path, File: file})
```

pkg/buildid/buildid.go
```diff
@@ -28,43 +28,39 @@ import (
 	"github.com/parca-dev/parca-agent/pkg/elfreader"
 )
 
-func BuildID(path string) (string, error) {
-	f, err := elf.Open(path)
-	if err != nil {
-		return "", fmt.Errorf("failed to open elf: %w", err)
-	}
+type ElfFile struct {
+	Path string
```
kakkoyun, line 32: This is more explicit, for sure. However, you can also use https://pkg.go.dev/os#File.Name

`Name returns the name of the file as presented to Open.`

You can ensure the file is open with an absolute path and everything can be consistent. This is not even needed, we don't do any async work on the paths, and we don't change the running directory during the agent's lifetime.

kakkoyun: So I think we can eliminate this additional abstraction.

pkg/buildid/buildid.go
kakkoyun: https://github.com/parca-dev/parca-agent/pull/1001#discussion_r1029046143

pkg/objectfile/object_file.go
```diff
@@ -191,15 +193,6 @@ func (f *ObjectFile) computeBase(addr uint64) error {
 	return nil
 }
 
-// mapping stores the parameters of a runtime mapping that are needed to
-// identify the ELF segment associated with a mapping.
-type mapping struct {
```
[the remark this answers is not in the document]
Sylfrena: Ah, I think this changed because I put all structs at the beginning of the file to make refactoring easier and then forgot to put it back 🙈 
I will put it back with a comment if you want

pkg/buildid/buildid.go
kakkoyun: Oh, I see. They don't embed the `os.File` even though they use it underneath. `debug/elf` is not the best and most eloquent Go package out there, for sure. How about using the existing `ObjectFile` struct and passing it around? I want to simplify things. Having all these ephemeral types confuses the developer.

How about converging everything on the `ObjectFile` type? Open it once and pass the instances of that around. If it's too much work, I'm fine with this. Nothing to do 🤷

pkg/objectfile/object_file_test.go
```diff
@@ -82,14 +80,6 @@ func TestComputeBase(t *testing.T) {
 			addr:      0x1000,
 			wantError: true,
 		},
-		{
```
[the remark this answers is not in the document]
Sylfrena: Fair, I removed it because we are no longer calling `elf.Open()` from `computeBase`; shall I create a new `TestOpenElf` and add it there (and move `TestOpenMalformedElf` in there too?)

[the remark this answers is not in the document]
kakkoyun: That would be perfect!

pkg/buildid/buildid.go
kakkoyun: As you said, it is a long way. And the Go team is not very keen on accepting changes to these packages. But we can try.

Sylfrena: > Oh, I see. They don't embed the `os.File` even though they use it underneath. `debug/elf` is not the best and most eloquent Go package out there, for sure. How about using the existing `ObjectFile` struct and passing it around? I want to simplify things. Having all these ephemeral types confuses the developer.
> 
> How about converging everything on the `ObjectFile` type? Open it once and pass the instances of that around. If it's too much work, I'm fine with this. Nothing to do shrug

It is not a lot of work but will cause a cyclic dependency between the `objectfile` and `buildid` packages; having both in one package will complicate the api more rn imo.
How about we either
 - just pass `path` as a parameter right now(and add a TODO there) or
 - stay with this `ElfFile`

till we come up with a better api or wait and see if we can add a `Name` method in the stdlib?

kakkoyun: Sounds good. Let's go with the `ElfFile` approach.

kakkoyun (approved):

PR #1004, opened 2022-11-09
.eh_frame: Improve RBP handling

javierhonduco:
The unwinder didn't handle some important cases such as when an expression has to be evaluated to find its value, or when it's been stored in another register. While we still don't handle these, let's be explicit about them and log ccordingly.

We also relied on the offset value to signify that the value did not change, this commits makes this explicit.

The size of the unwind table is the same, the previous field that stored the cfa type (`.cfa_type`, a u16) has been split in two fields with half the size (u8), `.cfa_type` and `.rbp_type`.

## Test Plan
Tested in Minikube and CRuby without issues.

### DIFF
--- a/pkg/profiler/cpu/maps.go
+++ b/pkg/profiler/cpu/maps.go
@@ -44,16 +44,23 @@ const (
 	maxUnwindSize         = maxUnwindTableSize * unwindTableShardCount
 )
 
-type BPFCFAType uint16
+type BpfCfaType uint16
 
 const (
-	CFARegisterUndefined BPFCFAType = iota
-	CFARegisterRbp
-	CFARegisterRsp
-	CFARegisterExpression
+	CfaRegisterUndefined BpfCfaType = iota
+	CfaRegisterRbp
+	CfaRegisterRsp
+	CfaRegisterExpression
 )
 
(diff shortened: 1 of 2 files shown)


pkg/profiler/cpu/maps.go
```diff
@@ -44,16 +44,23 @@ const (
...
+	CfaRegisterRbp
+	CfaRegisterRsp
+	CfaRegisterExpression
 )
 
-const CFAFakeOffset = int16(0xCFA)
+type BPFRbpType uint16
+
+const (
+	RBPRuleOffsetUnchanged BPFRbpType = iota
```
kakkoyun, line 59:
```suggestion
	RbpRuleOffsetUnchanged BPFRbpType = iota
```
If we want to be consistent with your other naming changes. I know it's hard with abbreviations. Let's at least be consistent.

kakkoyun (commented):
LGTM.

Just a comment on naming constant.

pkg/profiler/cpu/maps.go
javierhonduco: Ohh true, will fix 😄

kakkoyun (approved):

PR #1005, opened 2022-11-09
elfwriter-test: Clean the files

v-thakkar:

### DIFF
--- a/pkg/elfwriter/elfwriter_test.go
+++ b/pkg/elfwriter/elfwriter_test.go
@@ -217,7 +217,7 @@ func TestWriter_WriteCompressedHeaders(t *testing.T) {
 	file, err := os.Open("testdata/libc_compressed.debug")
 	require.NoError(t, err)
 	t.Cleanup(func() {
-		defer file.Close()
+		file.Close()
 	})
 
 	input, err := elf.NewFile(file)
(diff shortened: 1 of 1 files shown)


pkg/elfwriter/elfwriter_test.go
```diff
@@ -265,7 +265,7 @@ func TestWriter_HasLinks(t *testing.T) {
 	inElf, err := elf.Open("testdata/libc.so.6")
 	require.NoError(t, err)
 	t.Cleanup(func() {
-		inElf.Close()
+		defer inElf.Close()
```
kakkoyun, line 268:
```suggestion
	inElf.Close()
```

You don't need to `defer` in here.

kakkoyun (changes requested):

pkg/elfwriter/elfwriter_test.go
v-thakkar: Ok, I see that it's part of idiomatic Go. Thanks. Removed it at one other place in the file too.

kakkoyun: 👍  It's not about being idiomatic here. First of all, there are no other statements in that scope (the scope is the closure here). And `t.Cleanup` actually does the trick of the `defer.` It gets to run after the test exists.

kakkoyun (approved):

pkg/elfwriter/elfwriter_test.go
v-thakkar: Ah, yes. I see what you're saying now. :+1:

javierhonduco (commented):
Thanks for doing this!!

PR #1007, opened 2022-11-10
*: Add flag to allow skipping stripping of binaries

brancz:
There may be cases where we want to investigate whether there is a problem with re-written binaries, in those cases, it would be handy to have a flag to turn it off.


kakkoyun (approved):
LGTM

v-thakkar (approved):
I think README needs to be updated but otherwise LGTM.

PR #1010, opened 2022-11-10
Improve the agent metrics and add native histograms

metalmatze:
Consolidate the three existing metrics into one metric and differentiate by using different labels. Additionally, track the total amount of profile attempts (by counting the successful attempts) so that we can compute the percentage of errors / successes. 

Fortunately, client_golang 1.14 was released and we already use that version. Hence, I've added two native histograms to track the duration of obtaining histograms and the duration of symbolizing and converting to pprof.


kakkoyun (approved):
LGTM ❤️

PR #1021, opened 2022-11-15
.eh_frame: Pass the right pidtgid to add_stacks

javierhonduco:
In [0] `add_stacks` was modified to accept a u64 containing the tgid and pid in the form of `current_task->tgid << 32 | current_task->pid`, as the `bpf_get_current_pid_tgid` returns, but in the case of the dwarf based unwinder, only the 32 bit value containing the pid was passed

- [0]: Commit https://github.com/parca-dev/parca-agent/commit/3fba1033ce42a64bb53c206d64547dd8b3a5c283

## Test Plan

**Before this PR**
image

**After this PR**

image
image

### DIFF
--- a/bpf/cpu/cpu.bpf.c
+++ b/bpf/cpu/cpu.bpf.c
@@ -380,15 +380,24 @@ static __always_inline stack_unwind_table_t *find_unwind_table(pid_t pid,
   return NULL;
 }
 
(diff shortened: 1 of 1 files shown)


kakkoyun (approved):
LGTM. Thanks for better explanation.

And sorry for breaking things for unwinder.

PR #1022, opened 2022-11-15
buildid: Reduce `elf.Open` calls

marselester:
Hi! In this PR I tried to reduce `elf.Open` calls as mentioned in this issue https://github.com/parca-dev/parca-agent/issues/964.

### DIFF
--- a/internal/pprof/binutils/binutils.go
+++ b/internal/pprof/binutils/binutils.go
@@ -433,10 +433,8 @@ func (b *binrep) openELF(name string, start, limit, offset uint64, relocationSym
 	defer ef.Close()
 
 	buildID := ""
-	if f, err := os.Open(name); err == nil {
-		if id, err := elfexec.GetBuildID(f); err == nil {
-			buildID = fmt.Sprintf("%x", id)
-		}
+	if id, err := elfexec.GetBuildID(ef); err == nil {
+		buildID = fmt.Sprintf("%x", id)
 	}
 
 	var (
--- a/internal/pprof/elfexec/elfexec.go
+++ b/internal/pprof/elfexec/elfexec.go
@@ -118,12 +118,7 @@ func parseNotes(reader io.Reader, alignment int, order binary.ByteOrder) ([]elfN
 //
 // If no build-ID was found but the binary was read without error, it returns
 // (nil, nil).
-func GetBuildID(binary io.ReaderAt) ([]byte, error) {
-	f, err := elf.NewFile(binary)
-	if err != nil {
-		return nil, err
-	}
-
+func GetBuildID(f *elf.File) ([]byte, error) {
 	findBuildID := func(notes []elfNote) ([]byte, error) {
 		var buildID []byte
 		for _, note := range notes {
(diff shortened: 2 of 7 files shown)


internal/pprof/binutils/binutils.go
```diff
@@ -433,10 +433,8 @@ func (b *binrep) openELF(name string, start, limit, offset uint64, relocationSym
 	defer ef.Close()
 
 	buildID := ""
-	if f, err := os.Open(name); err == nil {
-		if id, err := elfexec.GetBuildID(f); err == nil {
-			buildID = fmt.Sprintf("%x", id)
-		}
+	if id, err := elfexec.GetBuildID(ef); err == nil {
```
kakkoyun, line 436: We shouldn't change the code in `internal/pprof`. This is vendored code, and all the changes will be lost when we run the script to update.

internal/pprof/elfexec/elfexec.go
```diff
@@ -118,12 +118,7 @@ func parseNotes(reader io.Reader, alignment int, order binary.ByteOrder) ([]elfN
 //
 // If no build-ID was found but the binary was read without error, it returns
 // (nil, nil).
-func GetBuildID(binary io.ReaderAt) ([]byte, error) {
-	f, err := elf.NewFile(binary)
-	if err != nil {
-		return nil, err
-	}
-
+func GetBuildID(f *elf.File) ([]byte, error) {
```
kakkoyun, line 121: ditto.

pkg/buildid/buildid.go
```diff
@@ -28,12 +27,8 @@ import (
...
 )
 
-func BuildID(path string) (string, error) {
-	f, err := elf.Open(path)
-	if err != nil {
-		return "", fmt.Errorf("failed to open elf: %w", err)
-	}
-
+// BuildID reads the build ID from an object file.
+func BuildID(path string, f *elf.File) (string, error) {
```
kakkoyun, line 31: We should also eliminate the use of the `path` here. I think can be read from `file.Name` or similar.

```suggestion
func BuildID(f *elf.File) (string, error) {
```

kakkoyun (commented):
Thanks for a lot for the contribution.
I have commented.

pkg/buildid/buildid.go
marselester: I checked the source code of the elf package and couldn't find a way to get the path, that's why the solution is not elegant :(

internal/pprof/binutils/binutils.go
marselester: oh, ok 👍

pkg/buildid/buildid.go
kakkoyun: If we pass the file that the path points to, we still have that information in the file, no? Or does the path point to something else?

v-thakkar:
Thanks for the PR @marselester. Appreciate your contribution but as you can see @Sylfrena is already working on this issue with her WIP PR: https://github.com/parca-dev/parca-agent/pull/1001. So let's handle this issue in that PR to avoid duplicating the efforts. Let us know if you need help finding issues to work on. Feel free to comment on the issue you're interested in and we're always happy to assign it to you. :)

marselester:
Sorry about that, somehow I missed that PR!

PR #1024, opened 2022-11-15
refactor(metadata): consolidate cgroup/process into procfs provider

maxbrunet:


pkg/metadata/procfs.go
```diff
@@ -0,0 +1,94 @@
...
+
+	if len(cgroups) == 1 || len(cgroups) == 2 {
+		cgroup = cgroups[0]
+	} else {
+	cgroupsLoop:
+		for _, cg := range cgroups {
+			for _, ctlr := range cg.Controllers {
+				if ctlr == "cpu" {
+					cgroup = cg
+					break cgroupsLoop
```
kakkoyun, line 75:
```suggestion
					return cg
```
I think it would be easier to read. No need to use the tags.

pkg/metadata/procfs.go
```diff
@@ -0,0 +1,94 @@
...
+			"executable":  model.LabelValue(executable),
+			"ppid":        model.LabelValue(strconv.Itoa(stat.PPID)),
+		}, nil
+	}}
+}
+
+func findCPUCgroup(cgroups []procfs.Cgroup) procfs.Cgroup {
+	var cgroup procfs.Cgroup
+
+	if len(cgroups) == 1 || len(cgroups) == 2 {
```
kakkoyun, line 67: It would be nice to explain the reasoning of the cases here as comments.

I know the previous version didn't also have them. The cases don't make any sense to me right now.

kakkoyun (commented):
Thanks for cleaning things up. I really like the procfs library.
I have a couple of comments inline.

cmd/parca-agent/main.go
```diff
@@ -308,9 +308,8 @@ func run(logger log.Logger, reg *prometheus.Registry, flags flags) error {
 		[]metadata.Provider{
 			metadata.ServiceDiscovery(logger, discoveryManager),
 			metadata.Target(flags.Node, flags.MetadataExternalLabels),
-			metadata.Cgroup(),
 			metadata.Compiler(),
-			metadata.Process(),
+			metadata.Procfs(),
```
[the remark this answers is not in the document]
maxbrunet: I like procfs :slightly_smiling_face:

pkg/metadata/procfs.go
maxbrunet: `len(cgroups) == 2` was wrong, it was for trailing new lines, I removed it, otherwise I do not know myself. I called the function `findCPUCgroup` because I assume that is what we are looking for. @javierhonduco any suggestions?

pkg/metadata/process.go
```diff
@@ -25,28 +27,62 @@ func Process() Provider {
...
+				return cg
+			}
+
+			for _, ctlr := range cg.Controllers {
+				if strings.Contains(ctlr, "systemd") {
+					return cg
+				}
+			}
+		}
+	}
```
kakkoyun, line 85:
```suggestion
	for _, cg := range cgroups {
		for _, ctlr := range cg.Controllers {
			if ctlr == "cpu" {
				return cg
			}
		}

		if strings.Contains(cg.Path, "systemd") {
			return cg
		}

		for _, ctlr := range cg.Controllers {
			if strings.Contains(ctlr, "systemd") {
				return cg
			}
		}
	}
```

nit: No need for the else.

pkg/metadata/process.go
```diff
@@ -25,28 +27,62 @@ func Process() Provider {
...
+
+	for _, cg := range cgroups {
+		for _, ctlr := range cg.Controllers {
+			if ctlr == "cpu" {
+				return cg
+			}
+		}
+
+		if strings.Contains(cg.Path, "systemd") {
```
maxbrunet, line 76: Should we be looking for `/system.slice/` or `/user.slice/` prefixes here?

https://systemd.io/CGROUP_DELEGATION/#systemds-unit-types

pkg/metadata/process_test.go
```diff
@@ -0,0 +1,201 @@
...
+					Path:        "/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod1ff39434b35faeef64159d11e3f96024.slice/docker-09af509f3db677a2275723fc71bff3d9b6d19e4d404c44822f2262f700adcd4b.scope",
+				},
+				{
+					HierarchyID: 2,
+					Controllers: []string{"freezer"},
+					Path:        "/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod1ff39434b35faeef64159d11e3f96024.slice/docker-a.scope",
+				},
+				{
+					HierarchyID: 1,
+					Controllers: []string{"name=systemd"},
```
maxbrunet, line 122: Is this case real? It is not listed as a supported controller:
https://systemd.io/CGROUP_DELEGATION/#controller-support

maxbrunet (commented):
@javierhonduco a couple questions about the systemd-related conditions

Should the condition be "has `cpu` controller" (cgroup v1) or "has no controller" (cgroup v2)? (without any mention of systemd)

pkg/metadata/procfs.go
javierhonduco: The name sounds good to me! The tricky thing here is that a process can be in multiple CPU cgroups, but we return the first one, right? Perhaps we could add a comment about this so the expectation is clear

pkg/metadata/process.go
javierhonduco: Sounds good to me!

maxbrunet:
Made some changes. Still two questions unanswered

> Should the condition be "has cpu controller" (cgroup v1) or "has no controller" (cgroup v2)? (without any mention of systemd)

and/or

https://github.com/parca-dev/parca-agent/pull/1024#discussion_r1027333812

pkg/metadata/process_test.go
kakkoyun: cc @javierhonduco I assume you've added this example.

pkg/metadata/process.go
```diff
@@ -25,28 +27,68 @@ func Process() Provider {
...
+		// Find first systemd slice
+		// https://systemd.io/CGROUP_DELEGATION/#systemds-unit-types
+		if strings.HasPrefix(cg.Path, "/system.slice/") || strings.HasPrefix(cg.Path, "/user.slice/") {
+			return cg
+		}
+
+		// FIXME: what are we looking for here?
+		// https://systemd.io/CGROUP_DELEGATION/#controller-support
+		for _, ctlr := range cg.Controllers {
+			if strings.Contains(ctlr, "systemd") {
```
kakkoyun, line 87: Yes, this conflicts with the documentation. Is there a version difference? Maybe `unified` mode.
Apparently this is an edge case since this is the second time we check the controllers.

maxbrunet: Unified mode is cgroup v2, this is always empty in v2 

> For cgroups version 1 hierarchies, this field contains a comma-separated list of the controllers bound to the hierarchy.  For the cgroups version 2 hierarchy, this field is empty.

https://man7.org/linux/man-pages/man7/cgroups.7.html

kakkoyun (approved):

pkg/metadata/process_test.go
javierhonduco: Sorry, somehow missed this! I think this was added by Maxime, but might be wrong 😄 . Either way, I am not sure about how real this case is 😿

PR #1034, opened 2022-11-16
Introduce debug-process-names and experimental-enable-dwarf-unwinding

kakkoyun:

```
// Hidden debug flags (only for debugging):
--debug-process-names []string 
Only attach profilers to specified processes. comm name will be used to match the given matchers. Accepts Go regex syntax (https://pkg.go.dev/regexp/syntax).

// These flags are experimental. Use them at your own peril.
--experimental-enable-dwarf-unwinding bool 
Unwind stack using .eh_frame information.
```
## Test

1. `sudo ./scripts/local-run.sh` with **`testdata/out/basic-cpp-no-fp`**:

<details>

```shell
sudo "${PARCA_AGENT}" \
            --node=local-test \
            --http-address=:7071 \
            --log-level=debug \
            --experimental-enable-dwarf-unwinding \
            --debug-process-names=basic-cpp-no-fp \
            --remote-store-address=localhost:7070 \
            --remote-store-insecure
```
CleanShot 2022-11-17 at 10 54 19@2x

</details>

2.  `sudo ./scripts/local-run.sh` with **CRuby interpreter**:

<details>

```shell
sudo "${PARCA_AGENT}" \
            --node=local-test \
            --http-address=:7071 \
            --log-level=debug \
            --experimental-enable-dwarf-unwinding \
            --debug-process-names=ruby \
            --remote-store-address=localhost:7070 \
            --remote-store-insecure
```

CleanShot 2022-11-17 at 14 59 16@2x

CleanShot 2022-11-17 at 14 59 02@2x

</details>

3. `make dev/up` with Minikube

<details>

CleanShot 2022-11-16 at 16 48 16

CleanShot 2022-11-16 at 16 51 04@2x

</details>

### DIFF
--- a/bpf/cpu/cpu.bpf.c
+++ b/bpf/cpu/cpu.bpf.c
@@ -162,6 +162,8 @@ u32 UNWIND_SAMPLES_COUNT = 7;
 
 /*================================ MAPS =====================================*/
 
+BPF_HASH(configuration, u8, u8, 1);
+BPF_HASH(debug_pids, int, u8, 32);
 BPF_HASH(stack_counts, stack_count_key_t, u64, MAX_STACK_COUNTS_ENTRIES);
 BPF_STACK_TRACE(stack_traces, MAX_STACK_TRACES);
 BPF_HASH(dwarf_stack_traces, int, stack_trace_t, MAX_STACK_TRACES);
@@ -342,6 +344,23 @@ static __always_inline void show_row(stack_unwind_table_t *unwind_table,
 static __always_inline bool has_unwind_information(pid_t pid) {
   stack_unwind_table_t *shard1 = bpf_map_lookup_elem(&unwind_table_1, &pid);
   if (shard1) {
+    return true;
+  }
+  return false;
+}
+
+static __always_inline bool is_debug_mode_enabled() {
+  u64 zero = 0;
+  void *val = bpf_map_lookup_elem(&configuration, &zero);
+  if (val) {
+    return true;
+  }
+  return false;
+}
+
+static __always_inline bool is_debug_enabled_for_pid(int pid) {
+  void *val = bpf_map_lookup_elem(&debug_pids, &pid);
+  if (val) {
     return true;
   }
   return false;
@@ -668,6 +687,10 @@ int profile_cpu(struct bpf_perf_event_data *ctx) {
 
   if (user_pid == 0)
     return 0;
+
+  if (is_debug_mode_enabled())
+    if (!is_debug_enabled_for_pid(user_tgid))
+      return 0;
 
   bool has_unwind_info = has_unwind_information(user_pid);
   // Check if the process is eligible for the unwind table or frame pointer
--- a/bpf/cpu/cpu.bpf.c
+++ b/bpf/cpu/cpu.bpf.c
@@ -60,6 +60,12 @@ enum stack_walking_method {
   STACK_WALKING_METHOD_DWARF = 1,
 };
 
+struct config_t {
+  bool debug;
+};
+
+const volatile struct config_t config = {};
+
 /*============================== MACROS =====================================*/
 
 #define BPF_MAP(_name, _type, _key_type, _value_type, _max_entries)            \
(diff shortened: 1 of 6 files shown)


pkg/profiler/cpu/cpu.go
```diff
@@ -263,54 +271,40 @@ func (p *CPU) Run(ctx context.Context) error {
...
+			return fmt.Errorf("failed to create procfs: %w", err)
+		}
 
-	dwarfStackTraces, err := m.GetMap(dwarfStackTracesMapName)
-	if err != nil {
-		return fmt.Errorf("get dwarf stack traces map: %w", err)
-	}
+		var matchers []*regexp.Regexp
+		for _, exp := range p.debugProcessNames {
+			regex, err := regexp.Compile(exp)
```
brancz, line 288: compiling a regex is rather expensive, we should do this once only

cmd/parca-agent/main.go
```diff
@@ -108,8 +108,11 @@ type flags struct {
 	DebuginfoStrip               bool          `kong:"help='Only upload information needed for symbolization. If false the exact binary the agent sees will be uploaded unmodified.',default='true'"`
 	DebuginfoUploadCacheDuration time.Duration `kong:"help='The duration to cache debuginfo upload exists checks for.',default='5m'"`
 
+	// Hidden debug flags (only for debugging):
+	DebugProcessNames []string `kong:"help='Only attach profilers to specified processes. comm name will be used to match the given matchers. Accepts Go regex syntax (https://pkg.go.dev/regexp/syntax).',hidden=''"`
```
brancz, line 112: Do we really need this to be a list? A single regex would be sufficient, no? If people really want to they can just use "or" in the regex.

brancz (commented):
I personally think a set of `comm`s would have been sufficient, but I'm happy with the regex as well. Only the compiling comment is something I think we need to address the rest I don't feel strongly about.

brancz:
Since it's all experimental and we can change flags arbitrarily especially with experimental features, let's just make this a single flag I would say.

pkg/profiler/cpu/cpu.go
kakkoyun: This actually happens once. It's outside the profiling loop and the watch loop.

cmd/parca-agent/main.go
kakkoyun: This is true. I also wanted to give the option to match against the exact names without thinking of the regex syntax. e.g. `etcd, kube-apiserver`

kakkoyun:
> Since it's all experimental and we can change flags arbitrarily especially with experimental features, let's just make this a single flag I would say.

One of the reasons, I want to have 2 flags, is to be able to use `debug-process-names` solely. It has many use cases while debugging things with the agent. Especially, after enabling system-wide profiling the information we collect from even a single node system can be overwhelming while trying to debug some specific thing.

bpf/cpu/cpu.bpf.c
```diff
@@ -162,6 +162,8 @@ u32 UNWIND_SAMPLES_COUNT = 7;
 
 /*================================ MAPS =====================================*/
 
+BPF_HASH(configuration, u8, u8, 1);
```
kakkoyun, line 165: @javierhonduco Do you know a more space-efficient way for these types of maps? I don't need the values; I want to check the existence of the keys.

brancz (approved):

javierhonduco:
> I personally think a set of comms would have been sufficient, but I'm happy with the regex as well. Only the compiling comment is something I think we need to address the rest I don't feel strongly about.

Having a regex is very useful for some applications that use named threads, they typically name them as `io-thread-$IO_THREAD_COUNT` :D

javierhonduco:
Oh, the only other thing I wanted to ask is to test if everything works for dwarf based stack unwinding. Could you run `cpp-basic-no-fp` and some simple Ruby script to check that everything's working fine? 

The output should not be too different from https://github.com/parca-dev/parca-agent/pull/978's test plan

kakkoyun:
> Oh, the only other thing I wanted to ask is to test if everything works for dwarf based stack unwinding. Could you run `cpp-basic-no-fp` and some simple Ruby script to check that everything's working fine?
> 
> The output should not be too different from #978 test plan

Updated the issue with the test results.

javierhonduco (approved):
LGTM!! 🎉

PR #1040, opened 2022-11-17
debuginfo: Make sure uploads do not block

kakkoyun:

Supersedes https://github.com/parca-dev/parca-agent/pull/1038

### DIFF
--- a/pkg/debuginfo/manager.go
+++ b/pkg/debuginfo/manager.go
@@ -54,6 +54,8 @@ type Manager struct {
 	*Extractor
 	*Uploader
 	*Finder
+
+	uploadTimeoutDuration time.Duration
 }
 
 // New creates a new Manager.
@@ -61,6 +63,7 @@ func New(
 	logger log.Logger,
 	reg prometheus.Registerer,
 	client Client,
+	uploadTimeout time.Duration,
 	cacheTTL time.Duration,
 	debugDirs []string,
 	stripDebuginfos bool,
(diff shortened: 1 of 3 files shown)


pkg/debuginfo/manager.go
```diff
@@ -123,25 +128,30 @@ func (di *Manager) removeAsUploading(buildID string) {
...
-			g.Go(func() error {
-				err := di.ensureUploaded(ctx, objFile)
-				if err != nil {
-					level.Error(logger).Log("msg", "failed to ensure debuginfo is uploaded", "err", err)
-				}
-				return nil
-			})
+	go func() {
+		_, err, _ := di.sfg.Do("ensure-uploaded", func() (interface{}, error) {
+			g := errgroup.Group{} // errgroup.WithContext doesn't work for this use case, we want to continue uploading even if one fails.
```
javierhonduco, line 133: Should we use [`WithContext`](https://cs.opensource.google/go/x/sync/+/refs/tags/v0.1.0:errgroup/errgroup.go;l=45;bpv=1;bpt=1)?

javierhonduco (approved):

pkg/debuginfo/manager.go
kakkoyun: I have added the comment for this question 😅

kakkoyun: But if the context gets canceled from the caller, we already propagate it down, so the grpc calls will be canceled. And group will exit.

javierhonduco: > I have added the comment for this question 😅

Oops, my bad, didn't see it

Issue #390, opened 2022-05-05
profiler: Cache seen stack trace ids in user-space


kakkoyun:
After we read stack ids from the BPF maps in each profile loop iteration, we clean the whole map. This might have unwanted consequences, such as stack id misses even though we have seen that stack id in previous iterations.

To ensure the soundness of stack resolution, we should record the seen ids for the session and check ids against it if it's not found in the BPF map.

kakkoyun:
> Just checking if this is still something you think it's worth doing! 😄

We don't have any data on this. Happy to close it until it becomes and actual issue.

Issue #1037, opened 2022-11-17
performance: ~15% of the CPU cycles are spent in the dbus/systemd packages


javierhonduco:
https://github.com/parca-dev/parca-agent/pull/971 introduced a library to query systemd, rather than relying on `systemctl`. While perhaps before we were using the same CPU as now (unsure), this puts more pressure on the GC, as well as spends more CPU cycles that perhaps before were spent waiting for the process to report back

![image](https://user-images.githubusercontent.com/959128/202477014-7773f199-a254-4aa6-929f-ba97488c459a.png)

https://pprof.me/8b20774

Seems like the library could be optimized a bit to reduce allocations, among others

cc @kakkoyun @maxbrunet

maxbrunet:
It must have been 0 CPU usage before, the cli was not available in the container. It is polling once every seconds, the frequency could be reduced:

https://github.com/parca-dev/parca-agent/blob/23cfe1e1406cea56c382558bfa68e7f922fd3783/pkg/discovery/systemd.go#L63

marselester:
I tried to reduce memory allocations in the decode function https://github.com/godbus/dbus/pull/344.

kakkoyun:
> I tried to reduce memory allocations in the decode function [godbus/dbus#344](https://github.com/godbus/dbus/pull/344).

Let's check the numbers. What's the effect on Agent?

marselester:
There is one more change I would like to make before checking the numbers...

marselester:
Here are the memory allocs profiles of the Agent with new changes vs main branch (each time parca-agent ran for 5 minutes).

## new changes

```sh
$ sudo ./dist/parca-agent
$ sleep 5m && curl --output agent_debug_dbus.allocs localhost:7071/debug/pprof/allocs
$ go tool pprof -source_path=/home/vagrant/go/pkg/mod/github.com/marselester/dbus/v5@v5.0.0-2022120720221210003314-8f55003f0c52/ -trim_path=github.com/godbus/dbus/v5@v5.1.0/ agent_debug_dbus.allocs
```

The dbus changes were pulled with 
`replace github.com/godbus/dbus/v5 v5.1.0 => github.com/marselester/dbus/v5 v5.0.0-20221210003314-8f55003f0c52`.

<details>

<summary>top</summary>

```
(pprof) top
Showing nodes accounting for 69.20MB, 68.51% of 101MB total
Dropped 71 nodes (cum <= 0.51MB)
[...]

PR #1041, opened 2022-11-17
profiler: Cache seen stack trace ids in user-space

marselester:
@kakkoyun did you mean something like this when describing stack id misses https://github.com/parca-dev/parca-agent/issues/390? Note, there is no thought put into concurrency or cache invalidation yet.

### DIFF
--- a/bpf/cpu/cpu.bpf.c
+++ b/bpf/cpu/cpu.bpf.c
@@ -108,6 +108,8 @@ typedef struct stack_count_key {
   int user_stack_id;
   int kernel_stack_id;
   int user_stack_id_dwarf;
+  int user_stack_id_fp;
+  int kernel_stack_id_fp;
 } stack_count_key_t;
 
 typedef struct unwind_tables_key {
@@ -428,13 +430,23 @@ static __always_inline void add_stacks(struct bpf_perf_event_data *ctx,
   int kernel_stack_id = bpf_get_stackid(ctx, &stack_traces, 0);
   if (kernel_stack_id >= 0) {
     stack_key.kernel_stack_id = kernel_stack_id;
+
+    stack_trace_type *kernel_stack = bpf_map_lookup_elem(&stack_traces, &kernel_stack_id);
+    if (kernel_stack == NULL) {
+      bpf_printk("kernel_stack is NULL, should not happen");
+    }
+    int stack_hash =
+        MurmurHash2((u32 *)kernel_stack,
+                    MAX_STACK_DEPTH * sizeof(u64) / sizeof(u32), 0);
+    bpf_printk("kernel stack hash %d", stack_hash);
+    stack_key.kernel_stack_id_fp = stack_hash;
   }
 
   if (method == STACK_WALKING_METHOD_DWARF) {
     int stack_hash =
         MurmurHash2((u32 *)unwind_state->stack.addresses,
                     MAX_STACK_DEPTH * sizeof(u64) / sizeof(u32), 0);
-    bpf_printk("stack hash %d", stack_hash);
+    bpf_printk("dwarf stack hash %d", stack_hash);
     stack_key.user_stack_id_dwarf = stack_hash;
     stack_key.user_stack_id = 0;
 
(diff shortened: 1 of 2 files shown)


kakkoyun:
> @kakkoyun did you mean something like this when describing stack id misses #390? Note, there is no thought put into concurrency or cache invalidation yet.

It's a good start. The idea is to eliminate the need to re-reading stack addresses from the eBPF map. We should still count them. 

We also need to consider what to do with the stack traces of eBPF maps. Shall we have another seen map in the kernel space to let the eBPF program know so that it won't bother to store the stack trace? 

So overall idea is that when we have stable hashes of the stack traces (which we should have), we don't need to keep syncing the stack traces themselves constantly. This could be a pre-mature optimization. We should make sure we got this right.

marselester:
That's a great idea!

So we can read from the `counts` BPF map on each iteration, check whether there is a stack trace cached in the user space using `stack hash` field:

- cache hit -- no need to read from the `stack_traces` BPF map
- cache miss -- read user/kernel space stack traces from the `stack_traces` BPF map

I assume that user and kernel space memory addresses should be combined to create `stack hash`.

```
stack hash = hash([ 0xdeadbeef; 0x123abcde; ...; 0x597be95a; 0xae5ee03; ... ]) = 'abcd'
```

The `counts` BPF map.

| { pid; userStackID; kernelStackID } | stack hash | seen
| ---                                 | ---        | ---
| { 10342; 1253; 0234 }               | abcd       | 45

The `stack_traces` BPF map.

| stack ID | memory addresses
| ---      | ---
| 1253     | [ 0xdeadbeef; 0x123abcde; ... ]
| 0234     | [ 0x597be95a; 0xae5ee03; ... ]

> Shall we have another seen map in the kernel space to let the eBPF program know so that it won't bother to store the stack trace?

Yeah, the `seen` BPF map can be added on the next optimization iteration 👍. Though I am not sure what strategy to use to invalidate the `seen` cache (BPF map) so it won't grow unbounded.

| stack hash | v
| ---        | ---
| abcd       | true

---

Perhaps the `stack_traces` BPF map should be indexed by `stack hash` so the `seen` BPF map won't be needed, also we might reduce a number of syscalls when reading the user and kernel space stack traces in one go.

marselester:
Hm, if there are multiple processes running the same program, perhaps there could be hash collisions. We could add PID to the hash 🤔.

```
stack hash = hash([ 10342; 0xdeadbeef; 0x123abcde; ...; 0x597be95a; 0xae5ee03; ... ]) = 'abcd'
```

marselester:
After reading `cpu.bpf.c` code I realized that there is already stack trace hashing used in DWARF stack walking `int user_stack_id_dwarf;`, so I added hashes for kernel space and user space (frame pointer method?).

```c
int user_stack_id_fp;
int kernel_stack_id_fp;
```

javierhonduco:
Hi @marselester! Thanks a lot for this!

Regarding the stack hashing, it's not necessary to add the frame pointer-based stack or kernel hashes, as the values returned by [`bpf_get_stackid`](https://man7.org/linux/man-pages/man7/bpf-helpers.7.html) is already a [hash](https://github.com/torvalds/linux/blob/3d7cb6b04c3f3115719235cc6866b10326de34cd/kernel/bpf/stackmap.c#L231)! 😄 

While I really appreciate the intent behind this PR, as definitely this project could be optimised to use fewer CPU cycles and memory, I think caching stack ids could result in difficult-to-debug issues, as well as it would increase the memory footprint of the agent, as the cache will use extra memory. 

I would be happy to add the extra complexity if this were a hotspot, but it seems like ~2.7% of the CPU cycles are spent here, which I am not sure if it would be worth it

https://pprof.me/de88b35

image

That being said, there are lots of areas we could improve in the agent when it comes to performance, and I would be happy to help you find something you might find interesting to work on.

For example:
- 5% of the CPU cycles are spent allocating memory in `compress/flate.NewWriter`
- We spent over 20% of the cycles searching in the GC, searching for object `runtime.scanobject`. This might mean that we make to many allocations that could be optimised, or that perhaps we could reduce the number of pointers we have
- We spent a disproportionate amount of cycles[ finding metadata for systemd units](https://github.com/parca-dev/parca-agent/issues/1037)

Hope this makes sense! Looking forward to seeing your contributions! ✨

marselester:
Thank you! Let me see if I can tackle the issues you've proposed...

> it's not necessary to add the frame pointer-based stack or kernel hashes

Silly me, I didn't know that :D

javierhonduco:
No worries at all! 😄

Issue #1031, opened 2022-11-16
Reduce BPF related memory allocations


brancz:
When trying to deploy parca-agent on low-memory nodes I get various memory related errors:

```
libbpf: Error in bpf_create_map_xattr(unwind_table_1):Operation not permitted(-1). Retrying without BTF.
libbpf: map 'unwind_table_1': failed to create: Operation not permitted(-1)
libbpf: permission error while running as root; try raising 'ulimit -l'? current value: 4096.0 MiB
libbpf: failed to load object 'parca'
level=error name=parca-agent ts=2022-11-16T08:38:09.3139736Z caller=main.go:153 err="load bpf object: failed to load BPF object: operation not permitted"
```

javierhonduco:
Just in case somebody bumps into this, I am working on a fix, need to do some more testing once I have the cycles. Will most likely be ready for review this evening or tomorrow morning 😄

brancz:
Closed by #1042

PR #1042, opened 2022-11-17
bpf: Reduce BPF map memory usage

javierhonduco:
Reduce BPF map memory usage by using one BPF map for the multiple shards.

When sharding was introduced, multiple BPF maps were added for each of the shards. This is very wasteful as the kernel must preallocate **and** lock the memory, so it can never be paged out.

The other issue is that machines with low memory count would not have enough memory to create all the maps.

Before this commit we were using 3 shards * 256 (MAX_PID_MAP_SIZE ) * 64B * 2 (size of each row) * 250k MAX_UNWIND_TABLE_SIZE + other metadata that's not too big ~= 24.6GB. This is obviously a ridiculous amount of memory.

This commit changes the sharded table scheme to just use one BPF map where they key is (pid, shard_id) with the same size, so it will roughly lock 8GB. This number is still rather large. I've reduced `MAX_PID_MAP_SIZE` to 100, which brings the locked memory now to 3.2GB.

The good news is that by using a single BPF map we can easily fine-tune the shard size to a more reasonable value in the future.

Memory usage should be reduced in the userspace code, too, as before we were naively adding all the shards to their buffer and then writting them. This commits reuses the created buffer and never holds the whole, BPF-friendly unwind table in memory.

## Testplan
image

No errors:
```
[javierhonduco@fedora parca-agent]$ sudo bpftool prog tracelog      | grep error
^C
[javierhonduco@fedora parca-agent]$
```

We hit the 2nd shard
```
[javierhonduco@fedora parca-agent]$ sudo bpftool prog tracelog      | grep -i shard | grep -v 'Shard 1'
      postmaster-736637  [004] d.h2.  5730.961535: bpf_trace_printk:     Shard 2
      postmaster-736637  [004] d.h2.  5730.961540: bpf_trace_printk:     Shard 2
      postmaster-736637  [010] d.h2.  5731.401629: bpf_trace_printk:     Shard 2
      postmaster-736637  [002] d.h2.  5732.581626: bpf_trace_printk:     Shard 2
      postmaster-736637  [002] d.h2.  5732.581631: bpf_trace_printk:     Shard 2
      postmaster-736637  [003] d.h2.  5735.131818: bpf_trace_printk:     Shard 2
      postmaster-736637  [003] d.h2.  5735.131823: bpf_trace_printk:     Shard 2
      postmaster-736637  [001] d.h2.  5736.231871: bpf_trace_printk:     Shard 2
      postmaster-736637  [001] d.h2.  5736.231876: bpf_trace_printk:     Shard 2
```

### DIFF
--- a/bpf/cpu/cpu.bpf.c
+++ b/bpf/cpu/cpu.bpf.c
@@ -29,9 +29,9 @@
 #define MAX_STACK_TRACES 1024
 // Number of items in the stack counts aggregation map.
 #define MAX_STACK_COUNTS_ENTRIES 10240
-// Size of the `<PID, unwind_table>` mapping. Determines how many
+// Size of the `<(pid, shard_id), unwind_table>` mapping. Determines how many
 // processes we can unwind.
-#define MAX_PID_MAP_SIZE 256
+#define MAX_PID_MAP_SIZE 100
 // Binary search iterations for dwarf based stack walking.
 // 2^20 can bisect ~1_048_576 entries.
 #define MAX_BINARY_SEARCH_DEPTH 20
@@ -104,6 +104,11 @@ typedef struct stack_count_key {
   int user_stack_id_dwarf;
 } stack_count_key_t;
 
+typedef struct unwind_tables_key {
+  int pid;
+  int shard;
+} unwind_tables_key_t;
+
 typedef struct unwind_state {
   u64 ip;
   u64 sp;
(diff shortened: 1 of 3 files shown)


javierhonduco:
Fixes https://github.com/parca-dev/parca-agent/issues/1031

brancz (approved):
Nice work! lgtm

v-thakkar (approved):
LGTM :rocket:

kakkoyun:
We should have merged this against the release branch 😅 Since it's a fix. We will merge the release branch when we tag something. I will cherry-pick.

brancz:
Does it really matter? We could also just merge things into main and merge everything back into the release branch. As long as it’s the latest release it doesn’t really matter does it?

PR #1046, opened 2022-11-18
Pin clang version

kakkoyun:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -102,6 +102,7 @@ jobs:
 
       - name: Test
         run: |
+          make clean bpf
           make test ENABLE_RACE=yes
 
       - name: Test unwind tables
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,6 +65,12 @@ jobs:
         with:
           version: "11"
 
+      - name: clang version
+        run:  |
+          clang -v
+          ld --version
+          ld.lld --version
+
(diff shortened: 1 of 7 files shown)


kakkoyun:
Snap failure on Ubuntu server so merging it.

javierhonduco (approved):

.github/workflows/build.yml
```diff
@@ -60,6 +60,11 @@ jobs:
           check-latest: true
           cache: true
 
+      - name: Set up Clang
```
javierhonduco, line 63: Should we mention here and below that this is installing not just clang but all the other related LLVM project tools?

Makefile
```diff
@@ -79,8 +79,8 @@ CGO_LDFLAGS_DYN = -L$(abspath $(LIBBPF_DIR)) -fuse-ld=ld -lelf -lz -lbpf
 
 # libbpf build flags:
 # CFLAGS = -g -O2 -Wall -fpie
-CFLAGS ?= -g -O2 -Werror -Wall -std=gnu89 # default CFLAGS
-LDFLAGS ?= -fuse-ld=lld
+CFLAGS ?= -g -O2 -Werror -Wall -std=gnu89 -fpic # default CFLAGS
```
kakkoyun, line 82: This was the fix. I wonder why I had removed it from the flags, even if it was in the default flags. 
fPIC and fPIE are practically the same things.

Issue #145, opened 2021-11-29
Elixir/Erlang VM support


brancz:
The Erlang VM has support for perf maps via the `ERL_FLAGS="+S 1 +JPperf true"` flags. However, even when setting those flags, profiling an Erlang process does not consistently work (occasionally individual addresses are symbolized).

Working theory: Erlang has a multi-process model, which could be a problem in this case if a process is short-lived (in the sense that a profiling loop of 10 seconds passed while the process was created and ended).

Ultimately even if the perf-map support works, it would be great for erlang users not to have to change anything about their deployment to reap the benefits, but it's a good intermediate step.

mkuratczyk:
It took me some time to get to this but it seems like Erlang support works just fine now. :)

All that's needed is `+JPperf map` in Erlang flags. [Documentation](https://www.erlang.org/doc/apps/erts/beamasm#linux-perf-support) suggests setting it to `true`, which also works. The difference is that when set to `true`, Erlang generates the mappings in two formats I mentioned above - setting it to `map` generates only one format and is sufficient.

I still see some unresolved symbols here and there but, these probably come from some libraries in our Docker image. A vast majority of symbols on this screenshot come from the JIT for sure:

![erlang-symbols](https://user-images.githubusercontent.com/9566114/165173884-04e3b5d9-4e89-4c61-985a-a4a146e33798.png)

brancz:
That is amazing!!!

tsloughter:
Hi, I'm the OpenTelemetry Erlang/Elixir maintainer and I'm interested in toying with this example and parca in sort of preparation for what the Otel Profiling SIG comes up with.

It isn't clear from the example how to actually use this with parca. I've played with using parca UI, have it installed locally via snap, and I've tried `perf` with Erlang, but how this demo and the parca agent are meant to work together I'm not seeing :)

kakkoyun:
Hey @tsloughter, maybe this repo and examples could help better https://github.com/parca-dev/parca-demo/tree/main/erlang

The idea behind the agent is to profile the whole system using eBPF. You need to drop the agent on the host and it will do the rest. If you configure it to send data to a Parca server you can see the collected profiles (you can see all this in the example above). For now, you need to set some flags for BEAM so that the agent can symbolize the collected stack traces.

I hope this helps. Let me know if it's not clear enough.

tsloughter:
@kakkoyun oh, you mean run the agent locally on my machine which will collect from the Erlang process running in docker?

[...]

PR #1051, opened 2022-11-18
feat(perf): support reading symbols from jitdump

maxbrunet:
## Motivation

The agent already supports reading symbols from [perf map files](https://github.com/torvalds/linux/blob/master/tools/perf/Documentation/jit-interface.txt), and there is a second format called `jitdump`:

https://github.com/torvalds/linux/blob/master/tools/perf/Documentation/jitdump-specification.txt

Some languages with `jitdump` support:

* Elixir/Erlang (#145)
* JVM via perf's [`libperf_jvmti.so` agent library](https://github.com/torvalds/linux/tree/master/tools/perf/jvmti) (#1)
* Node.js, Chrome V8 Javascript
* Wasmtime: https://docs.wasmtime.dev/examples-profiling-perf.html

See also https://developer.arm.com/documentation/101816/0800/JIT-profiling-support/JIT-profiling

## Changes

* Add a parser for the JITDUMP format
	* Everything is parsed, expect the EH Frames (if any) for now 
	* ~~It is currently optimistic and warns if a field cannot be read, hoping the rest can be (happy to discuss if it is good or bad)~~
	* They maybe some refactoring possible regarding how reads are done, I expect a lot of feedback :grinning:
* Add JITDUMP as a fallback to Perf map, looking for the file path in `/proc/<pid>/maps` (the file is `mmap()`'d to be used as marker)

Currently only JIT Code Loads records are used to mimic a Perf map, but the format has potential for more, it could be used to build Debug Info and upload to Parca, but the problem is JIT Code does not have a build ID. Perf Maps keep priority over jitdumps as they are more lightweight, but if we find advantages in the richer format, it might change in the future. (Otherwise, I might reconfigure the parser to discard unnecessary data as an optimization)

## Testing

I ~~have in mind to add~~ added jitdumps and their JSON snapshot to https://github.com/parca-dev/testdata

Right now, you can:

* Run NodeJS with the `--perf-prof` flag
* Run Erlang with the `ERL_FLAGS='+JPperf dump'` environment variable
* Run Wasmtime with the `--jitdump` flag


maxbrunet (commented):
I will not work on this until I get a 1st round of feedback.

Some comments from a self-review:

* Define static errors, so they can easily be caught by caller

pkg/perf/perf.go
```diff
@@ -139,20 +167,45 @@ func (p *cache) MapForPID(pid int) (*Map, error) {
...
-	perfFile := fmt.Sprintf("/proc/%d/root/tmp/perf-%d.map", pid, nsPid)
-	h, err := hash.File(p.fs, perfFile)
-	if err != nil {
-		if os.IsNotExist(err) {
-			return nil, ErrPerfMapNotFound
+	fileFmts := []string{
+		"/proc/%d/root/tmp/perf-%d.map",
+		"/proc/%d/root/tmp/jit-%d.dump",
+		"/proc/%d/cwd/jit-%d.dump",
+	}
```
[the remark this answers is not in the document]
kakkoyun: 💯

pkg/perf/perf.go
```diff
@@ -139,20 +167,45 @@ func (p *cache) MapForPID(pid int) (*Map, error) {
...
 
 	if p.pidMapHash[pid] == h {
 		return p.cache[pid], nil
 	}
 
-	m, err := ReadMap(p.fs, perfFile)
+	var m Map
+	var err error
+	switch {
+	case strings.HasSuffix(perfFile, ".map"):
```
kakkoyun, line 201: How about converting the `Map` to an interface type and have 2 different implementations for `.map` and `.dump`?

kakkoyun: `Lookup(addr uint64) (string, error)` can be the only method.

pkg/jit/jitdump.go
```diff
@@ -0,0 +1,386 @@
...
+	Header        *JITHeader             // the jitdump file header
+	CodeLoads     []*JRCodeLoad          // JITCodeLoad records
+	CodeMoves     []*JRCodeMove          // JITCodeMove records
+	DebugInfo     []*JRCodeDebugInfo     // JITCodeDebugInfo records
+	UnwindingInfo []*JRCodeUnwindingInfo // JITCodeUnwindingInfo records
+}
+
+// jitDumpParser is used to parse a jitdump file.
+type jitDumpParser struct {
+	logger     log.Logger
```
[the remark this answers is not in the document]
kakkoyun: > And the more I think about it, the more I am of the opinion, the parser should fail if it cannot parse a record (all or nothing). Records of unknown type are discarded, so if the specification is extended, it should keep working

👍 I think it is a good idea. 

> This package would be more portable if it does not depend on `go-kit/log`.

I don't mind depending on `go-kit/log`. We can think about it when people want to use it as a separate package.

pkg/jit/jitdump.go
```diff
@@ -0,0 +1,386 @@
...
+	}
+
+	if header.Version > JITHeaderVersion {
+		return nil, fmt.Errorf("wrong jitdump version: %d (expected: %d)", header.Version, JITHeaderVersion)
+	}
+
+	return header, nil
+}
+
+func (p *jitDumpParser) parseJRCodeLoad(prefix *JRPrefix) *JRCodeLoad {
```
kakkoyun, line 187: These helper functions should return errors with context, and the caller can log.

kakkoyun: We can use something like https://pkg.go.dev/go.uber.org/multierr

pkg/jit/jitdump.go
```diff
@@ -0,0 +1,386 @@
...
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+)
+
+// JITHeader represent a jitdump file header.
+type JITHeader struct {
```
kakkoyun, line 37: nit: We should consider which types and fields we export if we don't depend on reflection or serialization.

kakkoyun (commented):
This is great. We can improve some code structure and start testing it.
Thanks for realizing and enhancing , these parts.

pkg/perf/perf.go
maxbrunet: Sounds good, the idea went through my mind, I wired things up quickly to get it to work

pkg/jit/jitdump.go
maxbrunet: I wonder if it is possible to do with `fmt.Errorf()`, I've never tried, so I will try

pkg/jit/jitdump.go
maxbrunet: Agreed, but I am not sure how the package will be used yet, so I do not have an opinion on that, the ratio useful/expensive needs to be taken into account too

pkg/jit/jitdump.go
kakkoyun: Wrapping is possible but combining multiple errors is not. We use similar packages elsewhere.

pkg/perf/perf.go
maxbrunet: Actually, the interface feels like over-engineering too me, `MapFromDump()` would still be needed to compute end addresses and sort them to allow for binary search. And we would still have a switch-case to instantiate one or the other

kakkoyun: I don't think it's over-engineering. It will make this cleaner and easier to read. We write for humans to read and computers to run 😄

kakkoyun (commented):
I'm happy to merge this. But I see several to do items. Do you plan to handle them in this PR or shall we merge this? @maxbrunet

pkg/perf/perf.go
maxbrunet: hmm my thought exactly, simplicity is good for humans, the pattern would feel a bit forced to me :slightly_smiling_face:

Anyway I'd like to keep changes to the minimum in this file, we can revisit in another PR, I will only fixed the performance issues I introduced

maxbrunet:
> I'm happy to merge this. But I see several to do items. Do you plan to handle them in this PR or shall we merge this?

Yes, I'd like to handle them here, but before I do so, I would appreciate some feedback from @v-thakkar and/or @javierhonduco since self-requested a review :pray:

maxbrunet:
@kakkoyun Ready to merge! :bow: I'll do further work in the other PRs. :slightly_smiling_face:

pkg/perf/perf.go
```diff
@@ -55,6 +58,12 @@ var (
 	ErrNoSymbolFound      = errors.New("no symbol found")
 )
 
+var perfFileFmts = []string{
+	"/proc/%d/root/tmp/perf-%d.map",
+	"/proc/%d/root/tmp/jit-%d.dump",
+	"/proc/%d/cwd/jit-%d.dump",
```
vchuravy, line 64: LLVM supports an environment variable `JITDUMPDIR`

https://github.com/llvm/llvm-project/blob/bb666c69300c962c6a259c9d8e504601e557bc30/llvm/lib/ExecutionEngine/PerfJITEvents/PerfJITEventListener.cpp#L304

and otherwise will be `.debug/jit/llvm-IR-jit-20201109-09a2b8/jit-2397822.dump` in either `$HOME` or `cwd`.

vchuravy: `perf inject --jit` get's the filename from the corresponding perf event in the main file.

maxbrunet: Correct (https://github.com/parca-dev/parca-demo/pull/24#issuecomment-1331627461), I will add tackle that in the next PR

> `perf inject --jit` get's the filename from the corresponding perf event in the main file.

Do you know the details about how the filename ends up in the `perf.data` file? I had in mind to list all files under `.debug/jit/` ordered by mtime

vchuravy: From what I can tell you would need to parse the perf events and look for the MMAP2 and MMAP records (see tools/perf/perf-inject.c) the events are described here https://github.com/torvalds/linux/blob/97ee9d1c16963375eefdf964c429897d27e28956/include/uapi/linux/perf_event.h#L831

And then you can check that it ends in `jit-$PID.dump`

maxbrunet: Thank you, I think I see how it works on a high level, the agent does not record these events at this time and the target process can start before the agent, so it does not seem we will be able to use that

vchuravy: Actually it looks like you could simple scan /proc/.../maps according to https://github.com/llvm/llvm-project/blob/bb666c69300c962c6a259c9d8e504601e557bc30/llvm/lib/ExecutionEngine/PerfJITEvents/PerfJITEventListener.cpp#L341

vchuravy:
```
➜  ~ cat /proc/34437/maps | grep -G 'jit-[[:digit:]]*.dump' -
7fe2b5123000-7fe2b5124000 r-xp 00000000 103:03 21235307                  /home/vchuravy/.debug/jit/llvm-IR-jit-20221203-36c97d/jit-34437.dump
```

maxbrunet: Nice! Thank you so much for the research!

pkg/perf/perf.go
```diff
@@ -116,7 +116,7 @@ func MapFromDump(logger log.Logger, fs fs.FS, fileName string) (Map, error) {
 		}
 		// Some runtimes update their dump all the time (e.g. libperf_jvmti.so),
```
vchuravy, line 117: Yeah this will be the norm in case you attach to a running process like parca will do.

kakkoyun:
I want to proceed with this. Vaishali is on PTO. @javierhonduco could you take a look at this? Is there a blocker on your side?

javierhonduco:
Haven't had the time, yet. I will make sure I review it by end of the week.

maxbrunet:
I added tests, here is a first benchmark:

```console
$ ls -l testdata/jitdump/*.dump
-rw-r--r-- 1 maxime users   506419 Dec  9 12:11 testdata/jitdump/dotnet.dump
-rw-r--r-- 1 maxime users 13040830 Dec  9 12:11 testdata/jitdump/erlang.dump
-rw-r--r-- 1 maxime users    10582 Dec  9 12:11 testdata/jitdump/julia.dump
-rw-r--r-- 1 maxime users 13082624 Dec  9 12:11 testdata/jitdump/libperf-jvmti.dump
-rw-r--r-- 1 maxime users  1562469 Dec  9 12:11 testdata/jitdump/nodejs.dump
-rw-r--r-- 1 maxime users     6516 Dec  9 12:11 testdata/jitdump/php.dump
-rw-r--r-- 1 maxime users   129246 Dec  9 12:11 testdata/jitdump/wasmtime.dump

$ go test -bench=. -benchmem ./pkg/jit
goos: linux
goarch: amd64
pkg: github.com/parca-dev/parca-agent/pkg/jit
cpu: 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
BenchmarkLoadJITDump/dotnet-8               1468            969714 ns/op          692432 B/op      14414 allocs/op
BenchmarkLoadJITDump/erlang-8                 34          34686926 ns/op        22501897 B/op     835150 allocs/op
BenchmarkLoadJITDump/julia-8               21128             49061 ns/op           26057 B/op       1147 allocs/op
BenchmarkLoadJITDump/libperf-jvmti-8          39          27826079 ns/op        20026386 B/op     695169 allocs/op
BenchmarkLoadJITDump/nodejs-8                448           2826668 ns/op         2078581 B/op      39746 allocs/op
BenchmarkLoadJITDump/php-8                 38241             28755 ns/op           17012 B/op        538 allocs/op
BenchmarkLoadJITDump/wasmtime-8             7779            217346 ns/op          174055 B/op       3231 allocs/op
PASS
ok      github.com/parca-dev/parca-agent/pkg/jit        11.512s
```

vchuravy:
Would love to see this merged!

pkg/jit/jitdump.go
```diff
@@ -0,0 +1,460 @@
...
+// JITRecordType is the value identifying the record type.
+type JITRecordType uint32
+
+const (
+	JITCodeLoad          JITRecordType = 0    // record describing a jitted function
+	JITCodeMove                        = 1    // record describing an already jitted function which is moved
+	JITCodeDebugInfo                   = 2    // record describing the debug information for a jitted function
+	JITCodeClose                       = 3    // record marking the end of the jit runtime (optional)
+	JITCodeUnwindingInfo               = 4    // record describing a function unwinding information
+	JITCodeMax                         = iota // maximum record type
```
javierhonduco, line 66: Very nice! Good shout to handle this!

pkg/jit/jitdump.go
```diff
@@ -0,0 +1,460 @@
...
+	p.buf = bufio.NewReader(rd)
+
+	magic, err := p.buf.Peek(4)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read magic number from jitdump file: %w", err)
+	}
+
+	switch {
+	case bytes.Equal(magic, []byte{'J', 'i', 'T', 'D'}):
+		p.endianness = binary.BigEndian
```
javierhonduco, line 165: I am not sure if we should care about big-endian architectures as we don't really test them nor support them (https://github.com/parca-dev/parca-agent/pull/1272)

pkg/jit/jitdump.go
```diff
@@ -0,0 +1,460 @@
...
+	case bytes.Equal(magic, []byte{'D', 'T', 'i', 'J'}):
+		p.endianness = binary.LittleEndian
+	default:
+		return nil, fmt.Errorf("%w: %#x", err, magic)
```
javierhonduco, line 169: Perhaps you forgot to use `ErrWrongJITDumpMagic` here?

pkg/jit/jitdump.go
```diff
@@ -0,0 +1,460 @@
...
+func isUnexpectedIOError(err error) error {
+	if errors.Is(err, io.EOF) {
+		return io.ErrUnexpectedEOF
+	}
+	return err
+}
+
+// read reads structured binary data from the jitdump file into data.
+func (p *jitDumpParser) read(data any) error {
+	return binary.Read(p.buf, p.endianness, data)
```
javierhonduco, line 185: `binary.Read` (and Write) [allocate even on the fast path](https://github.com/golang/go/blob/3c7884a21b2929b5a42d6a1f237f5fee2ae72ba8/src/encoding/binary/binary.go#L232). While this might not seem like a problem, it will create a lot of short-lived allocations increase pressure on the GC even more.

We spend way too many cycles in the Garbage Collector already (~30%) and want to reduce this. I recently added even more `binary.Write`s and I am planning to remove them soon with more efficient low level bits, `binary.LittleEndian.Put*`.

Perhaps here we could use `binary.LittleEndian. Uint*`?

pkg/jit/jitdump.go
```diff
@@ -0,0 +1,460 @@
...
+
+		var err error
+		jr.Entries[i].Name, err = p.readString()
+		if err != nil {
+			return nil, fmt.Errorf("entries[%d].name: %w", i, err)
+		}
+		size += len(jr.Entries[i].Name) + 1 // +1 accounts for trimmed null termination
+	}
+
+	// Discard padding if any
```
javierhonduco, line 347: Awesome stuff!

pkg/jit/jitdump.go
```diff
@@ -0,0 +1,460 @@
...
+	var err error
+	dump.Header, err = p.parseJITHeader()
+	if err != nil {
+		return nil, fmt.Errorf("failed to read JIT dump header: %w", isUnexpectedIOError(err))
+	}
+
+	dump.CodeLoads = make([]*JRCodeLoad, 0)
+	dump.CodeMoves = make([]*JRCodeMove, 0)
+	dump.DebugInfo = make([]*JRCodeDebugInfo, 0)
+	dump.UnwindingInfo = make([]*JRCodeUnwindingInfo, 0)
```
javierhonduco, line 394: Should we accept these as arguments to move the allocations to the caller? cc @kakkoyun

javierhonduco (approved):
Looking great so far! ❤️  Did a first pass of reviews and things are very close to being ready. Pre-emptively approving but let's ensure that we get one more maintainer onboard 😄

javierhonduco:
Oh, another minor thing, now we have the latest `testdata` pulled in, but whenever we grep something in the repo (e.g. using ripgrep) it might match the binary files. Could you please add them to the exclusion list for binary / generated files? Cheers!

pkg/jit/jitdump.go
kakkoyun: Yes, we can get the whole `*JITDump` as an argument and leave the allocation to the caller. As a second iteration, we can consider using a pool if it creates a problem.

kakkoyun (commented):
This PR looks awesome. Let's try to reduce allocations. We can run the escape analyzer to measure. Let's merge this and iterate, I'd say.

pkg/jit/jitdump.go
maxbrunet: Indeed, excellent catch!

pkg/jit/jitdump.go
maxbrunet: Do you mean [`binary.ReadUvarint`](https://pkg.go.dev/encoding/binary#ReadUvarint)? Yes, that could be an optimization, but correct if I am wrong, it would be for `uint64` fields only

pkg/jit/jitdump.go
maxbrunet: Good suggestion!

pkg/jit/jitdump.go
maxbrunet: Maybe, just implementing the specification, it does not seem to cause any harm here

pkg/jit/jitdump.go
brancz: More like `p.endianness.Uint32|64(p.buf)` as what's currently the caller of this function knows exactly what type it wants to read. But the way `binary.Read` is written it will always perform allocations even when not necessary (and it's not necessary for reading integers).

maxbrunet:
> Could you please add them to the exclusion list for binary / generated files?

Sorry, what exclusion list are you referring to? I assume that would be in the `testdata` repo

pkg/jit/jitdump.go
maxbrunet: Ah I see the [`ByteOrder` interface](https://pkg.go.dev/encoding/binary#ByteOrder), it looks like it would still require to allocate a slice of bytes, they do not take an `io.Reader`

it is what `binary.Read` does: https://cs.opensource.google/go/go/+/refs/tags/go1.20:src/encoding/binary/binary.go;l=254

brancz: Sorry for clarification, it does, because the size cannot be predicted, and therefore is allocated on the heap. Whereas for uint64, a `make([]byte, 4)` is not allocated on the heap as the compiler knows exactly the size and can manage it on the stack.

maxbrunet: > Line 254 doesn't cause the allocation, line 232 does: [cs.opensource.google/go/go/+/refs/tags/go1.20:src/encoding/binary/binary.go;l=232](https://cs.opensource.google/go/go/+/refs/tags/go1.20:src/encoding/binary/binary.go;l=232)

Yes

---

Stack vs Heap: I see now, thank you for the explanation! I can easily observe the difference now:

```console
$ cat >main.go <<EOF
package main

func main() {
        n := 8
        _ = make([]byte, n)
        _ = make([]byte, 8)
}
EOF

$ go build -gcflags=-m main.go
# command-line-arguments
./main.go:3:6: can inline main
./main.go:5:10: make([]byte, n) escapes to heap
./main.go:6:10: make([]byte, 8) does not escape
```

maxbrunet:
New benchmark: -60% allocs! Thank you so much @javierhonduco @brancz for spotting this!

```
$ go test -bench=. -benchmem ./pkg/jit
goos: linux
goarch: amd64
pkg: github.com/parca-dev/parca-agent/pkg/jit
cpu: 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
BenchmarkLoadJITDump/dotnet-8               1426            752507 ns/op          630455 B/op       4453 allocs/op
BenchmarkLoadJITDump/erlang-8                 44          25866619 ns/op        19224474 B/op     295672 allocs/op
BenchmarkLoadJITDump/julia-8               26508             38855 ns/op           22103 B/op        445 allocs/op
BenchmarkLoadJITDump/libperf-jvmti-8          51          22224994 ns/op        17721051 B/op     272435 allocs/op
BenchmarkLoadJITDump/nodejs-8                504           2339107 ns/op         1906934 B/op      12791 allocs/op
BenchmarkLoadJITDump/php-8                 56421             20795 ns/op           14843 B/op        180 allocs/op
BenchmarkLoadJITDump/wasmtime-8             9472            148831 ns/op          160221 B/op       1010 allocs/op
PASS
ok      github.com/parca-dev/parca-agent/pkg/jit        10.304s
```

<details>
<summary>Benchstat before/after</summary>

```
goos: linux
goarch: amd64
pkg: github.com/parca-dev/parca-agent/pkg/jit
cpu: 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
                            │   old.txt    │               new.txt                │
                            │    sec/op    │    sec/op     vs base                │
LoadJITDump/dotnet-8          1005.9µ ± 6%   690.8µ ± 12%  -31.32% (p=0.000 n=10)
LoadJITDump/erlang-8           34.77m ± 1%   24.12m ±  1%  -30.63% (p=0.000 n=10)
LoadJITDump/julia-8            51.37µ ± 5%   36.74µ ±  3%  -28.48% (p=0.000 n=10)
LoadJITDump/libperf-jvmti-8    28.41m ± 1%   21.14m ±  2%  -25.62% (p=0.000 n=10)
LoadJITDump/nodejs-8           2.990m ± 7%   2.145m ±  9%  -28.26% (p=0.000 n=10)
LoadJITDump/php-8              27.45µ ± 2%   20.86µ ±  2%  -24.00% (p=0.000 n=10)
LoadJITDump/wasmtime-8         204.3µ ± 6%   149.3µ ±  3%  -26.90% (p=0.000 n=10)
geomean                        978.0µ        704.9µ        -27.93%

                            │   old.txt    │               new.txt                │
                            │     B/op     │     B/op      vs base                │
LoadJITDump/dotnet-8          676.2Ki ± 0%   615.7Ki ± 0%   -8.95% (p=0.000 n=10)
LoadJITDump/erlang-8          21.46Mi ± 0%   18.33Mi ± 0%  -14.57% (p=0.000 n=10)
LoadJITDump/julia-8           25.45Ki ± 0%   21.58Ki ± 0%  -15.18% (p=0.000 n=10)
LoadJITDump/libperf-jvmti-8   19.10Mi ± 0%   16.90Mi ± 0%  -11.51% (p=0.000 n=10)
LoadJITDump/nodejs-8          1.982Mi ± 0%   1.819Mi ± 0%   -8.26% (p=0.000 n=10)
LoadJITDump/php-8             16.61Ki ± 0%   14.50Ki ± 0%  -12.75% (p=0.000 n=10)
LoadJITDump/wasmtime-8        170.0Ki ± 0%   156.5Ki ± 0%   -7.95% (p=0.000 n=10)
geomean                       636.6Ki        564.4Ki       -11.35%

                            │   old.txt    │               new.txt               │
                            │  allocs/op   │  allocs/op   vs base                │
LoadJITDump/dotnet-8          14.414k ± 0%   4.453k ± 0%  -69.11% (p=0.000 n=10)
LoadJITDump/erlang-8           835.1k ± 0%   295.7k ± 0%  -64.60% (p=0.000 n=10)
LoadJITDump/julia-8            1147.0 ± 0%    445.0 ± 0%  -61.20% (p=0.000 n=10)
LoadJITDump/libperf-jvmti-8    695.2k ± 0%   272.4k ± 0%  -60.81% (p=0.000 n=10)
LoadJITDump/nodejs-8           39.75k ± 0%   12.79k ± 0%  -67.82% (p=0.000 n=10)
LoadJITDump/php-8               538.0 ± 0%    180.0 ± 0%  -66.54% (p=0.000 n=10)
LoadJITDump/wasmtime-8         3.231k ± 0%   1.010k ± 0%  -68.74% (p=0.000 n=10)
geomean                        18.21k        6.247k       -65.69%
```

</details>

javierhonduco:
Amazing, thanks so much!

kakkoyun (approved):
LGTM ❤️  Awesome work! 🤩

PR #1064, opened 2022-11-24
config: Make sure we handle errors when file path specified

kakkoyun:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,5 +22,7 @@ RUN chmod +x parca-agent
 
 # hadolint ignore=DL3029
 FROM --platform="${TARGETPLATFORM:-linux/amd64}" gcr.io/distroless/static@sha256:5759d194607e472ff80fff5833442d3991dd89b219c96552837a2c8f74058617
-COPY --chown=0:0 --from=builder /app/parca-agent /bin/parca-agent
-CMD ["/bin/parca-agent"]
+COPY --chown=0:0 --from=builder /app/parca-agent /parca-agent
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -22,5 +22,7 @@ RUN chmod +x parca-agent
 
 # hadolint ignore=DL3029
 FROM --platform="${TARGETPLATFORM:-linux/amd64}" gcr.io/distroless/static@sha256:5759d194607e472ff80fff5833442d3991dd89b219c96552837a2c8f74058617
-COPY --chown=0:0 --from=builder /app/parca-agent /bin/parca-agent
-CMD ["/bin/parca-agent"]
+COPY --chown=0:0 --from=builder /app/parca-agent /parca-agent
+COPY --chown=0:0 parca-agent.yaml /parca-agent.yaml
+
+CMD ["/parca-agent"]
```
maxbrunet, line 28: What about making the default proper locations?

```suggestion
COPY --chown=0:0 --from=builder /app/parca-agent /bin/parca-agent
COPY --chown=0:0 parca-agent.yaml /etc/parca-agent/parca-agent.yaml

ENTRYPOINT ["/bin/parca-agent"]
```

kakkoyun: This was my first version. Then I wanted to sync things with the Parca server.
We can have the proper locations and do the same in the Parca server.

kakkoyun: Let's handle this in a separate PR that targets the `main` branch. I'll create an issue for it.

Issue #1016, opened 2022-11-15
Renovate timeouts when updating Go dependencies


maxbrunet:
This blocks update PRs from being opened

```json
{
  "err": {
    "name": "ExecError",
    "cmd": "/bin/sh -c docker run --rm --name=renovate_go --label=renovate_child -v \"/mnt/renovate/gh/parca-dev/parca-agent\":\"/mnt/renovate/gh/parca-dev/parca-agent\" -v \"/tmp/renovate-cache\":\"/tmp/renovate-cache\" -v \"/tmp/containerbase\":\"/tmp/containerbase\" -e GOPATH -e GOPROXY -e GOFLAGS -e CGO_ENABLED -e GIT_CONFIG_KEY_0 -e GIT_CONFIG_VALUE_0 -e GIT_CONFIG_KEY_1 -e GIT_CONFIG_VALUE_1 -e GIT_CONFIG_KEY_2 -e GIT_CONFIG_VALUE_2 -e GIT_CONFIG_COUNT -e BUILDPACK_CACHE_DIR -e CONTAINERBASE_CACHE_DIR -w \"/mnt/renovate/gh/parca-dev/parca-agent\" docker.io/renovate/go:1.19.3 bash -l -c \"go get -d -t ./... && go mod tidy && go mod tidy\"",
[...]

PR #1066, opened 2022-11-25
Update Go dependencies

marselester:


maxbrunet:
Hey @marselester, FYI this is supposed to be automated, but we have an issue with our automation: #1016 

And we also do not want to bump the minimum Go version unless absolutely necessary. If Go 1.18 still works, we should keep it.

marselester:
Oh I see, sorry about that. I set the version back to 1.18.
It would be nice to have the most recent versions of the following packages (there could be performance improvements):

```
github.com/coreos/go-systemd/v22
github.com/klauspost/compress
github.com/godbus/dbus/v5
```

kakkoyun (approved):
LGTM.

We need to fix the automation on the long run.

Issue #1063, opened 2022-11-23
Set the size of the unwind maps dynamically


javierhonduco:
(related work before working on this issue: https://github.com/parca-dev/parca-agent/issues/1052)

BPF maps need to be locked in memory. This can be a problem as the unwind tables can be very large. However, we can perhaps explore the needed storage space before creating the maps and loading the programs and creating maps with the space we need. If things change, later on, we can re-create the maps with the right size (and possibly re-load the program, too)

There are a bunch of things to be thoroughly discussed and thought to ensure that memory usage is as low as possible while not degrading the user experience

cc/ @kakkoyun @v-thakkar @brancz

PR #1074, opened 2022-11-28
Conditionally set unwind tables map size

javierhonduco:
Right now, we set the unwind tables map size to 100 elements. This means that we havce to allocate and lock roughly 100 elements * 250k entries * 16 bytes per entry ~= 400 MB, just for this map.

We are currently working on some optimisations that will result in lower memory required for the unwind tables, but this will take some weeks to land, so in the meantime, this commit reduces the default unwind tables map size to 2, which would be around 8MB, and only bump it to a 100 elements if dwarf based unwinding is enabled.

We still want to allocate and lock some memory even if we know that dwarf unwinding is disabled, as in the future this would be transparently enabled and it's highly likely that at least few processes (or mappings, soon) will use the dwarf based unwind approach. This way we will at least ensure that we can already allocate enough for 500k unwind rows.

**Test Plan**

by default:
```
[javierhonduco@fedora parca-agent]$ sudo bpftool map | grep unwind_tables -A2
70: hash  name unwind_tables  flags 0x0
        key 8B  value 4000032B  max_entries 2  memlock 8003584B
        btf_id 254
```

with `--experimental-enable-dwarf-unwinding`
```
[javierhonduco@fedora parca-agent]$ sudo bpftool map | grep unwind_tables -A2
47: hash  name unwind_tables  flags 0x0
        key 8B  value 4000032B  max_entries 100  memlock 400007168B
        btf_id 230
```

### DIFF
--- a/bpf/cpu/cpu.bpf.c
+++ b/bpf/cpu/cpu.bpf.c
@@ -29,9 +29,6 @@
 #define MAX_STACK_TRACES 1024
 // Number of items in the stack counts aggregation map.
 #define MAX_STACK_COUNTS_ENTRIES 10240
-// Size of the `<(pid, shard_id), unwind_table>` mapping. Determines how many
-// processes we can unwind.
-#define MAX_PID_MAP_SIZE 100
 // Binary search iterations for dwarf based stack walking.
 // 2^20 can bisect ~1_048_576 entries.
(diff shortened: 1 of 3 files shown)


kakkoyun (approved):
LGTM ❤️

javierhonduco:
Remediation for https://github.com/parca-dev/parca-agent/issues/1063

PR #1075, opened 2022-11-28
cmd/parca-agent: Use structs to organize flag groups

metalmatze:
Some of these might even be moveable to Parca to be used in both projects.

Like `FlagsLogs` and how we configure HTTP should be in Parca and then used by the agent.
https://github.com/parca-dev/parca/pull/2193


kakkoyun (commented):
Looks great! I love the idea!

metalmatze:
Here's the diff for the `parca-agent --help`:
```diff
5a6,7
>       --log-format="logfmt"       Configure if structured logging as JSON or as
>                                   logfmt
42a45,61
>       --debuginfo-upstream-servers=https://debuginfod.elfutils.org,...
>                                   Upstream debuginfod servers. Defaults to
>                                   https://debuginfod.elfutils.org. It is an
>                                   ordered list of servers to try. Learn more at
>                                   https://sourceware.org/elfutils/Debuginfod.html
>       --debuginfo-http-request-timeout=5m
>                                   Timeout duration for HTTP request to upstream
>                                   debuginfod server. Defaults to 5m
>       --debuginfo-cache-dir="/tmp"
>                                   Path to directory where debuginfo is cached.
>       --debuginfo-upload-max-size=1000000000
>                                   Maximum size of debuginfo upload in bytes.
>       --debuginfo-upload-max-duration=15m
>                                   Maximum duration of debuginfo upload.
>       --debuginfo-uploads-signed-url
>                                   Whether to use signed URLs for debuginfo
>                                   uploads.
47,49d65
<       --debuginfo-temp-dir="/tmp"
<                                   The local directory path to store the interim
<                                   debuginfo files.
56,57d71
<       --debuginfo-upload-timeout-duration=2m
<                                   The timeout duration to cancel uplod requests.
```

cmd/parca-agent/main.go
```diff
@@ -356,11 +371,11 @@ func run(logger log.Logger, reg *prometheus.Registry, flags flags) error {
...
-				flags.DebuginfoUploadTimeoutDuration,
-				flags.DebuginfoUploadCacheDuration,
-				flags.DebuginfoDirectories,
-				flags.DebuginfoStrip,
-				flags.DebuginfoTempDir,
+				flags.Debuginfo.UploadTimeoutDuration,
+				flags.Debuginfo.UploadCacheDuration,
+				flags.Debuginfo.Directories,
+				flags.Debuginfo.Strip,
+				flags.Debuginfo.CacheDir,
```
kakkoyun, line 378:
```suggestion
				flags.Debuginfo.TempDir,
```
This is semantically more correct. We don't cache.

README.md
```diff
@@ -94,22 +97,26 @@ Flags:
                                   Interval between batch remote client writes.
                                   Leave this empty to use the default value of
                                   10s.
+      --debuginfo-cache-dir="/tmp"
+                                  Path to directory where debuginfo is cached.
+      --debuginfo-upload-max-size=1000000000
```
kakkoyun, line 102: This flag is not used in the agent.

README.md
```diff
@@ -94,22 +97,26 @@ Flags:
...
+                                  Maximum size of debuginfo upload in bytes.
+      --debuginfo-upload-max-duration=15m
+                                  Maximum duration of debuginfo upload.
+      --debuginfo-uploads-signed-url
```
kakkoyun, line 106: Ditto.

brancz:
I think putting the flags into structs is a great idea overall, but not a great abstraction, I would stay away from reusing them across projects (as demonstrated by the debuginfo flags).

metalmatze:
Alright. Going to revert the debuginfo ones. 
However, for things like Logs, HTTP, gRPC it still makes a lot of sense.

metalmatze:
Parca has been a dependency. Not really sure why it now fails with the `pattern packages/app/web/build: no matching files found`. Any ideas?

kakkoyun:
> Parca has been a dependency. Not really sure why it now fails with the `pattern packages/app/web/build: no matching files found`. Any ideas?

We might be failed to build the UI. But we don't have this pattern for the agent 🤔

brancz:
I think it's because Go compilation works in packages, and previously the packages that end up embedding the UI were not compiled, but now they are because of the new usage of the flag struct. I personally think we should just copy these structs, the reuse is just unnecessary coupling of the projects in my opinion.

kakkoyun:
@metalmatze Shall we revive this?

PR #1077, opened 2022-11-28
docs(README): remove pod-label-selector from README

morremeyer:
parca 0.10.0 does not have a `--pod-label-selector` flag anymore.


brancz (approved):
Thank you!

PR #1092, opened 2022-12-05
pkg/objectfile:remove cache size and expired duratioin limit

zdyj3170101136:
Because the object file cache use the mapping id as key, 
so it is no need to add cache limit and expired duration.


kakkoyun:
@zdyj3170101136 Sorry we haven't been fast enough to review this. We are planning to improve caching around object files. The idea is to keep them around as long as we need them and make sure we sync any changes to the files.

Issue #1052, opened 2022-11-21
Move away from monolithic unwind tables


javierhonduco:
(This work is in progress. Planned to start ~end of November)

### Context
At the moment, native stack unwinding with dwarf-based information creates a monolithic unwind table. This table contains all the unwind information for every executable section for which we found dwarf unwind information. 

This proved useful as it allowed us to prototype this unwinder and ensure that our approach was feasible. However, this approach has some drawbacks:

- **High memory and CPU usage in userspace**: the table is built by building the subtables for each mapped executable. We then sort by the program counter, as we want to binary search over it. While sorting could be avoided by ensuring the table is constructed in the order of the already sorted mapped executable sections, this still forces us to keep the whole table in memory at some point in time, for every process. Another issue is that we need to adjust the program counter of the unwind tables to account for the base address where the executables have been loaded;
- **Wasted memory in BPF maps**: almost every single application on the systems we profile will dynamically link libc and few other libraries. Having to load this data, even if it's from a cache in the future, and then add it to the final unwind table, is not ideal

### Proposed solution
Rather than building a single, large, monolithic table, keep an unwind table per mapped executable. This way we would be able to reuse unwind tables, significantly reducing the work required to produce the unwind tables. This is very important as we would like to enable this new feature by default without incurring a big performance hit.

This will require adding extra process information, such as the ranges for each mapped executable we care about, as well as the base address so we can adjust the program counters.

#### Issues with this approach
One of the potential problems with this approach is the waste of space in the unwind tables. Unwind tables live in a BPF map that's preallocated and locked in memory which doesn't allow for dynamically sized types. For this reason, we use arrays to represent this table and no matter if an unwind table has a single element it will use the whole space.

We should be careful when setting the default value for the unwind tables, as a value too large will result in "external fragmentation" and space wasted, while a value too small will result in having to check in too many shards.

PR #1101, opened 2022-12-07
.eh_frame: Don't allocate the full unwind table by default

javierhonduco:
As we are going to enable DWARF-based unwinding by default soon, which
means that the DWARF unwind information parser and the table generation
will be called significantly more times.

During prototyping, some shortcuts were taken, such as generating the
unwind table ahead of time. The unwind table is an array of unwind rows,
which for "medium-sized" binaries it can reach 200k entries. For the largest
binaries we've seen, the unwind tables were up to 2M entries big.

Generating and keeping these large slices in memory is not good for
performance, as we need to re-allocate them multiple times when
regrowing the backing memory buffer, as well as copying the data over, if
necessary.

While we could reasonably estimate the final size of the buffer using the
`.text` size as a proxy, this is not necessary for the majority of
cases, and we can process one unwind entry at a time.

This is what this commits implements. It refactors the table generation
code to change from returning said unwind table in a slice, to accepting
a callback function that can then process each element in the way it
finds it suitable.

Note that there are way more optimisations that we could do to speed up
the code but we'll first profile this code in production once we carry
out the monolithic unwind table work, as its architecture will reduce
the redundant work.

__Note__: The unwind generation table is still generating the full table ahead
of time. This would have to be changed to take advantage of this new API.

## Benchmarks

```
$ go test -v github.com/parca-dev/parca-agent/pkg/stack/unwind -bench=. -count=30
```

```
[javierhonduco@fedora parca-agent]$ benchstat /tmp/old /tmp/new
name                                  old time/op    new time/op    delta
ParsingLibcDwarfUnwindInformation-12    34.5ms ±43%    10.8ms ±32%  -68.60%  (p=0.000 n=30+30)

name                                  old alloc/op   new alloc/op   delta
ParsingLibcDwarfUnwindInformation-12    32.7MB ± 0%     5.6MB ± 0%  -82.80%  (p=0.000 n=30+29)

name                                  old allocs/op  new allocs/op  delta
ParsingLibcDwarfUnwindInformation-12      150k ± 0%       83k ± 0%  -44.76%  (p=0.000 n=30+30)
```

### DIFF
--- a/internal/dwarf/frame/table.go
+++ b/internal/dwarf/frame/table.go
@@ -42,7 +42,6 @@ type DWRule struct {
 // that we have unwind information for.
 type InstructionContext struct {
 	loc           uint64
-	address       uint64
 	CFA           DWRule
 	Regs          map[uint64]DWRule
 	initialRegs   map[uint64]DWRule
@@ -54,6 +53,38 @@ type InstructionContext struct {
 
 func (instructionContext *InstructionContext) Loc() uint64 {
 	return instructionContext.loc
+}
+
+type InstructionContextIterator struct {
+	ctx         *Context
+	lastReached bool
+	done        bool
+}
+
+func (ici *InstructionContextIterator) HasNext() bool {
+	return !ici.done
+}
+
+func (ici *InstructionContextIterator) Next() *InstructionContext {
+	for ici.ctx.buf.Len() > 0 {
+		lastPcBefore := ici.ctx.lastInsCtx.loc
+		executeDwarfInstruction(ici.ctx)
+		lastPcAfter := ici.ctx.lastInsCtx.loc
+		// We are at an instruction boundary when there's a program counter change.
+		if lastPcBefore != lastPcAfter {
+			return ici.ctx.lastInsCtx
+		}
+	}
+
+	// Account for the last instruction boundary.
+	if !ici.lastReached {
+		ici.lastReached = true
+		return ici.ctx.currInsCtx
+	}
+
+	// We are done iterating.
+	ici.done = true
+	return nil
 }
 
 // RowState is a stack where `DW_CFA_remember_state` pushes
@@ -86,8 +117,8 @@ func (stack *StateStack) pop() RowState {
 
 // Context represents a function.
 type Context struct {
-	// An entry for each object code instruction that we have unwind information for.
-	instructions    []InstructionContext
+	currInsCtx      *InstructionContext
+	lastInsCtx      *InstructionContext
 	rememberedState *StateStack
 	// The buffer where we store the dwarf unwind entries to be parsed for this function.
 	buf   *bytes.Buffer
@@ -95,11 +126,7 @@ type Context struct {
 }
 
 func (ctx *Context) currentInstruction() *InstructionContext {
-	return &ctx.instructions[len(ctx.instructions)-1]
-}
-
-func (ctx *Context) InstructionContexts() []InstructionContext {
-	return ctx.instructions
+	return ctx.currInsCtx
 }
 
 // Instructions used to recreate the table from the .debug_frame data.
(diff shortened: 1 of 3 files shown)


pkg/stack/unwind/unwind_table.go
```diff
@@ -261,9 +265,11 @@ func (ptb *UnwindTableBuilder) readFDEs(path string) (frame.FrameDescriptionEntr
 }
 
 func buildUnwindTable(fdes frame.FrameDescriptionEntries) UnwindTable {
-	table := make(UnwindTable, 0, len(fdes))
+	table := make(UnwindTable, 0)
 	for _, fde := range fdes {
-		table = append(table, buildTableRows(fde)...)
+		forEachUnwindRow(fde, func(unwindRow UnwindTableRow) {
```
kakkoyun, line 270: I'm sorry, but I need to reject this hard. I don't understand what this provides over simply returning table rows and consuming those rows in different places. There is no redundancy saving or similar. This makes this code more complicated to read. Not to mention this is not idiomatic Go. Also, `callback` implies that this is something async, but it is not.

kakkoyun (changes requested):
Even though the results are impressive, I believe we should find another approach besides `callback`s. Can't we achieve the same thing with different patterns? I'm happy to help to experiment with this.

internal/dwarf/frame/table.go
```diff
@@ -186,37 +182,48 @@ const low_6_offset = 0x3f
...
+			RetAddrReg:    cie.ReturnAddressRegister,
+			codeAlignment: cie.CodeAlignmentFactor,
+			dataAlignment: cie.DataAlignmentFactor,
+		},
 		buf:             bytes.NewBuffer(initialInstructions),
 		rememberedState: newStateStack(),
 	}
+}
+
+func executeCIEInstructions(cie *CommonInformationEntry, context *Context) *Context {
```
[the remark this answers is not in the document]
javierhonduco: That's a good idea! I was thinking of further refactoring this in a later PR as the general structure of the code is slightly awkward

pkg/stack/unwind/unwind_table.go
javierhonduco: You are correct that in this function there are no savings. We are still allocating the slice without even doing preallocation right. This is obviously not great, and `buildUnwindTable` does a lot of allocations.

Sorry that I wasn't super clear in the description. This was done on purpose as I didn't want to change how the table is generated for BPF. This PR is just about the parser and how we moved the responsibility of allocation of the table from the parser to the caller. 

I would prefer to produce the full table for this particular use case in this PR. The plan was to fully remove `buildUnwindTable` in a forthcoming PR that changes how the table is generated (https://github.com/parca-dev/parca-agent/issues/1052).

javierhonduco:
As discussed in the chat, removing the callbacks resulted in a big perf regression (see commit msg) 😮. I couldn't spot anything obvious by inspecting the code.

kakkoyun:
> As discussed in the chat, removing the callbacks resulted in a big perf regression (see commit msg) 😮. I couldn't spot anything obvious by inspecting the code.

Probably because of the closure we return. Have you tried to return the pointer directly?

javierhonduco:
> Probably because of the closure we return. Have you tried to return the pointer directly?

Which one?

pkg/stack/unwind/unwind_table.go
kakkoyun: We discussed the issue offline. And we are trying to find a compromise here. If we can't, we agree to keep the pattern because the gains are too impressive to neglect.

pkg/stack/unwind/unwind_table.go
```diff
@@ -284,14 +300,14 @@ type UnwindTableRow struct {
 	RA frame.DWRule
 }
 
-func buildTableRows(fde *frame.FrameDescriptionEntry) []UnwindTableRow {
-	rows := make([]UnwindTableRow, 0)
-
-	frameContext := frame.ExecuteDwarfProgram(fde)
-
-	instructionContexts := frameContext.InstructionContexts()
+func nextUnwindRow(fde *frame.FrameDescriptionEntry) func() *UnwindTableRow {
```
kakkoyun, line 303: @javierhonduco this one ☝️

kakkoyun:
> > Probably because of the closure we return. Have you tried to return the pointer directly?
> 
> Which one?

https://github.com/parca-dev/parca-agent/pull/1101/files#r1044656024

pkg/stack/unwind/unwind_table.go
javierhonduco: I'm not sure how to rewrite this without passed or implicit callbacks. Need to put more thought into this. Will take a look next week when I'm back. If I can't figure it out will let you know. I might be missing some insights

javierhonduco: Wasn't very inspired last week I guess :D Added some comment in the PR

javierhonduco:
The new approach that doesn't use callbacks has pretty much the same perf characteristics as the initial commit:

```
[javierhonduco@fedora parca-agent]$ benchstat /tmp/old /tmp/new2
name                                  old time/op    new time/op    delta
ParsingLibcDwarfUnwindInformation-12    58.9ms ±37%    13.4ms ±25%  -77.21%  (p=0.000 n=18+30)

name                                  old alloc/op   new alloc/op   delta
ParsingLibcDwarfUnwindInformation-12    32.7MB ± 0%     5.6MB ± 0%  -82.89%  (p=0.000 n=19+29)

name                                  old allocs/op  new allocs/op  delta
ParsingLibcDwarfUnwindInformation-12      150k ± 0%       79k ± 0%  -47.21%  (p=0.000 n=18+30)
```

Slightly better allocs/op 79k vs 83k, slightly worse timing but falls within the error margin 😄

javierhonduco:
Rebased + `s/lastInsContext/lastInsCtx` which somehow slipped through!

kakkoyun (approved):
I think I loved this approach more. Thanks a lot for investing more time in this. ❤️ 

The following is a mere suggestion as a result of my own research and experiment over the weekend. Feel free to ignore it. I'd be happy to accept the current implementation.

-----

I also checked a couple of approaches to this problem. I have a couple of suggestions for the interface:

**Option 1 (Prometheus approach):**

```
type Iterator interface {
	// At returns the current value.
	At() Value
	// Next advances the iterator by one.
	Next() bool
	// Err returns optional error if Next is false.
	Err() error
}
```

Usage:
```
it := client.Books(ctx)
for it.Next() {
	book := it.At()
	process(book)
}
if it.Err() != nil {
	return it.Err()
}
```

**Option 2 (Golang proposed):**

```
type Iterator interface {
	// Next advances the iterator by one, returns the value and if we need to continue
	Next() (val Value, ok bool) // or hasNext bool
	// Err returns optional error if Next returns false.
	Err() error
}
```
Usage:
```
it := client.Books(ctx)
for {
	book, ok := it.Next()
	process(book) // the book can be check if it's nil
	if !ok {
		break
	}
}
if it.Err() != nil {
	return it.Err()
}
```

**Option 3 (Google cloud-go approach):**

```
type Iterator interface {
	// Next advances the iterator by one, returns the value and an error
	Next() (val Value, err error) // if an error exists stop iterating
}

var Done  = errors.New("iterator done")
```

Usage:
```
it := client.Books(ctx)
loop:
for {
	book, err := it.Next()
	switch err {
	case nil:
		process(book)
	case iterator.Done:
		break loop
	default:
		return err
	}
}
```

Considering we don't return any errors, we can go with option 2 without returning error.
This would simplify the interface even more and it would be easy to port to the default iterator of Go if that proposal gets accepted.

References:
- https://github.com/prometheus/prometheus/blob/main/tsdb/chunks/chunks.go#L136
- https://pkg.go.dev/google.golang.org/api/iterator
- https://github.com/googleapis/google-cloud-go/wiki/Iterator-Guidelines
- https://github.com/golang/go/discussions/54245

javierhonduco:
Thanks for the comprehensive survey on Go iterators! I don't particularly prefer any of them, so thinking of landing the current one and once Go's iterator proposal materializes we can use it

PR #1104, opened 2022-12-09
*: Use new signed URL API for uploads when possible

brancz:
Just opening this early. This first needs https://github.com/parca-dev/parca/pull/2248 to merge.


pkg/debuginfo/manager.go
```diff
@@ -270,6 +303,43 @@ func (di *Manager) debuginfoSrcPath(ctx context.Context, buildID string, objFile
...
+		return uploadViaSignedURL(ctx, uploadInstructions.SignedUrl, r)
+	case debuginfopb.UploadInstructions_UPLOAD_STRATEGY_UNSPECIFIED:
+		return fmt.Errorf("upload strategy unspecified, must set one of UPLOAD_STRATEGY_GRPC or UPLOAD_STRATEGY_SIGNED_URL")
+	default:
+		return fmt.Errorf("unknown upload strategy: %v", uploadInstructions.UploadStrategy)
+	}
+}
+
+func uploadViaGRPC(ctx context.Context, debuginfoClient debuginfopb.DebuginfoServiceClient, uploadInstructions *debuginfopb.UploadInstructions, r io.Reader) error {
+	_, err := parcadebuginfo.NewGrpcUploadClient(debuginfoClient).Upload(ctx, uploadInstructions, r)
```
kakkoyun, line 320: How costly is it to create this client? Shall we store it as part of the manager as `Uploader`?

pkg/debuginfo/upload.go
```diff
@@ -1,111 +0,0 @@
...
-	h, err := hash.File(src.Path)
-	if err != nil {
-		return err
-	}
-
-	expBackOff := backoff.NewExponentialBackOff()
-	expBackOff.InitialInterval = time.Second
-	expBackOff.MaxElapsedTime = time.Minute
-
-	err = backoff.Retry(func() error {
```
kakkoyun, line 90: Maybe it's not needed, but we are losing to the ability to retry with the new implementation. How big of a problem is it?

pkg/debuginfo/manager.go
brancz: this has basically no cost, it's just a wrapper around the grpc client https://github.com/parca-dev/parca/blob/5541ffa12d1e0f66783092972db93f9a350f207b/pkg/debuginfo/client.go#L49-L51

pkg/debuginfo/upload.go
brancz: we still end up retrying through the profiling loop but I felt this was now obsolete since there is at most one upload anyway for any given debuginfo

kakkoyun (approved):
Thanks for the explanations ❤️

Issue #715, opened 2022-08-17
Add support for dynamically downloading and loading BTF for the running host


kakkoyun:
By using https://github.com/aquasecurity/btfhub, we can download the specific BTF definitions for the host that Agent runs on.

If `CONFIG_DEBUG_INFO_BTF=y` is enabled in the Kernel config, it's easy to find BTF definitions for the kernel under `/sys/kernel/btf/vmlinux`. However, if it's not there we can download it from the hub and load the correct definitions when initializing libbpf loader.

v-thakkar:
I think we should start with the 5.2 kernel requirement and add in the documentation that it's possible to support older kernels. We should only add this feature if we find the users that require it.

kakkoyun:
> I think we should start with the 5.2 kernel requirement and add in the documentation that it's possible to support older kernels. We should only add this feature if we find the users that require it.

I recently revisited what the supported features for the kernel version are. FWIW, BTF type format is supporter version 4.18, and above https://github.com/iovisor/bcc/blob/master/docs/kernel-versions.md I think we can have it since we claim to support 4.19 and above.

v-thakkar:
While basic support for BTF indeed existed since 4.18, the more important thing to have the information about is when support for the BTF APIs we'll have in use with the stack unwinding feature landed in the kernel. And depending upon that we can decide if it's worth adding support for the CO-RE via BTF Hub(for the older versions) or not.

For example, some stuff from older BTF things might already be outdated anyway.

kakkoyun:
Let's see how it goes. As of now, all the things we need for unwinding are in the perf_event context, so we might not need this.

Issue #1114, opened 2022-12-12
Implement an eBPF-based global process discovery mechanism


kakkoyun:
TBA

kakkoyun:
This has been done with #1507 
There will be more improvements in the area.

Issue #697, opened 2022-08-09
discovery: Add process discovery mechanism


kakkoyun:
- Use for relabelling
- Use for eh_frame provider
- Use for metadata labels

heylongdacoder:
👋 hello, is this to add a new `ProcessDiscoverer` under the discovery manager and its responsibility is to discover all the processes in the machine and enrich the label for certain pid?

PR #1131, opened 2022-12-14
Pull BTF information if missing using BTFHub

kakkoyun:
This PR enables the agent to run in a valid `CO:RE` fashion. By these changes, the eBPF programs will be able to run flawlessly for different Kernel versions and different architectures (even though we only support x86 right now, soon we will add ARM64 support).

To be able to achieve this goal:
- It downloads the supported Kernel BTF definitions from BTF Hub during builds time, embeds them to the binary, and loads those definitions in the case of missing BTF definitions for a kernel on a running host. The host's BTF definition is always prioritized.
- It converts all the helper functions with CO:RE equivalents.

Moreover, PR makes changes to the build pipeline to support building multiple eBPF objects files and consuming them to generate necessary BTF files using different definitions for corresponding Kernel versions.
 
Fixes https://github.com/parca-dev/parca-agent/issues/715

Related/Blocker for: #1209 #697 #1114

### TODO:
* [x] Add CI cache for "btfhub-archive" downloads
* [x] Cross-generate BTF for arm/x86
* [x] Support multiple eBPF object files
* [x] Add missing CI tooling: `bpftool`, ?, ...
* [x] Check if the CI host machines are BTF enabled
* [x] Modify the eBPF program to read from the task struct
  * [x] PT_REGS
  * [x] (interim or test only) task_struct->PID 
* [ ] Testing (using temporary task_struct->PID)
  * [x] Test on x86
  * [x] Test on CI
  * [ ] Test on arm64 (with lima or making CI work on ARM)
  * [ ] Check error cases when BTF files are missing


bpf/profiler/cpu.bpf.c
```diff
@@ -670,9 +670,9 @@ static __always_inline void set_initial_state(bpf_user_pt_regs_t *regs) {
   // we aren't reading garbage data.
   unwind_state->stack.len = 0;
 
-  unwind_state->ip = regs->ip;
-  unwind_state->sp = regs->sp;
-  unwind_state->bp = regs->bp;
+  unwind_state->ip = PT_REGS_IP_CORE(regs);
+  unwind_state->sp = PT_REGS_SP_CORE(regs);
+  unwind_state->bp = PT_REGS_FP_CORE(regs);
```
javierhonduco, line 675: Niiiiiiiice!! ❤️

kakkoyun:
I'll pick this up after merging #1142. We need to add tests.

kakkoyun:
After the changes in #1805, this will be easier to test.
I will start a new PR soon rather than lose my hair while rebasing this branch.

kakkoyun:
I don't think it worths resolving the conflicts.

Issue #1141, opened 2022-12-16
Testing the Agent


javierhonduco:
## The state of testing affairs 

At the moment we have:

- **unit-tests**: to ensure the correctness of small well-contained pieces of code, such as our [Kernel symbolizer](https://github.com/parca-dev/parca-agent/blob/main/pkg/ksym/ksym_test.go). This helps us ensure that the behaviour is correct, but not less importantly has everything in place for us to add regression tests if we encounter bugs that we then fix;
- **e2e tests**: running on Minikube that have the potential of testing our Kubernetes / systems metadata collection;

These two are great and are very necessary. However, we have some blindspots, that are only going to get worse once we start shipping more advanced features, such as the [DWARF-based unwinder](https://github.com/parca-dev/parca-agent/issues/768). In my opinion, we are missing three main things:
1. A way to ensure that the **kernels we want to support work** (it's easy to introduce regressions, such as a feature that's not supported in a kernel we care about, or some code pattern that the verifier in some kernels rejects. These two are things I've spotted before, not fabricated examples)
1. Ensuring that resource usage, e.g. **memory locking etc is within what we expect** (we had to recently [remediate this](https://github.com/parca-dev/parca-agent/pull/1074)) 
1. **Correctness testing**: on any kernel we support, running test programs and our unwinder and verifying that we get correct stack traces

### VM testing 
For both 1) and - partially - 2) we need to add some testing in VMs with different kernels. Instead of running all of our integration tests (which may be flaky) or unit tests  (which don't add much value in different kernels), I would like to propose the following for the kernels we support: 

- Can create the maps, so we ensure that our logic to create a map of size set at runtime is correct (note: we'll make this logic smarter to load as much as we can for a given host with a ceiling of 400-500MB);
- Load our BPF program, to ensure that the verifier is happy with our code;

This will also test that the BTF-pulling system for hosts that don't have it locally works, too (https://github.com/parca-dev/parca-agent/pull/1131).

Ideally, we'll publish a prototype of this by the end of the day with the following goals:
- quick iteration times: each VM should take less than 1min from spinning up to reporting results
- simplicity: simple to understand, and hopefully simple(r) to debug
- can be run locally, but also in CI. We would need to decide if we always want to run it in CI or only in certain occasions as it might slow things down (e.g. once a day, or just if bpf/ gets modified)

### Profile correctness tests

(This will be something to tackle early next year)

[...]

PR #1142, opened 2022-12-16
Add initial kernel testing infrastructure

javierhonduco:
This PR ensures that we can load the BPF program in different kernels with 1.6G of available memory.

See https://github.com/parca-dev/parca-agent/issues/1141

### Running locally

```
$ ./env.sh # to install bluebox
```

`qemu-system-x86_64` and `curl` should be installed, too.

```
$ time make vmtest
[...]
Test summary: 4 passed, 0 failed

real	0m53.361s
user	0m55.341s
sys	0m2.298s
```

### Test Plan
**locally, happy path**

```
$ make vmtest
[...]
=============
Test results:
=============
- ✅ 5.4
- ✅ 5.10
- ✅ 5.18
- ✅ 5.19

Test summary: 4 passed, 0 failed
```

**locally, error path**

```
$ make vmtest
[...]
=============
Test results:
=============
- ✅ 5.4
- ❌ 5.10
- ✅ 5.18
- ✅ 5.19

Test summary: 3 passed, 1 failed
(See logs in kerneltest/logs/)
make: *** [Makefile:222: vmtest] Error 1
```
**GH Actions**

See the `Kernel tests` action below

image

## TODO
- [x] Return the right status code
- [x] create `/kernels` and `/logs` directories
- [x] Fix batch tests so they are only run if the kernel supports them (we need this anyways once we use batch map APIs)
- [x] Set memlock to non unlimited
- [ ] Create our own repo where to host the kernels. Alternatively we could cache them on GH actions
- [ ] Decide how and when to run in CI (e.g. on PR or periodically? individual jobs or as-is?). We can decide this now or later on

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,6 +35,7 @@ jobs:
             [
               "**.go",
               ".github/workflows/build.yml",
+              ".github/workflows/vmtest.yml",
               ".go-version",
               "3rdparty",
               "Makefile",
@@ -115,3 +116,15 @@ jobs:
         uses: golangci/golangci-lint-action@0ad9a0988b3973e851ab0a07adf248ec2e100376 # v3.3.1
         with:
           version: ${{ env.GOLANGCI_LINT_VERSION }}
+
+      - name: Build initramfs
+        run: |
+          go install github.com/florianl/bluebox@v0.0.1
+          make initramfs
+
+      - name: Upload initramfs
+        uses: actions/upload-artifact@v3
+        with:
+          name: initramfs
+          path: |
+            kerneltest/initramfs.cpio
--- a/.github/workflows/vmtest.yml
+++ b/.github/workflows/vmtest.yml
@@ -1,0 +1,25 @@
+name: vmtest
+on:
+  workflow_run:
+    workflows: [Build]
+    types:
+      - completed
+
+jobs:
+  kernel-tests:
+    name: Kernel tests
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+
+    runs-on: ubuntu-latest
+    steps:
+        - name: Check out the code
+          uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b # v3.2.0
+        - name: Install dependencies
+          run: sudo apt -y install qemu-system-x86 curl
+        - name: Download previously generated initramfs
+          uses: actions/download-artifact@v3
+          with:
+            name: initramfs
+            path: kerneltest/
+        - name: Run tests
+          run: ./kerneltest/vmtest.sh
--- a/pkg/profiler/cpu/cpu_test.go
+++ b/pkg/profiler/cpu/cpu_test.go
@@ -23,11 +23,14 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/parca-dev/parca-agent/pkg/byteorder"
+	"github.com/parca-dev/parca-agent/pkg/profiler"
 )
 
-// The intent of these tests is to ensure that the BPF library we use,
-// (libbpfgo in this case) behaves in the way we expect.
-
+// The intent of these tests is to ensure that libbpfgo behaves the
+// way we expect.
+//
+// We also use them to ensure that different kernel versions load our
+// BPF program.
 func SetUpBpfProgram(t *testing.T) (*bpf.Module, error) {
 	t.Helper()
 
@@ -35,6 +38,10 @@ func SetUpBpfProgram(t *testing.T) (*bpf.Module, error) {
 		BPFObjBuff: bpfObj,
 		BPFObjName: "parca",
 	})
+	require.NoError(t, err)
+
+	memLock := uint64(1200 * 1024 * 1024) // ~1.2GiB
+	_, err = profiler.BumpMemlock(memLock, memLock)
 	require.NoError(t, err)
 
 	bpfMaps, err := initializeMaps(m, byteorder.GetHostByteOrder())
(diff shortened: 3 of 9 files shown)


kerneltest/kerneltest.sh
```diff
@@ -0,0 +1,127 @@
...
+echo "============="
+for kernel in "${kernel_versions[@]}"; do
+    if did_test_pass "$kernel"; then
+        echo "- ✅ $kernel"
+    else
+        echo "- ❌ $kernel"
+    fi
+done
+
+# BUG: we need to check for batch operation support and don't run these tests.
```
javierhonduco, line 117: If we plan on using bach ops, we can add some detection and not run the tests, otherwise, perhaps we can move this to another test? cc @v-thakkar @kakkoyun

kerneltest/kerneltest.sh
```diff
@@ -0,0 +1,127 @@
...
+# use_kernel "5.4"
+# run_tests "5.4" "1.5G"
+# did_test_pass "5.4"
+#
+# BUG: on 4.19. Fix backedge, the verifier spotted the loop and it's not happy, we need to unroll it here.
```
javierhonduco, line 122: This one will take a bit longer to tackle. Will take another look once DWARF unwinding is closer to the finishing line. Either way, I think it would be worth discussing the minimum kernel version we want to support. Let's do that in January. @kakkoyun @v-thakkar

kakkoyun: Better to run in a dedicated pipeline.

kerneltest/kerneltest.sh
kakkoyun: Better to run in a dedicated pipeline.

.github/workflows/build.yml
```diff
@@ -115,3 +115,15 @@ jobs:
...
         with:
           version: ${{ env.GOLANGCI_LINT_VERSION }}
+
+  kernel-tests:
+    name: Kernel testing
+    runs-on: ubuntu-latest
+    needs: go-build-test
+    steps:
+        - name: Install qemu and curl
+          run: sudo apt -y install qemu-system-x86 curl
```
kakkoyun, line 125: To note, our previous experiences with QEMU were sub-par. It was painfully slow. Mainly emulating ARM on these action machines. That being said, we can always try to run these when we merge to the main, and it'd reduce the pain on PRs. Let's see how this goes.

kakkoyun (commented):
Nice!!!

kerneltest/kerneltest.sh
javierhonduco: Hiii! What do you mean by pipeline? 

I've added some basic feature checks right now. Ideally, in the future, we would add this check to the `profiler` struct + wrap the error with `%w` in libbpfgo so the error can be unwrapped. Going with the current approach as we don't use batch operations anywhere yet, and changing the error to be wrapped in libbpfgo would take some days to land.

.github/workflows/build.yml
javierhonduco: Indeed! That's something I want to be mindful of. I think that the past experiences were particularly bad because we were emulating arm64. On x86 this shouldn't be as slow, despite [GH actions not supporting nested virtualization](https://github.com/community/community/discussions/8305).

Either way, I suggest that we first see how long it would take, and then decide on the course of action. For example, we could do the VM tests periodically. Something like every day? That way we can ensure that the profiler works fine in different kernels for most commits in `main` and there shouldn't be that many changes to bisect if something breaks.

If we go for this approach we need to decide how is the responsibility of troubleshooting breakages going to be distributed among the maintainers/team members.

javierhonduco:
The CI failures are known https://github.com/parca-dev/parca-agent/pull/1185

kerneltest/kerneltest.sh
javierhonduco: Resolving to clear things up here, but let's chat about this in the next few weeks

README.md
```diff
@@ -59,7 +59,7 @@ Flags:
   -h, --help                      Show context-sensitive help.
       --log-level="info"          Log level.
       --http-address=":7071"      Address to bind HTTP server to.
-      --node="hostname"
```
javierhonduco: Ugh, this bot is removing my change to make our README.md pass... https://github.com/parca-dev/parca-agent/compare/fda48e4dfaa7f09bfac98d8998c5e4ac2f29e3b6..ff2d60ea39d5b9a4fdf7f9053cfcc1aa44ca11bd

kerneltest/kerneltest.sh
kakkoyun: Hi 👋 What I meant is to run kernel tests as in their own GitHub action, not part of the `build.yaml`. It would be easier to track down the failures and how long do they take.

.github/workflows/build.yml
kakkoyun: I would be happy to experiment. If they take too much time, we can move on to the daily approach and make the GH action open an issue in case of a failure. It should be straightforward to implement. 

Referring to my previous comment, let's separate the action so that we can track how long it takes quickly.

kakkoyun (commented):
Great work! 

I have a couple of comments in line.

pkg/profiler/cpu/cpu_test.go
```diff
@@ -85,7 +92,29 @@ func TestDeleteExistentKey(t *testing.T) {
 	require.NoError(t, err)
 }
 
+func hasBatchOperations(t *testing.T) bool {
```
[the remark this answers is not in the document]
javierhonduco: This is the most deterministic way I can think of

> I feel like this could fail for any other reason.

But you are right about this! I was planning to add one more test that ensures that batch operations are supported in the kernel release that we expect or greater. That way we will ensure that this code is not broken. Sorry that I forgot to mention this. Will cut another PR after this one to add this

.github/workflows/build.yml
javierhonduco: Cool, will create another workflow. Hopefully we should be able to do that 😄

pkg/profiler/cpu/cpu_test.go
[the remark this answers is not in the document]
kakkoyun: 👍 Awesome!

README.md
```diff
@@ -59,7 +59,8 @@ Flags:
   -h, --help                      Show context-sensitive help.
       --log-level="info"          Log level.
       --http-address=":7071"      Address to bind HTTP server to.
-      --node="hostname"           The name of the node that the process is
+      --node="hostname"
```
kakkoyun, line 62: Why do we have the new line here?

kakkoyun: Go home, @pre-commit-ci-lite. You're drunk!

kakkoyun:
Hmm, why do we have merged commits on this? Did I do it?

javierhonduco:
Thought you had used the update branch feature to merge `main` in this branch

.github/workflows/vmtest.yml
```diff
@@ -0,0 +1,25 @@
+name: vmtest
+on:
+  workflow_run:
+    workflows: [Build]
```
kakkoyun, line 4:
```suggestion
    workflows: ["Build"]
```

This could be the culprit. We need to try.

javierhonduco:
The `vmtest` workflow did not start because according to [the documentation](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows):

> To trigger a workflow runs on the workflow_run event, the workflow file should exist on the master or default branch.

Given that it worked fine when it was in the `Build` workflow, it should probably work now. We will iterate if this is not the case. We considered `workflow_call` and `workflow_dispatch` but decided not to use them as it would require extra logic in our workflows to ensure that they were called by the right workflow, which would complicate our setup

kakkoyun (approved):
LGTM

PR #1138, opened 2022-12-15
plg/process: Cache hash file

marselester:
An attempt to tackle https://github.com/parca-dev/parca-agent/issues/1100.

```
# new
BenchmarkPIDMappingFileCache-2   	  135170	      8710 ns/op	    3468 B/op	      32 allocs/op
# old
BenchmarkPIDMappingFileCache-2   	   16863	     76147 ns/op	   36268 B/op	      33 allocs/op
```

PR #1145, opened 2022-12-16
pkg/debuginfo: ensureUploaded benchmark

marselester:
@brancz [mentioned](https://github.com/parca-dev/parca-agent/pull/1138#issuecomment-1354576440) that `EnsureUploaded` has a room for perf improvements.

> Judging by what I'm seeing on demo I think we have much bigger fish to fry right now (allocations in EnsureUploaded and profiler loop stacks look like they should give us a major wins).

I found a TODO comment that mentions caching the hash, so I was thinking to prepare a benchmark for that code path to begin with.

```
// TODO: We should be able to cache the hash further to avoid
// re-hashing the same binary and getting to the same result
// again.
```

```
$ go test -benchmem -run=. -bench ^BenchmarkEnsureUploadedAlreadyExists$ ./pkg/debuginfo/
goos: linux
goarch: amd64
pkg: github.com/parca-dev/parca-agent/pkg/debuginfo
cpu: Intel(R) Core(TM) i5-10600 CPU @ 3.30GHz
BenchmarkEnsureUploadedAlreadyExists-2 1000000 1131 ns/op 138 B/op 7 allocs/op
```

I guess the hashes can be cached (`buildID` is a key) in a `sync.Map` since the files are uploaded concurrently.

### DIFF
--- a/pkg/debuginfo/client.go
+++ b/pkg/debuginfo/client.go
@@ -21,17 +21,26 @@ import (
 	"google.golang.org/grpc"
 )
 
-type NoopClient struct{}
+type NoopClient struct {
+	ShouldInitiateUploadF func(in *debuginfopb.ShouldInitiateUploadRequest, opts ...grpc.CallOption) (*debuginfopb.ShouldInitiateUploadResponse, error)
+	InitiateUploadF       func(in *debuginfopb.InitiateUploadRequest, opts ...grpc.CallOption) (*debuginfopb.InitiateUploadResponse, error)
+}
 
 func (c *NoopClient) Upload(ctx context.Context, opts ...grpc.CallOption) (debuginfopb.DebuginfoService_UploadClient, error) {
 	return nil, nil
 }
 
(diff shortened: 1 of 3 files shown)


pkg/debuginfo/client.go
```diff
@@ -21,17 +21,26 @@ import (
 	"google.golang.org/grpc"
 )
 
-type NoopClient struct{}
+type NoopClient struct {
+	ShouldInitiateUploadF func(in *debuginfopb.ShouldInitiateUploadRequest, opts ...grpc.CallOption) (*debuginfopb.ShouldInitiateUploadResponse, error)
```
kakkoyun, line 25: Rather than having these functions as part of `NoopClient,` we could implement a client for testing in the test file. That way, it would be easier to follow. What do you think?

marselester: That's a good idea 👍

marselester: I've decoupled a test client from `NoopClient`. It can be made exportable later on if some package needs it.

kakkoyun (approved):
LGTM. Thank you so much.

PR #1150, opened 2022-12-19
pkg/profiler: export prometheus metrics for bpf maps

### LINKED ISSUE #1093: Add BPF map introspection facilities
## The problem

We currently don't have a way to query the kernel for information on the BPF maps we are creating. We would like to know the number of items they store, and how much memory they are taking (perhaps also how big their key and their values are).

This would be super useful for debugging, and potentially we could also send this information via Prometheus to understand how maps are sized across different installations. For example, different kernels might have different layouts for the map's metadata, increasing or reducing the overhead.

## Proposed API

Something like this pseudo-code:

```go
bpfModule := loadBPF()

for bpfMap := bpfModule.MapsIterator(); bpfMap; bpfMap.Next() {
  name := bpfMap.Name() 
  lockedMemory := bpfMap.LockedMemory() 

  fmt.Println("map name:", name, "bytes locked", lockedMemory)
}
```

## Other tools in this space

```
$ sudo bpftool prog | tail
130: cgroup_skb  tag 6deef7357e7b4530  gpl
	loaded_at 2022-12-05T11:44:33+0000  uid 0
	xlated 64B  jited 104B  memlock 4096B
	pids systemd(1)
131: cgroup_skb  tag 6deef7357e7b4530  gpl
	loaded_at 2022-12-05T11:44:33+0000  uid 0
	xlated 64B  jited 104B  memlock 4096B
	pids systemd(1)
```

Shows every map in the system (we just want the maps for a given BPF program), with their locked memory, etc

cc @kakkoyun @v-thakkar

Sylfrena:
image

Fixes #1093 

### DIFF
--- a/pkg/profiler/cpu/bpf_metrics.go
+++ b/pkg/profiler/cpu/bpf_metrics.go
@@ -1,0 +1,114 @@
+// Copyright 2022 The Parca Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package cpu
+
+import (
+	"bufio"
+	"bytes"
+	_ "embed"
+	"fmt"
+	"io"
+	"os"
+	"regexp"
+	"strconv"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+)
+
+var memlockRegex = regexp.MustCompile(`^memlock:\s+(\d+)$`)
+
+func (c *bpfMetricsCollector) getBPFMetrics() []*bpfMetrics {
+	var bpfMapsNames []string
+	var bpfMetricArray []*bpfMetrics
+
+	it := c.m.Iterator()
+
+	for {
+		mapBpf := it.NextMap()
+		if mapBpf != nil {
+			bpfMapsNames = append(bpfMapsNames, mapBpf.Name())
+		} else {
+			break
+		}
+	}
+
+	for _, mapName := range bpfMapsNames {
+		bpfMap, err := c.m.GetMap(mapName)
+		if err != nil {
+			level.Debug(c.logger).Log("msg", "error fetching bpf map", "err", err)
+			continue
+		}
+
+		bpfMaxEntry := float64(bpfMap.GetMaxEntries())
+		bpfMapKeySize := float64(bpfMap.KeySize())
+		bpfMapValueSize := float64(bpfMap.ValueSize())
+		bpfMapFd := fmt.Sprint(bpfMap.FileDescriptor())
+
+		path := fmt.Sprintf("/proc/%d/fdinfo/", c.pid) + bpfMapFd
+		data, err := readFileNoStat(path)
+		if err != nil {
+			level.Debug(c.logger).Log("msg", "Unable to read fds for agent process", "agent_pid", c.pid, "err", err)
+		}
+
+		bpfMemlock, err := FdInfoMemlock(c.logger, data)
+		if err != nil {
+			level.Debug(c.logger).Log("msg", "error getting memory locked for file descriptor", "err", err)
+		}
+
+		bpfMetricArray = append(bpfMetricArray,
+			&bpfMetrics{
+				mapName:         mapName,
+				bpfMapKeySize:   bpfMapKeySize,
+				bpfMapValueSize: bpfMapValueSize,
+				bpfMaxEntry:     bpfMaxEntry,
+				bpfMemlock:      float64(bpfMemlock),
+			},
+		)
+	}
+	return bpfMetricArray
+}
+
+// FDInfoMemlock returns the memory locked by the fd for a process using fdinfo data.
+func FdInfoMemlock(logger log.Logger, data []byte) (int, error) {
+	var text, memlockValue string
+
+	scanner := bufio.NewScanner(bytes.NewReader(data))
+	for scanner.Scan() {
+		text = scanner.Text()
+		if memlockRegex.MatchString(text) {
+			memlockValue = memlockRegex.FindStringSubmatch(text)[1]
+		}
+	}
+
+	memlockValueInt, err := strconv.Atoi(memlockValue)
+	if err != nil {
+		level.Debug(logger).Log("msg", "error converting memlock to integer type", "err", err)
+	}
+	return memlockValueInt, nil
+}
+
+func readFileNoStat(filename string) ([]byte, error) {
+	const maxBufferSize = 1024 * 1024
+
+	f, err := os.Open(filename)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	reader := io.LimitReader(f, maxBufferSize)
+	return io.ReadAll(reader)
+}
--- a/pkg/profiler/cpu/bpf_metrics_test.go
+++ b/pkg/profiler/cpu/bpf_metrics_test.go
@@ -1,0 +1,44 @@
+// Copyright 2022 The Parca Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package cpu
+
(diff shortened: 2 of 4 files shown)


Sylfrena:
Not really ready for review yet, still a few things to take care of.

pkg/profiler/cpu/bpf_metrics.go
```diff
@@ -0,0 +1,109 @@
...
+		bpfMemlock, err := FDInfoMemlock(logger, bpfMapFd, p)
+		if err != nil {
+			level.Debug(logger).Log("msg", "error getting memory locked for file descriptor", "err", err)
+		}
+
+		mapMetrics.WithLabelValues(mapName, bpfMapKeySize, bpfMapValueSize, bpfMaxEntry).Set(float64(bpfMemlock))
+	}
+}
+
+// FDInfoMem returns the memory locked by the fd for a process
```
Sylfrena, line 70:
```suggestion
// FDInfoMemlock returns the memory locked by the fd for a process
```

pkg/profiler/cpu/bpf_metrics.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	for _, mapName := range bpfMapsNames {
+		bpfMap, err := m.GetMap(mapName)
+		if err != nil {
+			level.Debug(logger).Log("msg", "error fetching bpf map", "err", err)
+			continue
+		}
+
+		bpfMaxEntry := fmt.Sprint(bpfMap.GetMaxEntries())
```
v-thakkar, line 55: Love that we're also adding this to the metrics.

pkg/profiler/cpu/bpf_metrics.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	memlockValueInt, err := strconv.Atoi(memlockValue)
+	if err != nil {
+		level.Debug(logger).Log("msg", "error converting memlock to integer type", "err", err)
+	}
+	return memlockValueInt, nil
+}
+
+// This is an intenal function from the prometheus/procfs/utils pkg
```
v-thakkar, line 96: Typo: intenanl->internal

What is the purpose of this comment? If it's just for informational purposes, it's ok to drop it.

pkg/profiler/cpu/bpf_metrics.go
```diff
@@ -0,0 +1,108 @@
...
+		if err != nil {
+			level.Debug(logger).Log("msg", "error getting memory locked for file descriptor", "err", err)
+		}
+
+		mapMetrics.WithLabelValues(mapName, bpfMapKeySize, bpfMapValueSize, bpfMaxEntry).Set(float64(bpfMemlock))
+	}
+}
+
+// FDInfoMemlock returns the memory locked by the fd for a process
+func FDInfoMemlock(logger log.Logger, fd string, p procfs.Proc) (int, error) {
```
v-thakkar, line 70: Will be nice to have a small test for this function.

v-thakkar (commented):
Have some general comments but looking good otherwise :tada:

pkg/profiler/cpu/bpf_metrics.go
kakkoyun, line 70:
```suggestion
func FDInfoMemlock(logger log.Logger, fd string, pid int) (int, error) {
```
Only PID is used. It's better to pass minimal types.

kakkoyun, line 70: Also, this function looks pretty generic, it could be part of the `process` package as a helper.

pkg/profiler/cpu/bpf_metrics.go
```diff
@@ -0,0 +1,108 @@
...
+
+	memlockValueInt, err := strconv.Atoi(memlockValue)
+	if err != nil {
+		level.Debug(logger).Log("msg", "error converting memlock to integer type", "err", err)
+	}
+	return memlockValueInt, nil
+}
+
+// This is an intenal function from the prometheus/procfs/utils pkg
+func ReadFileNoStat(filename string) ([]byte, error) {
```
kakkoyun, line 97:
```suggestion
func readFileNoStat(filename string) ([]byte, error) {
```

No need to export the method, it's only used here.

pkg/profiler/cpu/bpf_metrics.go
```diff
@@ -0,0 +1,108 @@
...
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/procfs"
+
+	bpf "github.com/aquasecurity/libbpfgo"
+)
+
+func GetBPFMetrics(logger log.Logger, m *bpf.Module, p procfs.Proc, mapMetrics *prometheus.GaugeVec) {
```
kakkoyun, line 35:
```suggestion
func getBPFMetrics(logger log.Logger, m *bpf.Module, p procfs.Proc, mapMetrics *prometheus.GaugeVec) {
```
No need to export, as it seems. Also, PID could be enough.

One other idea worth exploring it this method to be part of the `maps.go` and `epbpfMaps` type.

kakkoyun (commented):
Great job! I've some suggestions and as @v-thakkar proposed it'd be nice to have tests.

pkg/profiler/cpu/metrics.go
```diff
@@ -60,5 +61,13 @@ func newMetrics(reg prometheus.Registerer) *metrics {
 				NativeHistogramBucketFactor: 1.1,
 			},
 		),
+		obtainMapMetrics: promauto.With(reg).NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name:        "parca_agent_bpf_map_memlock",
+				Help:        "Metrics for BPF maps created by the agent",
+				ConstLabels: map[string]string{"type": "cpu"},
+			},
+			[]string{"bpf_map_name", "bpf_map_keySize", "bpf_map_valueSize", "bpf_max_entries"},
```
brancz, line 70: This doesn't look quite right, I would expect to have a distinct metric for entries, key size, value size and memlock value. Only `bpf_map_name` sounds like something that should be a label.

pkg/profiler/cpu/metrics.go
```diff
@@ -60,5 +61,13 @@ func newMetrics(reg prometheus.Registerer) *metrics {
 				NativeHistogramBucketFactor: 1.1,
 			},
 		),
+		obtainMapMetrics: promauto.With(reg).NewGaugeVec(
```
brancz, line 64: Is it possible that we add/remove new maps at runtime? If so then we should rather implement a Prometheus collector, rather than setting the metrics. A collector instead collects the metrics at scrape time, rather than setting them synchronously like here. If maps might come and go over time we should implement a collector (the implemented functionality would still be super close to what was implemented in this PR, just thinking about future-proofing this).

pkg/profiler/cpu/metrics.go
```diff
@@ -62,3 +65,65 @@ func newMetrics(reg prometheus.Registerer) *metrics {
...
+	ch <- descBPFMapMaxEntries
+}
+
+func (c *bpfMetricsCollector) Collect(ch chan<- prometheus.Metric) {
+	for _, bpfMetrics := range c.getBPFMetrics() {
+		ch <- prometheus.MustNewConstMetric(descBPFMemlock, prometheus.GaugeValue, bpfMetrics.bpfMemlock, bpfMetrics.mapName)
+		ch <- prometheus.MustNewConstMetric(descBPFMapKeySize, prometheus.GaugeValue, bpfMetrics.bpfMapKeySize, bpfMetrics.mapName)
+		ch <- prometheus.MustNewConstMetric(descBPFMapValueSize, prometheus.GaugeValue, bpfMetrics.bpfMapValueSize, bpfMetrics.mapName)
+		ch <- prometheus.MustNewConstMetric(descBPFMapMaxEntries, prometheus.GaugeValue, bpfMetrics.bpfMaxEntry, bpfMetrics.mapName)
+		level.Info(c.logger).Log("msg", "memlock send", "memlockval", bpfMetrics.bpfMemlock)
```
brancz, line 127: nit: info level is going to be much too noisy at normal operation and since this log line does exactly what the metrics do, I would say let's just skip it altogether

brancz (commented):
Just need that test and whatever the linter wants fixed about the license header and I think this is good to merge!

pkg/profiler/cpu/metrics.go
Sylfrena: Good catch, yep, we don't really need this : )

pkg/profiler/cpu/bpf_metrics_test.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"testing"
+
+	"github.com/go-kit/log"
+	"github.com/stretchr/testify/require"
+)
+
+func TestFdInfoMemlock(t *testing.T) {
+	data := []byte(
+		"pos:	0\nflags:	02000002\nmnt_id:	15\nino:	2081\nmap_type:	1\nkey_size:	4\nvalue_size:	1\nmax_entries:	32\nmap_flags:	0x0\nmap_extra:	0x0\nmemlock:	4096\nmap_id:	68\nfrozen:	0")
```
brancz, line 28: can we make this a multi-line string? this is super hard to read

brancz (approved):
Nice work!

javierhonduco:
👏 nice job!!

javierhonduco:
👏 nice one!!

kakkoyun:
Very neat 👏

PR #1065, opened 2022-11-24
pkg/profile: Replace gzip package to reduce allocs

marselester:
A drop-in replacement for gzip has impressively low allocs with decent speed/compression https://github.com/klauspost/compress/pull/216. See, gzip related code no longer shows up in the pprof top20.

<details>

```
(pprof) top20 -cum
Showing nodes accounting for 104.70MB, 40.74% of 257.02MB total
Dropped 109 nodes (cum <= 1.29MB)
Showing top 20 nodes out of 125
      flat  flat%   sum%        cum   cum%
         0     0%     0%   100.85MB 39.24%  github.com/oklog/run.(*Group).Run.func1
         0     0%     0%   100.85MB 39.24%  github.com/parca-dev/parca-agent/pkg/profiler/cpu.(*CPU).Run
         0     0%     0%   100.85MB 39.24%  main.run.func6
         0     0%     0%   100.85MB 39.24%  main.run.func6.1
         0     0%     0%   100.85MB 39.24%  runtime/pprof.Do
         0     0%     0%    73.71MB 28.68%  github.com/godbus/dbus/v5.(*Conn).inWorker
    8.58MB  3.34%  3.34%    73.71MB 28.68%  github.com/godbus/dbus/v5.(*unixTransport).ReadMessage
    9.11MB  3.55%  6.88%    64.13MB 24.95%  github.com/godbus/dbus/v5.DecodeMessage
         0     0%  6.88%    55.02MB 21.41%  github.com/godbus/dbus/v5.(*decoder).Decode
      50MB 19.46% 26.34%    55.02MB 21.41%  github.com/godbus/dbus/v5.(*decoder).decode
    3.02MB  1.17% 27.51%    54.02MB 21.02%  github.com/coreos/go-systemd/v22/dbus.(*Conn).SubscribeUnitsCustom.func1
         0     0% 27.51%       51MB 19.84%  github.com/coreos/go-systemd/v22/dbus.(*Conn).ListUnits (inline)
         0     0% 27.51%       51MB 19.84%  github.com/coreos/go-systemd/v22/dbus.(*Conn).ListUnitsContext
   12.49MB  4.86% 32.37%       50MB 19.45%  github.com/coreos/go-systemd/v22/dbus.(*Conn).listUnitsInternal
         0     0% 32.37%    44.89MB 17.47%  github.com/parca-dev/parca-agent/pkg/metadata.(*StatelessProvider).Labels
[...]

PR #1161, opened 2022-12-21
pkg/profiler: Use pool of gzip writers

marselester:
klauspost's Stateless gzip compressor allowed to reduce CPU consumption by 20%, see https://github.com/parca-dev/parca-agent/pull/1065.

After profiling Parca Agent with Vtune I found an elevated CPI rate (124.4) in flate/stateless.go https://github.com/klauspost/compress/discussions/717. The author recommends to use a pool of gzip writers to get 2x-3x speed up.

<details>

```
(pprof) top20 -cum
Showing nodes accounting for 175.04MB, 36.80% of 475.70MB total
Dropped 155 nodes (cum <= 2.38MB)
Showing top 20 nodes out of 118
      flat  flat%   sum%        cum   cum%
         0     0%     0%   189.97MB 39.93%  github.com/godbus/dbus/v5.(*Conn).inWorker
   31.86MB  6.70%  6.70%   189.97MB 39.93%  github.com/godbus/dbus/v5.(*unixTransport).ReadMessage
         0     0%  6.70%   157.60MB 33.13%  github.com/godbus/dbus/v5.(*decoder).Decode
  111.01MB 23.34% 30.04%   157.60MB 33.13%  github.com/godbus/dbus/v5.(*decoder).decode
         0     0% 30.04%   157.10MB 33.03%  github.com/godbus/dbus/v5.decodeMessageBody
         0     0% 30.04%   133.76MB 28.12%  github.com/oklog/run.(*Group).Run.func1
         0     0% 30.04%   133.76MB 28.12%  github.com/parca-dev/parca-agent/pkg/profiler/cpu.(*CPU).Run
         0     0% 30.04%   133.76MB 28.12%  main.run.func6
         0     0% 30.04%   133.76MB 28.12%  main.run.func6.1
         0     0% 30.04%   133.76MB 28.12%  runtime/pprof.Do
    6.03MB  1.27% 31.30%   130.23MB 27.38%  github.com/coreos/go-systemd/v22/dbus.(*Conn).SubscribeUnitsCustom.func1
         0     0% 31.30%   124.20MB 26.11%  github.com/coreos/go-systemd/v22/dbus.(*Conn).ListUnits (inline)
         0     0% 31.30%   124.20MB 26.11%  github.com/coreos/go-systemd/v22/dbus.(*Conn).ListUnitsContext
   26.13MB  5.49% 36.80%   122.20MB 25.69%  github.com/coreos/go-systemd/v22/dbus.(*Conn).listUnitsInternal
         0     0% 36.80%    96.06MB 20.19%  github.com/godbus/dbus/v5.Store
         0     0% 36.80%    96.06MB 20.19%  github.com/godbus/dbus/v5.store
         0     0% 36.80%    96.06MB 20.19%  github.com/godbus/dbus/v5.storeInterfaces
         0     0% 36.80%    96.06MB 20.19%  github.com/godbus/dbus/v5.storeSlice
         0     0% 36.80%    65.06MB 13.68%  github.com/godbus/dbus/v5.(*Call).Store
         0     0% 36.80%    65.06MB 13.68%  github.com/godbus/dbus/v5.storeSliceIntoSlice

ROUTINE ======================== github.com/parca-dev/parca-agent/pkg/profiler.(*RemoteProfileWriter).Write in pkg/profiler/profile_writer.go
         0    13.52MB (flat, cum)  2.84% of Total
         .          .     78:// Write sends the profile using the designated write client.
         .          .     79:func (rw *RemoteProfileWriter) Write(ctx context.Context, labels model.LabelSet, prof *profile.Profile) error {
         .          .     80:	buf := bytes.NewBuffer(nil)
         .          .     81:	zw := rw.pool.Get().(*gzip.Writer)
         .          .     82:	zw.Reset(buf)
         .     7.81MB     83:	if err := prof.WriteUncompressed(zw); err != nil {
         .          .     84:		rw.pool.Put(zw)
         .          .     85:		return err
         .          .     86:	}
         .     5.71MB     87:	zw.Close()
         .          .     88:	rw.pool.Put(zw)
         .          .     89:
         .          .     90:	_, err := rw.profileStoreClient.WriteRaw(ctx, &profilestorepb.WriteRawRequest{
         .          .     91:		Normalized: true,
         .          .     92:		Series: []*profilestorepb.RawProfileSeries{{
```

</details>

I am yet to check the CPI rate after introducing the pool.


marselester:
The gzip compressor no longer shows up in the top of CPI rate.

Screenshot 2022-12-21 at 21 25 21

zdyj3170101136:
may be send uncompressed data is ok?
the agent and the server is both running in same network.
huandreds of KB data would not save too much network resource.

brancz:
It's a trade-off. If it's only compiled processes I would agree, but we also resolve perf-maps and kernel symbols. For those types of workloads it is a good idea to use some compression. I'd be ok with having an option to turn off compression as a flag if people want that for their setup, I would still say the default should be to have compression on. We actually went with exactly the same thing in Prometheus.

pkg/profiler/profile_writer.go
```diff
@@ -59,29 +60,34 @@ func (fw *FileProfileWriter) Write(_ context.Context, labels model.LabelSet, pro
...
-	zw, err := gzip.NewWriterLevel(buf, gzip.StatelessCompression)
-	if err != nil {
-		return err
-	}
-	if err = prof.WriteUncompressed(zw); err != nil {
-		zw.Close()
+	zw, _ := rw.pool.Get().(*gzip.Writer)
+	zw.Reset(buf)
+	if err := prof.WriteUncompressed(zw); err != nil {
+		rw.pool.Put(zw)
```
kakkoyun, line 84: Don't we need to close the `zw` regardless?

marselester: Since `zw` gets reset before it's used again, a `Close` call is redundant (it writes a gzip footer). I will add it for symmetry since that code path won't be taken often anyways.

kakkoyun: Thanks for the explanation ❤️

pkg/profiler/profile_writer.go
```diff
@@ -59,29 +60,35 @@ func (fw *FileProfileWriter) Write(_ context.Context, labels model.LabelSet, pro
...
+	// pool of gzip encoders helps to reduce GC pressure.
+	pool sync.Pool
 }
 
 // NewRemoteProfileWriter creates a new RemoteProfileWriter.
 func NewRemoteProfileWriter(profileStoreClient profilestorepb.ProfileStoreServiceClient) *RemoteProfileWriter {
 	return &RemoteProfileWriter{
 		profileStoreClient: profileStoreClient,
+		pool: sync.Pool{New: func() interface{} {
+			z, _ := gzip.NewWriterLevel(nil, gzip.BestSpeed)
```
kakkoyun, line 72: Let's not ignore errors :)

pkg/profiler/profile_writer.go
```diff
@@ -59,29 +60,35 @@ func (fw *FileProfileWriter) Write(_ context.Context, labels model.LabelSet, pro
...
 
 // Write sends the profile using the designated write client.
 func (rw *RemoteProfileWriter) Write(ctx context.Context, labels model.LabelSet, prof *profile.Profile) error {
 	buf := bytes.NewBuffer(nil)
-	zw, err := gzip.NewWriterLevel(buf, gzip.StatelessCompression)
-	if err != nil {
-		return err
-	}
-	if err = prof.WriteUncompressed(zw); err != nil {
+	zw, _ := rw.pool.Get().(*gzip.Writer)
```
kakkoyun, line 81: ditto

kakkoyun (commented):
This is very close to merging. Let's not ignore errors. And we can merge this PR.

Thanks for this work.

pkg/profiler/profile_writer.go
marselester: It's safe enough because `gzip.NewWriter()` does the same thing https://github.com/klauspost/compress/blob/master/gzip/gzip.go#L58. There will be no error since `gzip.BestSpeed` is a constant.

pkg/profiler/profile_writer.go
marselester: Since the pool contains only `*gzip.Writer`, then the type casting should always succeed unless some other object was added to the pool. In that case the program should panic, therefore there mustn't be an underscore. Does it sound reasonable?

pkg/profiler/profile_writer.go
kakkoyun: I hear you, however, the underlying implementation can change, and I don't feel comfortable ignoring errors. We can change the API and return the error to the caller and let it handle it.

pkg/profiler/profile_writer.go
kakkoyun: For sure. I would rather program against the interface and in this case, the contract (interface) tells me that there could be an error. Underlying implementation can change. It's a low-hoop to jump, either propagate or handle the error here. What do you think?

pkg/profiler/profile_writer.go
marselester: Sounds good.

marselester: Oh, `sync.Pool.New` accepts `func() interface{}`, so in order to return an error along with `*gzip.Writer` I would have to introduce a struct to wrap those two arguments.

marselester: If there is an error, then a pool would return `nil` and type casting will panic `rw.pool.Get().(*gzip.Writer)`, so the error will be spotted immediately. Additionally, we can pass a logger to `NewRemoteProfileWriter()` and log an error so it will be easier to figure out why `sync.Pool.New` returned a nil.

kakkoyun (approved):
LGTM 👍

PR #1171, opened 2022-12-26
Fix missing alignment in dbus

marselester:
During dbus refactoring I've introduced a bug (missed an alignment and broke parsing of bool values) which was fixed at https://github.com/godbus/dbus/pull/348. Sorry about that.


maxbrunet (approved):
LGTM :heart:

PR #1176, opened 2022-12-30
(chore) Spelling fix

andrewa-stripe:
uplod->upload

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ Flags:
                                   The duration to cache debuginfo upload exists
                                   checks for.
       --debuginfo-upload-timeout-duration=2m
-                                  The timeout duration to cancel uplod requests.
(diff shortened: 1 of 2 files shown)


brancz (approved):
Thank you!

brancz:
Actually, it looks like that caused "requests." to go to the next line. You can either run `make README.md` locally or just edit the generated docs to match.

@maxbrunet I think we have some "auto-fix" linters, don't we? Could we auto-fix this type of thing as well?

maxbrunet:
> @maxbrunet I think we have some "auto-fix" linters, don't we? Could we auto-fix this type of thing as well?

We do, it is intended to be used with pre-commit and building the agent would be to heavy for a hook, but let's try using it in this workflow: PR #1183

(the main issue I can think of is if both workflows try to commit they might conflict)

PR #1188, opened 2023-01-05
vmtest: Fix download from another workflow

javierhonduco:


javierhonduco:
Merging without review because it's impossible to iterate on this otherwise :(

PR #1193, opened 2023-01-05
Move vmtest to the build workflow

javierhonduco:
To make things simpler. Other ways of implementing this have the following issues:

- need to maintain state between workflows, which makes everything more complex;
- can't be tested properly before landing as they won't execute until they are on `main`;

In the end, this approach is the simplest, and we can still see [the execution time for this job](https://github.com/parca-dev/parca-agent/actions/runs/3849187277/usage). It should not be more than 1-2 minutes and has to run after the build anyways, so these two steps have to execute serially 


maxbrunet (commented):
Could you still create a distinctive job for it? Similarly to how we do in [.github/workflows/container.yml](https://github.com/parca-dev/parca-agent/blob/main/.github/workflows/container.yml)

javierhonduco:
The runtime for the past execution of the vmtests: https://github.com/parca-dev/parca-agent/actions/runs/3849187277/usage

maxbrunet (approved):
LGTM!

javierhonduco:
Thanks, Maxime! Let's wait to see what Kemal thinks before committing!

kakkoyun (approved):
LGTM

Issue #1187, opened 2023-01-05
let parca-agent do not store nonormalized address.


zdyj3170101136:
current, the agent compute base address from executable file, and the pprof file's address is normalized.
address -= base.

it is difficulty for github.com/google/pprof file to deal with this kind of file.
it's function like profile.Merge is assume the address is unnormalized.

it is possible to store the information which we need to compute base address into debuginfo?
parca-agent store raw address.
parca use debuginfo to compute base address.

kakkoyun:
I will propose changes regarding these mechanisms.

PR #1196, opened 2023-01-06
pkg/address: Remove normalizer and compute base address on the server-side

zdyj3170101136:
fix https://github.com/parca-dev/parca-agent/issues/1187


brancz:
This isn’t quite equivalent to what we had before, it would mean that the server must already have the binary to be able to accept profiling data. Since the agent must always have the binary present (as it’s literally what’s being executed), we should extract the necessary information on the agent and send it along with the profiling data. The server can then normalize the data and return the data again in the expected representation when exporting as pprof.

zdyj3170101136:
> 这与我们之前的不完全相同，这意味着服务器必须已经具有二进制文件才能接受分析数据。由于代理必须始终存在二进制文件（因为它实际上是正在执行的内容），我们应该提取有关代理的必要信息并将其与分析数据一起发送。然后，服务器可以规范化数据并在导出为 pprof 时以预期的表示形式再次返回数据。
 pprof file which it's location do not have function name, it need debuginfo to get function name.


> the server must already have the binary to be able to accept profiling data.

what do you mean ` the server must already have the binary to be able to accept profiling data.`?

In old, the server need debuginfo to tran address to function name.
Now, the server need debuginfo to compute base and  tran address to function name.

kakkoyun:
@zdyj3170101136 I'll add an explanation for this. Stay tuned.

kakkoyun:
We have an ongoing initiative the overhaul this both in agent and server. You can follow the improvements in the following RFCs.
https://docs.google.com/document/d/1QN15z1mGWZ9t5_EM66TctWR0Thou78OV54EqVNbBsFs/edit#heading=h.ajdd04yoeml7

Thanks a lot again for your contribution ❤️

PR #1216, opened 2023-01-11
fix/elfwriter:wrong first program header type null

zdyj3170101136:


pkg/elfwriter/elfwriter.go
```diff
@@ -630,9 +630,9 @@ func (w *Writer) writeSections() {
 	w.seek(w.seekSectionHeader, io.SeekStart)
 	w.u64(uint64(shoff))
 	w.seek(w.seekSectionNum, io.SeekStart)
-	w.u64(uint64(shnum)) // e_shnum
+	w.u16(uint16(shnum)) // e_shnum
```
kakkoyun, line 633: Could you please point me to the section of the documentation/specification for the ELF?

kakkoyun (commented):
Looking good.
To make a sanity check, could you add related references to the specification to the PR description?

pkg/elfwriter/elfwriter.go
zdyj3170101136: there is already have sectionHeader struct in function elfwriter.writeFileHeader.
so it is no need to add duplicate comment.
截屏2023-01-11 下午6 41 15

kakkoyun: Aha ok, apologies my mistake 🤗

kakkoyun (approved):
LGTM

PR #1221, opened 2023-01-14
pkg/debuginfo: Cache ELF hashes in ensureUploaded

marselester:
I've added caching of ELF hashes as it was recommended in the TODO comment.

```
$ go test -benchmem -run=. -bench ^BenchmarkEnsureUploadedInitiateUploadError$ ./pkg/debuginfo/
# new
BenchmarkEnsureUploadedInitiateUploadError-2   	     901	   1415440 ns/op	  422646 B/op	     613 allocs/op
# old
BenchmarkEnsureUploadedInitiateUploadError-2   	     888	   1450389 ns/op	  455509 B/op	     618 allocs/op
```

### DIFF
--- a/pkg/debuginfo/manager.go
+++ b/pkg/debuginfo/manager.go
@@ -52,7 +52,7 @@ type Manager struct {
 	stripDebuginfos bool
 	tempDir         string
 
-	// hashCache caches ELF hashes (buildID is a key).
+	// hashCache caches ELF hashes (hashCacheKey is a key).
 	hashCache                         *sync.Map
 	shouldInitiateUploadResponseCache cache.Cache
 	debuginfoSrcCache                 cache.Cache
@@ -160,6 +160,14 @@ func (di *Manager) EnsureUploaded(ctx context.Context, objFiles []*objectfile.Ma
 	}()
 }
 
+// hashCacheKey is a cache key to retrieve the hashes of debuginfo files.
+// Caching reduces allocs by 7.22% (33 kB/operation less) in ensureUpload,
+// and it shaves 4 allocs per operation.
+type hashCacheKey struct {
+	buildID string
+	modtime int64
+}
+
 func (di *Manager) ensureUploaded(ctx context.Context, objFile *objectfile.MappedObjectFile) error {
 	buildID := objFile.BuildID
 	if di.alreadyUploading(buildID) {
@@ -181,6 +189,7 @@ func (di *Manager) ensureUploaded(ctx context.Context, objFile *objectfile.Mappe
 
 	var r io.ReadSeeker
 	size := int64(0)
+	var modtime time.Time
 
 	if di.stripDebuginfos {
 		if err := os.MkdirAll(di.tempDir, 0o755); err != nil {
@@ -212,6 +221,7 @@ func (di *Manager) ensureUploaded(ctx context.Context, objFile *objectfile.Mappe
 			return fmt.Errorf("failed to stat the file: %w", err)
 		}
 		size = stat.Size()
+		modtime = stat.ModTime()
 
 		r = f
 	} else {
@@ -226,6 +236,7 @@ func (di *Manager) ensureUploaded(ctx context.Context, objFile *objectfile.Mappe
 			return fmt.Errorf("failed to stat the file: %w", err)
 		}
 		size = stat.Size()
+		modtime = stat.ModTime()
 
 		r = f
 	}
@@ -233,17 +244,21 @@ func (di *Manager) ensureUploaded(ctx context.Context, objFile *objectfile.Mappe
 	// The hash is cached to avoid re-hashing the same binary
 	// and getting to the same result again.
 	var (
+		key = hashCacheKey{
+			buildID: buildID,
+			modtime: modtime.Unix(),
+		}
 		h   string
 		err error
 	)
-	if v, ok := di.hashCache.Load(buildID); ok {
+	if v, ok := di.hashCache.Load(key); ok {
 		h = v.(string) //nolint:forcetypeassert
 	} else {
 		h, err = hash.Reader(r)
 		if err != nil {
 			return fmt.Errorf("hash debuginfos: %w", err)
 		}
-		di.hashCache.Store(buildID, h)
+		di.hashCache.Store(key, h)
 
 		if _, err = r.Seek(0, io.SeekStart); err != nil {
 			return fmt.Errorf("failed to seek to the beginning of the file: %w", err)
--- a/pkg/debuginfo/manager.go
+++ b/pkg/debuginfo/manager.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/go-kit/log"
@@ -51,6 +52,8 @@ type Manager struct {
 	stripDebuginfos bool
 	tempDir         string
 
+	// hashCache caches ELF hashes (buildID is a key).
+	hashCache                         *sync.Map
 	shouldInitiateUploadResponseCache cache.Cache
 	debuginfoSrcCache                 cache.Cache
 	uploadingCache                    cache.Cache
@@ -76,6 +79,7 @@ func New(
 		logger:          logger,
 		metrics:         newMetrics(reg),
 		debuginfoClient: debuginfoClient,
+		hashCache:       &sync.Map{},
 		shouldInitiateUploadResponseCache: cache.New(
 			cache.WithMaximumSize(512), // Arbitrary cache size.
 			cache.WithExpireAfterWrite(cacheTTL),
@@ -107,7 +111,7 @@ func New(
 // time, they will be written to the same file.
 //
 // Most of the time, the file is, erm, eventually consistent-ish,
-// and once all the writers are done, the debug file looks is an ELF
+// and once all the writers are done, the debug file looks as an ELF
 // with the correct bytes.
 //
 // However, I don't believe there's any guarantees on this, so the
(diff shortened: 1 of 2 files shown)


pkg/debuginfo/manager.go
```diff
@@ -51,6 +52,8 @@ type Manager struct {
 	stripDebuginfos bool
 	tempDir         string
 
+	// hashCache caches ELF hashes (buildID is a key).
+	hashCache                         *sync.Map
```
marselester, line 56: With a `sync.Map` the benchmark shows one less allocation comparing to `cache.Cache`.

pkg/debuginfo/manager.go
```diff
@@ -226,26 +230,34 @@ func (di *Manager) ensureUploaded(ctx context.Context, objFile *objectfile.Mappe
...
-		return fmt.Errorf("failed to seek to the beginning of the file: %w", err)
+		if _, err = r.Seek(0, io.SeekStart); err != nil {
+			return fmt.Errorf("failed to seek to the beginning of the file: %w", err)
+		}
 	}
 
 	initiateResp, err := di.debuginfoClient.InitiateUpload(ctx, &debuginfopb.InitiateUploadRequest{
 		BuildId: buildID,
-		Hash:    hash,
+		Hash:    h,
```
marselester, line 255: I was thinking to switch to highwayhash, but I wasn't sure if there is a specific reason to stick to xxhash in debuginfo package, i.e., xxhash returns strings such as `2402cb2193253c5c`, and highwayhash returns uint64.

Note, highwayhash has been used in the following files.

```
./pkg/ksym/ksym.go:	"github.com/parca-dev/parca-agent/pkg/hash"
./pkg/perf/perf.go:	"github.com/parca-dev/parca-agent/pkg/hash"
./pkg/process/maps.go:	"github.com/parca-dev/parca-agent/pkg/hash"
```

kakkoyun (commented):
There are two issues I see here:
1) The benchmark only shows the allocation; what's the effect on the CPU? Is this a bottleneck right now? Do we have any profiles to point to the problem?
2) What happens if the file associated with the build ID changes? How likely is it? Do we need to use a timestamp as part of the key?

javierhonduco:
I am unsure if the added complexity is worth it given that the performance improvement seems small 0.025% improvement. As Kemal says, if this is a very hot code path, it might be worth taking a deeper look and perhaps redesign this part of the codebase

asubiotto:
drive-by comment: @marselester, you can use the [benchstat](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat) tool to compare benchmark output `$ benchstat old.txt new.txt`. It's a little easier to interpret since it does the performance improvement math for you.

marselester:
Thank you guys for the feedback. The reason I thought the caching should be implemented here is the TODO comment I found in code. I assumed that empirically it was already discovered that binaries get often rehashed wasting CPU cycles.

```go
// TODO: We should be able to cache the hash further to avoid
// re-hashing the same binary and getting to the same result
// again.
```

> what's the effect on the CPU? Is this a bottleneck right now? Do we have any profiles to point to the problem?

Hashing isn't a bottleneck, syscalls and gc dominate the CPU profile.

<details>

```
(pprof) top
Showing nodes accounting for 8470ms, 77.56% of 10920ms total
Dropped 189 nodes (cum <= 54.60ms)
Showing top 10 nodes out of 160
      flat  flat%   sum%        cum   cum%
    5720ms 52.38% 52.38%     5720ms 52.38%  runtime/internal/syscall.Syscall6
     970ms  8.88% 61.26%      970ms  8.88%  runtime.futex
     410ms  3.75% 65.02%     1000ms  9.16%  runtime.scanobject
     350ms  3.21% 68.22%      350ms  3.21%  runtime.procyield
     250ms  2.29% 70.51%      250ms  2.29%  runtime.osyield
     210ms  1.92% 72.44%      380ms  3.48%  runtime.findObject
     150ms  1.37% 73.81%      150ms  1.37%  runtime.memclrNoHeapPointers
     150ms  1.37% 75.18%      180ms  1.65%  runtime.spanOf (inline)
     130ms  1.19% 76.37%      130ms  1.19%  runtime.heapBits.next (inline)
     130ms  1.19% 77.56%      130ms  1.19%  runtime.usleep

(pprof) top20 -cum
Showing nodes accounting for 5.91s, 54.12% of 10.92s total
Dropped 189 nodes (cum <= 0.05s)
Showing top 20 nodes out of 160
      flat  flat%   sum%        cum   cum%
         0     0%     0%      7.59s 69.51%  testing.(*B).runN
         0     0%     0%      7.56s 69.23%  github.com/parca-dev/parca-agent/pkg/debuginfo.BenchmarkEnsureUploadedInitiateUploadError
         0     0%     0%      7.54s 69.05%  github.com/parca-dev/parca-agent/pkg/debuginfo.(*Manager).ensureUploaded
         0     0%     0%      7.54s 69.05%  testing.(*B).launch
     0.02s  0.18%  0.18%      6.69s 61.26%  github.com/parca-dev/parca-agent/pkg/debuginfo.(*Extractor).Extract
     0.02s  0.18%  0.37%      5.74s 52.56%  syscall.RawSyscall6
     5.72s 52.38% 52.75%      5.72s 52.38%  runtime/internal/syscall.Syscall6
     0.01s 0.092% 52.84%      4.32s 39.56%  syscall.Syscall6
     0.01s 0.092% 52.93%      4.10s 37.55%  os.OpenFile
         0     0% 52.93%      4.09s 37.45%  os.openFileNolog
         0     0% 52.93%      3.99s 36.54%  syscall.Open (inline)
         0     0% 52.93%      3.99s 36.54%  syscall.openat
         0     0% 52.93%      3.98s 36.45%  os.Open (inline)
     0.02s  0.18% 53.11%      2.70s 24.73%  runtime.systemstack
         0     0% 53.11%      2.14s 19.60%  runtime.gcBgMarkWorker
     0.06s  0.55% 53.66%      1.77s 16.21%  os.(*file).close
         0     0% 53.66%      1.73s 15.84%  os.(*File).Close
     0.02s  0.18% 53.85%      1.71s 15.66%  syscall.Syscall
         0     0% 53.85%      1.59s 14.56%  runtime.gcBgMarkWorker.func2
     0.03s  0.27% 54.12%      1.59s 14.56%  runtime.gcDrain
```

</details>

Caching reduces allocs by 7.22% (33 kB/operation less) in ensureUpload, and it shaves 5 allocs per operation.
That should reduce GC pressure a bit.
It is not a fantastic result, so perhaps the added complexity isn't worth it as pointed out by @javierhonduco.

Here are the results including benchstat reports as advised by @asubiotto.

<details>

```
$ go test -benchmem -count=10 -run=. -bench ^BenchmarkEnsureUploadedInitiateUploadError$ ./pkg/debuginfo/ | tee bench-old.txt
$ benchstat bench-old.txt 
name                                 time/op
EnsureUploadedInitiateUploadError-2  1.43ms ± 3%

name                                 alloc/op
EnsureUploadedInitiateUploadError-2   456kB ± 0%

name                                 allocs/op
EnsureUploadedInitiateUploadError-2     618 ± 0%
$ go test -benchmem -count=10 -run=. -bench ^BenchmarkEnsureUploadedInitiateUploadError$ ./pkg/debuginfo/ | tee bench-new.txt
$ benchstat bench-new.txt
name                                 time/op
EnsureUploadedInitiateUploadError-2  1.42ms ± 6%

name                                 alloc/op
EnsureUploadedInitiateUploadError-2   423kB ± 0%

name                                 allocs/op
EnsureUploadedInitiateUploadError-2     613 ± 0%
$ benchstat bench-old.txt bench-new.txt
name                                 old time/op    new time/op    delta
EnsureUploadedInitiateUploadError-2    1.43ms ± 3%    1.42ms ± 6%    ~     (p=0.481 n=10+10)

name                                 old alloc/op   new alloc/op   delta
EnsureUploadedInitiateUploadError-2     456kB ± 0%     423kB ± 0%  -7.22%  (p=0.000 n=10+10)

name                                 old allocs/op  new allocs/op  delta
EnsureUploadedInitiateUploadError-2       618 ± 0%       613 ± 0%  -0.81%  (p=0.000 n=10+10)
```

</details>

Since I've used a 10kb file in the benchmark, I was curious how the hashing performs on bigger files.
It took only 3.38% more time to hash a 142mb file, i.e., 5.55µs vs 5.74µs.
The allocs are the same: 32.88 kB/op, 5 allocs/op.

<details>

```
$ go test -bench=Hash -count=10 ./pkg/debuginfo/ | tee bench.txt
$ benchstat bench.txt
name          time/op
Hash10kb-12   5.55µs ± 1%
Hash65mb-12   5.68µs ± 3%
Hash142mb-12  5.74µs ± 3%
$ grep Benchmark bench.txt | sed 's/Benchmark[A-z0-9]*/BenchmarkX/g' | split -l 10 -a 1 - bench_
$ benchstat bench_a bench_c
name  old time/op  new time/op  delta
X-12  5.55µs ± 1%  5.74µs ± 3%  +3.38%  (p=0.000 n=8+9)
```
</details>

<details>

<summary>hash benchmark</summary>

```go
var h string

// BenchmarkHash10kb-12    	  214392	      5514 ns/op	   32888 B/op	       5 allocs/op
func BenchmarkHash10kb(b *testing.B) {
	name := filepath.Join("../../internal/pprof/binutils/testdata", "exe_linux_64")
	f, err := os.Open(name)
	if err != nil {
		b.Fatal(err)
	}
	b.Cleanup(func() {
		f.Close()
	})

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		h, err = hash.Reader(f)
		if err != nil {
			b.Fatal(err)
		}
	}
}

// BenchmarkHash65mb-12    	  216780	      5497 ns/op	   32888 B/op	       5 allocs/op
func BenchmarkHash65mb(b *testing.B) {
	name := filepath.Join("../../dist", "parca-agent")
	f, err := os.Open(name)
	if err != nil {
		b.Fatal(err)
	}
	b.Cleanup(func() {
		f.Close()
	})

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		h, err = hash.Reader(f)
		if err != nil {
			b.Fatal(err)
		}
	}
}

// BenchmarkHash142mb-12    	  199964	      5661 ns/op	   32888 B/op	       5 allocs/op
func BenchmarkHash142mb(b *testing.B) {
	name := filepath.Join("../../dist", "parca")
	f, err := os.Open(name)
	if err != nil {
		b.Fatal(err)
	}
	b.Cleanup(func() {
		f.Close()
	})

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		h, err = hash.Reader(f)
		if err != nil {
			b.Fatal(err)
		}
	}
}
```

</details>

> What happens if the file associated with the build ID changes? How likely is it? Do we need to use a timestamp as part of the key?

The debuginfo won't be uploaded to Parca because the stale hash will match debuginfo that's already stored by Parca https://github.com/parca-dev/parca/blob/main/pkg/debuginfo/store.go#L200.

```go
initiateResp, err := di.debuginfoClient.InitiateUpload(ctx, &debuginfopb.InitiateUploadRequest{
	BuildId: buildID,
	Hash:    hash,
	Size:    size,
})
```

I think that's unlikely to happen because the linker produces a build ID that is always the same in an identical output file, though there is caveat.

> --build-id=style The md5 and sha1 styles produces an identifier that is always the same in an identical output file, but will be unique among all nonidentical output files. It is not intended to be compared as a checksum for the file’s contents. A linked file may be changed later by other tools, but the build ID bit string identifying the original linked file does not change. https://sourceware.org/binutils/docs/ld/Options.html

According to https://github.com/parca-dev/parca-agent/blob/main/pkg/buildid/buildid.go#L143 the .text section gets hashed if GNU build ID doesn't exist.

@kakkoyun idea to add a timestamp would eliminate the guesswork.

kakkoyun (approved):
LGTM  ❤️ 💚 💙 💛 💜

Issue #1157, opened 2022-12-21
metadata: Metadata association problems with children processes


kakkoyun:
Metadata association is straightforward right now (just matching exact PIDs). It causes problems associating the relevant information from service discovery (e.g. kubernetes, systemd). We need to attach all the metadata information to the processes within the encapsulator (cgroup, container, or pod). See the example below.

![](https://cdn.discordapp.com/attachments/877547706334199821/1055056854054621214/Screenshot_2022-12-21_at_10.34.13.png)

```
$ ps auxww
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
rabbitmq       1  0.0  0.0   2616  1668 ?        Ss   09:01   0:00 /bin/sh /opt/rabbitmq/sbin/rabbitmq-server
rabbitmq      11 17.2  2.6 4496400 439992 ?      Sl   09:01   9:55 /usr/local/lib/erlang/erts-12.3.2/bin/beam.smp -W w -MBas ageffcbf -MHas ageffcbf -MBlmbcs 512 -MHlmbcs 512 -MMmcs 30 -P 1048576 -t 5000000 -stbt db -zdbbl 128000 -sbwt none -sbwtdcpu none -sbwtdio none -JPperf true -B i -- -root /usr/local/lib/erlang -progname erl -- -home /var/lib/rabbitmq -- -pa  -noshell -noinput -s rabbit boot -boot start_sasl -syslog logger [] -syslog syslog_error_logger false -kernel prevent_overlapping_partitions false
```

We should have associated the metadata with the given process for this particular case.
https://github.com/parca-dev/parca-agent/blob/5a538215709969388c2fa31441bcad06e9d7b2e9/pkg/discovery/kubernetes.go#L127

However, we rely on Kubernetes API to deliver updates and heavily on caching in this path. It's possible that we might have missed some updates. It needs further investigation.

For replication:
```shell
kubectl apply -f https://github.com/rabbitmq/cluster-operator/releases/latest/download/cluster-operator.yml
kubectl apply -f https://raw.githubusercontent.com/rabbitmq/cluster-operator/main/docs/examples/hello-world/rabbitmq.yaml
```
* https://github.com/docker-library/rabbitmq/blob/master/3.11/ubuntu/Dockerfile
* https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/scripts/rabbitmq-server

kakkoyun:
[...]

PR #1232, opened 2023-01-18
metadata: Fix metadata association problems concerning children processes

kakkoyun:
Fixes https://github.com/parca-dev/parca-agent/issues/1157

In this PR, to associate the metadata for the container to all processes in the container, we traverse all the PID namespaces (type: `pid` and `pid_for_children`) and group the process by the PID namespace. And then, as a further check, we check if they belong to the same groups (CPU or systemd). You can see the result of the association in the below screenshots.

This is a stopgap solution to be able to release a patch. A long-term, cleaner solution will be provided for the main branch. It needs a little more time.

After merging this, we can release a patch version.

### Test Output

CleanShot 2023-01-18 at 18 35 48@2x
CleanShot 2023-01-18 at 18 32 59@2x

xref: https://kubernetes.io/docs/tasks/configure-pod-container/share-process-namespace/

### DIFF
--- a/pkg/discovery/discovery_manager.go
+++ b/pkg/discovery/discovery_manager.go
@@ -1,4 +1,6 @@
 // Copyright 2022 The Parca Authors
+// TODO: This license is not consistent with license used in the project.
+//       Delete the inconsistent license and above line and rerun pre-commit to insert a good license.
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
--- a/pkg/discovery/discovery_manager.go
+++ b/pkg/discovery/discovery_manager.go
@@ -1,6 +1,4 @@
 // Copyright 2022 The Parca Authors
-// TODO: This license is not consistent with license used in the project.
-//       Delete the inconsistent license and above line and rerun pre-commit to insert a good license.
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
--- a/pkg/namespace/namespace.go
+++ b/pkg/namespace/namespace.go
@@ -15,7 +15,11 @@
 package namespace
 
 import (
+	"fmt"
+	"os"
+	"path/filepath"
 	"sort"
+	"syscall"
 
 	"github.com/parca-dev/parca-agent/pkg/cgroup"
 	"github.com/prometheus/procfs"
@@ -37,6 +41,7 @@ func (n ns) Inode() uint32 {
 	return n.Namespace.Inode
 }
 
+// NamespacesForPID returns the namespaces of the process with the given PID.
 func NamespacesForPID(pid int) ([]Namespace, error) {
 	proc, err := procfs.NewProc(pid)
 	if err != nil {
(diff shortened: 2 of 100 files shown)


kakkoyun:
I officially declare my dislike of @pre-commit bot.
These auto-fixes are not productive.

pkg/namespace/namespace.go
```diff
@@ -0,0 +1,104 @@
...
+
+		cgs, err := p.Cgroups()
+		if err != nil {
+			return nil, err
+		}
+		ccg := cgroup.FindContainerGroup(cgs)
+
+		for _, namespace := range namespaces {
+			for _, inode := range nsInodes {
+				if namespace.Inode == inode && ccg.Path == cg.Path {
```
maxbrunet, line 78: Nicely done! This should handle the case where containers share the same PID namespace

https://kubernetes.io/docs/tasks/configure-pod-container/share-process-namespace/

scripts/LICENSE.txt
```diff
@@ -1,4 +1,4 @@
-Copyright 2023 The Parca Authors
+Copyright 2022 The Parca Authors
```
maxbrunet, line 1: Yes, let's keep 2022 for now, the check is not flexible, here is the related issue Lucas-C/pre-commit-hooks#23 and someone opened a PR today!

pkg/discovery/discovery_manager.go
```diff
@@ -1,4 +1,6 @@
 // Copyright 2022 The Parca Authors
+// TODO: This license is not consistent with license used in the project.
+//       Delete the inconsistent license and above line and rerun pre-commit to insert a good license.
```
maxbrunet, line 3: Never fight a bot, it does not get tired :laughing:

The TODOs from 2022 → 2023 → 2022 pass would need to be cleaned up, so the check can succeed

The autofix was added recently to assist contributors with style nits, if the majority thinks it does not help, I have no strong opinion for keeping it, although it only seems to require a bit of tuning/fixing here

maxbrunet (approved):
LGTM!

pkg/discovery/discovery_manager.go
kakkoyun: I wrote those when I was frustrated with the bot. We can keep it if we tune it. 
Let's give it more chance ☺️ 

I don't want them to fight against me when Skynet has risen.

pkg/namespace/namespace.go
```diff
@@ -0,0 +1,104 @@
...
+					break
+				}
+			}
+		}
+	}
+	pids := []int{}
+	for pid := range adjPIDs {
+		pids = append(pids, pid)
+	}
+	sort.Ints(pids)
```
javierhonduco, line 89: Why do we need to sort them?

javierhonduco (approved):
LGTM! Let's check on the demo environment to ensure that it's working as expected.

Two things that I thought of but that can be done later on:
- Was this tested in minikube, too? Just in case the groups v1 env there could bite us.
- Once our plates are less full perhaps we should add tests for `[pkg/cgroup/cgroup.go](https://github.com/parca-dev/parca-agent/pull/1232/files#diff-96eea7eb30bf90352099178ebbdf1fb0e7cb8ded3d96246b252574dc9627e15d) and other  of the files that handle this logic.

kakkoyun:
> lgtm, unless we have reports of this not working for users (or too expensive to perform) I'm not even convinced we need anything else

The operation only gets triggered when Kubernetes reschedule the pod/container. So in regular operation, it's very infrequent.

kakkoyun:
> * Once our plates are less full perhaps we should add tests for `[pkg/cgroup/cgroup.go](https://github.com/parca-dev/parca-agent/pull/1232/files#diff-96eea7eb30bf90352099178ebbdf1fb0e7cb8ded3d96246b252574dc9627e15d) and other  of the files that handle this logic.

I agree 👍 To note, none of those functions are new. Logically nothing has changed.

> * Was this tested in minikube, too? Just in case the groups v1 env there could bite us.

The screenshots from the tests that I have conducted on Minikube.

PR #1236, opened 2023-01-19
Introduce ADRs

kakkoyun:

Introducing [ADR](https://adr.github.io/)s to the project to communicate and document the design decisions we made while we built the agent.

In Polar Signals, we have used [RFC](https://en.wikipedia.org/wiki/Request_for_Comments)s heavily to document our internal decisions. However, they are not all public, and there is value in documenting certain decisions for the projects publically. Thus, I'm creating this initiative. 

I have already added some placeholders that I will fill in soon. These will be reduced versions of the RFCs we have created already.


kakkoyun:
This needs to be communicated among the maintainers. It's better to have a proposal for this type of change first. Sorry if this caused any confusion.

PR #1238, opened 2023-01-19
Eliminate CGo usage from cgroup package

kakkoyun:

We should remove as many CGo dependencies as possible. Switching from Go to C code at runtime always comes with increased costs in terms of CPU usage, so that is another reason to avoid mixed code.

### TODO
- [ ] Checkout: https://github.com/iovisor/bcc/blob/master/examples/cgroupid/cgroupid.c

### DIFF
--- a/pkg/cgroup/cgroup.go
+++ b/pkg/cgroup/cgroup.go
@@ -23,54 +23,8 @@ import (
 	"unsafe"
 
 	"github.com/prometheus/procfs"
+	"golang.org/x/sys/unix"
 )
-
-/*
-#define _GNU_SOURCE
-#include <stdlib.h>
-#include <stdio.h>
-#include <sys/types.h>
-#include <sys/stat.h>
-#include <fcntl.h>
-#include <stdint.h>
-
-struct cgid_file_handle
-{
-  //struct file_handle handle;
-  unsigned int handle_bytes;
-  int handle_type;
-  uint64_t cgid;
-};
-
-uint64_t get_cgroupid(char *path) {
-  struct cgid_file_handle *h;
-  int mount_id;
-  int err;
-  uint64_t ret;
-
-  h = malloc(sizeof(struct cgid_file_handle));
-  if (!h)
-    return 0;
-
-  h->handle_bytes = 8;
-  err = name_to_handle_at(AT_FDCWD, path, (struct file_handle *)h, &mount_id, 0);
-  if (err != 0) {
-    free(h);
-    return 0;
-  }
-
-  if (h->handle_bytes != 8) {
-    free(h);
-    return 0;
-  }
-
-  ret = h->cgid;
-  free(h);
-
-  return ret;
-}
-*/
-import "C"
 
 // FindContainerGroup returns the cgroup with the cpu controller or first systemd slice cgroup.
 func FindContainerGroup(cgroups []procfs.Cgroup) procfs.Cgroup {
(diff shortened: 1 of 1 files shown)


pkg/cgroup/cgroup.go
```diff
@@ -119,12 +73,14 @@ func PathV2AddMountpoint(path string) (string, error) {
...
-	if ret == 0 {
-		return 0, fmt.Errorf("GetCgroupID on %q failed", pathWithMountpoint)
+	hf, _, err := unix.NameToHandleAt(unix.AT_FDCWD, pathWithMountpoint, 0)
+	if err != nil {
+		return 0, fmt.Errorf("GetCgroupID on %q failed: %w", pathWithMountpoint, err)
+	}
+	if hf.Size() != 8 {
+		return 0, fmt.Errorf("GetCgroupID on %q failed: unexpected size", pathWithMountpoint)
 	}
+	ret := *(*uint64)(unsafe.Pointer(&hf.Bytes()[0]))
```
javierhonduco, line 83: Writing down some notes, as I was confused at first. Reading [the source code](https://cs.opensource.google/go/x/sys/+/master:unix/syscall_linux.go;l=2270;drc=a6f4650873b0647f4a7562d1bc45c27d4da686d7;bpv=1;bpt=1?q=NameToHandleAt&ss=go%2Fx%2Fsys) helped quite a bit 😄 .

Please feel free to correct this if I got some stuff wrong.

#### Before

We always allocated `cgid_file_handle` with `uint64_t cgid` (the handle data), which should always fit.

#### Now

[`.Bytes` + `.Type` are allocated + 32 Bytes for the handle](https://cs.opensource.google/go/x/sys/+/master:unix/syscall_linux.go;l=2279;drc=a6f4650873b0647f4a7562d1bc45c27d4da686d7;bpv=1;bpt=1?q=NameToHandleAt&ss=go%2Fx%2Fsys), which should be enough for us.

Then the `Bytes()` call that was throwing me off because thought that it was returning the whole raw buffer, including `Bytes` and `Type`, but [they are skipped](https://cs.opensource.google/go/x/sys/+/master:unix/syscall_linux.go;l=2270;drc=a6f4650873b0647f4a7562d1bc45c27d4da686d7;bpv=1;bpt=1?q=NameToHandleAt&ss=go%2Fx%2Fsys).


One more thing I wonder why they did `.Bytes +4` rather than something like the below which I think would be easier for maintenance and clearer? I guess it's due to perf reasons, but not sure.

```golang
return unsafe.Slice((*byte(unsafe.Pointer(uintptr(unsafe.Pointer(&fh.fileHandle.Bytes))+unsafe.Sizeof(&fh))), n)
```

javierhonduco (approved):
Great improvement!! 🎉

kakkoyun:
The ID extraction mechanism has issues (current and upcoming). I'm double-checking it.

javierhonduco:
The other day I forgot to suggest this, should we add this new implementation while keeping the old one and then run both and compare for mismatches? 

That way we can ensure that the new implementation behaves exactly as the current one. What do you think?

kakkoyun:
> The other day I forgot to suggest this, should we add this new implementation while keeping the old one and then run both and compare for mismatches?
> 
> That way we can ensure that the new implementation behaves exactly as the current one. What do you think?

For sure. FWIW, the previous implementation also gives an empty string. 

On the other hand, I'm also questioning the need for this. It's not used anywhere. And we don't add it as a metadata label.

brancz:
Outdated since #2958

PR #1244, opened 2023-01-23
.eh_frame: Add helpers for memory mappings

javierhonduco:
Add helpers to retrieve the information we need for native stack unwinding from process memory mappings.

This is the first step towards moving away from monolithic unwind tables to improve performance and the code: https://github.com/parca-dev/parca-agent/issues/1052.

Eventually, we might consolidate all these `proc/maps` helpers under the same package but can keep them in unwind right now as this is something we will only use here for the time being.

## Test Plan

Added unittests

```
$ sudo /usr/local/go/bin/go test -v github.com/parca-dev/parca-agent/pkg/stack/unwind
[...]
--- PASS: TestBuildUnwindTable (0.00s)
PASS
ok      github.com/parca-dev/parca-agent/pkg/stack/unwind       0.040s
```

(running under sudo to also ensure no panics/weird behaviours with all the processes of my box, in CI, these don't run as they could be flaky and not easy to repro)

### DIFF
--- a/pkg/stack/unwind/maps.go
+++ b/pkg/stack/unwind/maps.go
@@ -1,0 +1,129 @@
+// Copyright 2022 The Parca Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package unwind
+
+import (
+	"fmt"
+
+	"github.com/prometheus/procfs"
+)
+
+// ExecutableMapping represents an executable memory mapping.
+type ExecutableMapping struct {
+	LoadAddr   uint64
+	StartAddr  uint64
+	EndAddr    uint64
+	Executable string
+	mainExec   bool
+}
+
+// IsMainObject returns whether this executable is the "main executable".
+// which triggered the loading of all the other mappings.
+//
+// We care about this because if Linux ASLR is enabled, we have to
+// modify the loaded addresses for the main object.
+func (pm *ExecutableMapping) IsMainObject() bool {
+	return pm.mainExec
+}
+
+// IsJitted returns whether an executable mapping is JITed or not.
+// The detection is done by checking if the executable mapping is
+// not backed by a file.
+//
+// We don't check for the writeable flag as `mprotect(2)` may be
+// called to make it r+w only.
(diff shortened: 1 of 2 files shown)


pkg/stack/unwind/maps.go
```diff
@@ -0,0 +1,129 @@
...
+	Executable string
+	mainExec   bool
+}
+
+// MainObject returns whether this executable is the "main executable".
+// which triggered the loading of all the other mappings.
+//
+// We care about this because if Linux ASLR is enabled, we have to
+// modify the loaded addresses for the main object.
+func (pm *ExecutableMapping) MainObject() bool {
```
kakkoyun, line 37:
```suggestion
func (pm *ExecutableMapping) IsMainObject() bool {
```
Just for following the existing method naming conventions

pkg/stack/unwind/maps.go
```diff
@@ -0,0 +1,129 @@
...
+// The reason why we need to find the loaded base address is that ELF executables
+// aren't typically loaded in one large executable section, but split in several
+// mappings. For example, the .rodata section, as well as .eh_frame might go in
+// sections without executable permissions, as they aren't needed.
+func ListExecutableMappings(rawMappings []*procfs.ProcMap) ExecutableMappings {
+	result := make([]*ExecutableMapping, 0, executableMappingCount(rawMappings))
+	firstSeen := false
+	for idx, rawMapping := range rawMappings {
+		if rawMapping.Perms.Execute {
+			var LoadAddr uint64
```
kakkoyun, line 103:
```suggestion
			var loadAddr uint64
```
I don't know why the linter did not pick up these types of nits.

kakkoyun (commented):
I love this package. I'd be happy to move this directly under `proc`.

One thing that would be nice to see is the intended usage of the package. It always helps me to refine the API for the types.

pkg/stack/unwind/maps.go
javierhonduco: Good call!

javierhonduco:
> I'd be happy to move this directly under `proc`.

I would suggest keeping it here until we prove it works well. As soon as we have more use-cases within Parca Agent, we can evaluate moving it elsewhere.

> One thing that would be nice to see is the intended usage of the package. It always helps me to refine the API for the types.

These helpers are to be used in stack unwinding. Right now the logic was hacked together, not only the code is more complicated but the logic is also buggy. This new code is more correct, clearer, is tested, and provides the necessary high level helpers that we will need in per-executable unwind tables, such as `IsJitted()`

pkg/stack/unwind/maps.go
javierhonduco: Not sure but perhaps because it's in this inner scope and it being public or protected won't change its visibility?

kakkoyun: It's a local variable, yes. It's not very common to see an exported local variable. Of course, it won't change anything functionally, but it'd be nice to abide by the convention. Probably, we miss a linter configuration.

Sylfrena (approved):

PR #1261, opened 2023-01-27
Update LICENSE.txt

Sylfrena:


metalmatze (approved):

kakkoyun:
@maxbrunet was tracking a patch to fix it. Maybe we should wait for comment.

maxbrunet:
Yes, we need to add the flag `--use-current-year` in `.pre-commit-config.yaml` instead. We can leave an old date in `LICENSE.txt`

See https://github.com/Lucas-C/pre-commit-hooks/releases/tag/v1.4.1

scripts/LICENSE.txt
```diff
@@ -1,4 +1,4 @@
-Copyright 2022 The Parca Authors
+Copyright 2023 The Parca Authors
```
Sylfrena, line 1:
```suggestion
Copyright 2022 The Parca Authors
```

Sylfrena:
@maxbrunet is this what you expected? It seems doing this will require the change to _all_ files...would you like me to accept the suggestions and update the PR?

PR #1271, opened 2023-01-30
.eh_frame: unwind table per executable

javierhonduco:
This commit reworks the current DWARF-based unwinder and surrounding userspace facilities to deal with unwind tables per executable, rather than per process. It also removes the need to specify the processes names, only `--experimental-enable-dwarf-unwinding` is required.

The main benefits of this change are:

- Improved efficiency. We don't need to globally sort all the unwind entries, which in some cases resulted in a nontrivial CPU sorting millions of items. We also only produce the unwind tables for every unique executable once, and just add some metadata when a process contains a known mapping. We can completely skip the parsing of DWARF information as well as the interpretation to generate the table. This is a major win as most libraries are, well, shared, such as `libc`! :)
- Better space utilisation, as we use every single byte per shard. Before we were using a new shard for every new unwind table, without using the rest of the available space in the current shard. Most of the space was wasted before.

With this proposed approach, using ~200MB of kernel memory for the unwind table shards, we can store up to 12.5M unwind rows. Tests on my relatively busy machine showed that I only used half of the available space.

Now we have the following graph:

```
	process -> _n_ mappings -> _m_ chunks -> each stored in a shard
```

Beyond the changes above, in previous commits, the process mapping helpers got correctness improvements, while adding thorough tests and making the code simpler.

Simplifying the code and making it more correct is one of the shared goals for these changes, too.

Fixes https://github.com/parca-dev/parca-agent/issues/1052.

Additional changes
==================

Added a new clang-format configuration to allow for twice as many characters in a line to increase reliability. We all have big monitors nowadays, don't we? :D

What's not in scope
===================

- Some memory allocations are not reused, and we do them again and again, but they don't show up in profiles yet, so I haven't optimised them. Will keep an eye on these and see if this changes in production.

- Some other low-level optimisations haven't been done because they haven't been needed yet, such as improved `binary.Write`, or faster zeroing of buffers.

Future work
===========

- Continue polishing this feature;
- Properly dealing with all shards used. Right now we just clean up most of the things, but we don't ensure we give a fair chance for other processes to be profiled, too;
- More automated tests to ensure stack correctness;
- Processes with more than 120 executable mappings aren't supported yet;
- Improved metrics;
- JIT support;

Among many others!

Test plan
=========

```
$ bin/parca
```

```
$ sudo bpftool prog tracelog | tee bpf_logs.txt
```

Not unexpected errors showed up.

```
$ make && make debug/build && sudo dist/parca-agent-debug --node=test --remote-store-insecure  --remote-store-address=127.0.0.1:7070 --experimental-enable-dwarf-unwinding
```

All profiles appeared as expected with 0 reported errors.

**Python**
image

**Node.js**

image

Automated tests
===============

```
$ make test
```

```
$ make vmtest
[...
=============
Test results:
=============
- ✅ 5.4
- ✅ 5.10
- ✅ 5.18
- ✅ 5.19]
```

Both passed.

### DIFF
--- a/bpf/.clang-format
+++ b/bpf/.clang-format
@@ -1,0 +1,5 @@
+---
+BasedOnStyle: LLVM
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+ColumnLimit: 160
--- a/bpf/Makefile
+++ b/bpf/Makefile
@@ -11,7 +11,7 @@ format: c/fmt
 
 .PHONY: c/fmt
 c/fmt:
-	clang-format -i --style=LLVM $(BPF_SRC) $(BPF_HEADERS)
+	clang-format -i --style=file $(BPF_SRC) $(BPF_HEADERS)
 
 .PHONY: format-check
 format-check:
--- a/bpf/cpu/cpu.bpf.c
+++ b/bpf/cpu/cpu.bpf.c
@@ -23,18 +23,31 @@
 #define MAX_STACK_DEPTH_PER_PROGRAM 15
 // Number of BPF tail calls that will be attempted.
 #define MAX_TAIL_CALLS 10
-// Number of frames to walk in total.
+// Maximum number of frames.
 #define MAX_STACK_DEPTH 127
-// Number of stacks.
-#define MAX_STACK_TRACES 1024
+_Static_assert(MAX_TAIL_CALLS *MAX_STACK_DEPTH_PER_PROGRAM >= MAX_STACK_DEPTH, "Not enough iterations to traverse the whole stack");
+// Number of unique stacks.
+#define MAX_STACK_TRACES_ENTRIES 1024
 // Number of items in the stack counts aggregation map.
 #define MAX_STACK_COUNTS_ENTRIES 10240
+// Maximum number of processes we are willing to track.
+#define MAX_PROCESSES 5000
 // Binary search iterations for dwarf based stack walking.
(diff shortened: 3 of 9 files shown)


brancz (approved):
Incredible! I skimmed the strategy and it makes perfect sense to me, and the thorough testing also makes me confident in merging this.

Additionally here's my output from running this all day long on my home lab, which looks great! https://gist.githubusercontent.com/brancz/c107124710dad0c70ccc23f74e994556/raw/feebb9427d5f4fa01357b0c9e7978daf324c183b/gistfile1.txt

javierhonduco:
Let me add the tests and re-enable the frame pointer unwinder before merging!

javierhonduco:
Now I am thinking that perhaps the logs are too chatty, will adjust over the next couple of days.

bpf/.clang-format
```diff
@@ -0,0 +1,5 @@
+---
+BasedOnStyle: LLVM
+AllowShortIfStatementsOnASingleLine: false
```
kakkoyun, line 3: ❤️

bpf/.clang-format
```diff
@@ -0,0 +1,5 @@
...
+AllowShortLoopsOnASingleLine: false
```
kakkoyun, line 4: ❤️

kakkoyun (commented):
Awesome PR!

v-thakkar:
Absolutely amazing work :tada:

PR #1275, opened 2023-01-30
pkg/profiler/cpu: fix off-by-one in invariants check

javierhonduco:
And add more context to the assertions.


javierhonduco:
Mmm actually, will sleep on this because we only check whether the live shard is full in `PersistUnwindTable` after we append even more data to the live shard...

javierhonduco:
Either way the extra context will be useful and I believe there was still an off-by-one, so hopefully this PR can help us understand what happens in demo

brancz (approved):

PR #1276, opened 2023-01-31
Do not enable pprof profiling until BPF program is loaded

javierhonduco:
Set the pprof profile handler only once we have loaded our BPF program to avoid SIGPROFs [1] while we are loading it which results in increased loading time in the best scenario, and failure to start the Agent in the worst case.

Loading our program takes a little bit, mostly due to the verification process, and if any signals are received during that time, the kernel will abort the loading process [2] and return with -EAGAIN. Libbpf will retry up to 5 times [3], and then return the error.

1: https://github.com/golang/go/blob/2ab0e04681332c88e1bfb5fe5a72d35c1c5aae8a/src/runtime/proc.go#L4658
2: https://github.com/torvalds/linux/blob/master/kernel/bpf/verifier.c#L13793-L13794
3: https://github.com/libbpf/libbpf/blob/d73ecc91e1f9a2f2782e00f010a5a0d6abec09a4/src/bpf.h#L68-L69

Test Plan
=========

```bash
$ while true; do curl http://127.0.0.1:7071/debug/pprof/profile; done
```

No sigprofs are delivered until the BPF program is loaded (notice there are multiple BPF programs loaded by libbpf to calibrate different features):

```dtrace
# can be run with bpftrace ./file
uprobe:/proc/<PARCA_AGENT_PID>/exe:sys_bpf_prog_load { 
  @loaded=1;
  printf("BPF LOAD CALLED\n");
} 

tracepoint:signal:signal_deliver 
/comm == "parca-agent-deb" && args->sig == 27/ {  # SIGPROF
  printf("sigprof. loaded: %d (might have already been loaded)\n", @loaded); # for this to work has to be run as soon as parca agent is executed, otherwise we might miss the probe. Running it under GDB with a breakpoint in main and continuing from there does the trick
}
```

```
Attaching 2 probes...
BPF LOAD CALLED
BPF LOAD CALLED
BPF LOAD CALLED
BPF LOAD CALLED
BPF LOAD CALLED
BPF LOAD CALLED
BPF LOAD CALLED
BPF LOAD CALLED
BPF LOAD CALLED
sigprof. loaded: 1 (might have already been loaded)
sigprof. loaded: 1 (might have already been loaded)
sigprof. loaded: 1 (might have already been loaded)
sigprof. loaded: 1 (might have already been loaded)
sigprof. loaded: 1 (might have already been loaded)
```

### DIFF
--- a/cmd/parca-agent/main.go
+++ b/cmd/parca-agent/main.go
@@ -327,9 +327,25 @@ func run(logger log.Logger, reg *prometheus.Registry, flags flags) error {
 	mux.Handle("/metrics", promhttp.HandlerFor(reg, promhttp.HandlerOpts{}))
 	mux.HandleFunc("/debug/pprof/", pprof.Index)
 	mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
-	mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
 	mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
 	mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
+	// Set the pprof profile handler only once we have loaded our BPF program to avoid
+	// SIGPROFs [1] while we are loading it which results to increased loading time in
+	// the best scenario, and failure to start the Agent in the worst case.
+	//
+	// This happens because our program takes a little time to load, mostly due to the
+	// verification process, and if any signals are received during that time, the kernel
+	// will abort the loading process [2] and return with -EAGAIN. Libbpf will retry up to
+	// 5 times [3], and then return the error.
+	//
+	// - [1]: https://github.com/golang/go/blob/2ab0e04681332c88e1bfb5fe5a72d35c1c5aae8a/src/runtime/proc.go#L4658
+	// - [2]: https://github.com/torvalds/linux/blob/master/kernel/bpf/verifier.c#L13793-L13794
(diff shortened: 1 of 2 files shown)


brancz (approved):

cmd/parca-agent/main.go
```diff
@@ -327,9 +327,25 @@ func run(logger log.Logger, reg *prometheus.Registry, flags flags) error {
...
+	//
+	// This happens because our program takes a little time to load, mostly due to the
+	// verification process, and if any signals are received during that time, the kernel
+	// will abort the loading process [2] and return with -EAGAIN. Libbpf will retry up to
+	// 5 times [3], and then return the error.
+	//
+	// - [1]: https://github.com/golang/go/blob/2ab0e04681332c88e1bfb5fe5a72d35c1c5aae8a/src/runtime/proc.go#L4658
+	// - [2]: https://github.com/torvalds/linux/blob/master/kernel/bpf/verifier.c#L13793-L13794
+	// - [3]: https://github.com/libbpf/libbpf/blob/d73ecc91e1f9a2f2782e00f010a5a0d6abec09a4/src/bpf.h#L68-L69
+	bpfProgramLoaded := make(chan bool, 1)
```
kakkoyun, line 344: nit/optimization: I would have suggested using an unbuffered channel with an empty struct type and just closing the channel to signal that the program is loaded.

https://go.dev/play/p/m3vq4oltG-D

javierhonduco: Oh, didn't know about this, cool!

PR #1277, opened 2023-01-31
Discover java processes via hsperfdata

v-thakkar:
Fixes: #1306 

Java Hotspot VM provides telemetry data through jvmstat performace counters and publishes it through a memory-mapped file in a temp directory /tmp/hsperfdata_{user}/{pid}. Currently, in parca-agent there is no way to determine if a certain process is a Java process. So above naming convention will help us to find running Java processes in the system as well as add it as a metadata.

Proposed PR does the following things:
- For the given pid it searches in the global /tmp/ directory first to check in case process is being run on a host
- If not then it searches for the associated NSPid to find the hsperfdata file associated with the pid
- It also adds a new metadata label 'java', see #1366 for more details 

Output from `jps`:
Screenshot from 2023-02-27 06-24-56

Parca-agent output for the process running on host:
Screenshot from 2023-02-27 05-05-18

Parca-agent output for the process running in the container:
Screenshot from 2023-02-28 08-28-38

P.S. Applications for testing - https://github.com/v-thakkar/Java-testing

### DIFF
--- a/pkg/hsperfdata/hsperfdata.go
+++ b/pkg/hsperfdata/hsperfdata.go
@@ -1,0 +1,150 @@
+// Copyright 2022-2023 The Parca Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package hsperfdata
+
+import (
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"sync"
+
+	"github.com/go-kit/log"
+
+	"golang.org/x/sync/singleflight"
+
+	"github.com/parca-dev/parca-agent/pkg/perf"
+)
+
+const hsperfdata = "/tmp/hsperfdata_*"
+
+type Cache struct {
+	pids   map[int]struct{}
+	fs     fs.FS
+	logger log.Logger
+	mu     sync.Mutex
+	nsPID  map[int]int
+	group  singleflight.Group
+}
+
+type realfs struct{}
+
+func (f *realfs) Open(name string) (fs.File, error) {
+	return os.Open(name)
+}
+
+func NewCache(logger log.Logger) *Cache {
+	return &Cache{
+		pids:   make(map[int]struct{}),
+		fs:     &realfs{},
+		logger: logger,
+		nsPID:  map[int]int{},
+	}
+}
+
+func (c *Cache) Exists(pid int) bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	_, ok := c.pids[pid]
+	return ok
+}
+
+// IsJavaProcess returns true if the hsperfdata file exists for a given pid.
+// It first searches in all hsperfdata user directories for the processes
+// running on host and then searches in /proc/{pid}/root/tmp for processes
+// running in containers. Note that pids are assumed to be unique regardless
+// of username.
+func (c *Cache) IsJavaProcess(pid int) (bool, error) {
+	// Check if the pid is in the cache.
+	if c.Exists(pid) {
+		return true, nil
+	}
+
+	// Use singleflight to prevent concurrent requests for the same pid
+	ch := c.group.DoChan(strconv.Itoa(pid), func() (interface{}, error) {
+		// Fast path to find the processes running on the host.
+		// List all directories that match the pattern /tmp/hsperfdata_*
+		dirs, err := filepath.Glob(hsperfdata)
+		if err != nil {
+			return false, fmt.Errorf("failed to list directories: %w", err)
+		}
+
+		// Loop over all directories and search for the hsperfdata file for the given pid
+		for _, dir := range dirs {
+			hsperfdataPath := filepath.Join(dir, strconv.Itoa(pid))
+			if path, err := c.fs.Open(hsperfdataPath); err == nil {
+				defer path.Close()
+				c.pids[pid] = struct{}{}
+				return true, nil
+			}
+		}
+
+		// Slow path for the processes running in containers.
+		// Search for the pid via nspids.
+		nsPid, found := c.nsPID[pid]
+		if !found {
+			nsPids, err := perf.FindNSPIDs(c.fs, pid)
+			if err != nil {
+				if os.IsNotExist(err) {
+					return false, fmt.Errorf("%w when reading status", perf.ErrProcNotFound)
+				}
+				return false, err
+			}
+			// If we didn't find the pid in the root PID namespace, try to find it in the
+			// namespaces of the process. Store the namespace PID in the nsPID map to
+			// avoid searching for it again in the future.
+			// Note that the PID of the root PID namespace will always be the last element
+			// in the slice returned by perf.FindNSPIDs.
+			c.nsPID[pid] = nsPids[len(nsPids)-1]
+			nsPid = c.nsPID[pid]
+		}
+
+		// TODO(vthakkar): Check for the process mount point.
+		perfdataFiles := fmt.Sprintf("/proc/%d/root/tmp/", pid)
+
+		files, err := os.ReadDir(perfdataFiles)
+		if err != nil {
+			return false, fmt.Errorf("error reading %s: %w", perfdataFiles, err)
+		}
+
+		for _, f := range files {
+			if f.IsDir() {
+				if name := f.Name(); strings.HasPrefix(name, "hsperfdata") {
+					if path, err := c.fs.Open(filepath.Join(perfdataFiles, name, strconv.Itoa(nsPid))); err == nil {
+						defer path.Close()
+						c.pids[pid] = struct{}{}
+						return true, nil
+					}
+				}
+			}
+		}
+		return false, nil
+	})
+
+	result := <-ch
+	if result.Err != nil {
+		return false, result.Err
+	}
+
+	val, ok := result.Val.(bool)
+	if !ok {
+		return false, fmt.Errorf("failed to convert the result to bool")
+	}
+
+	return val, nil
+}
--- a/pkg/metadata/java_process.go
+++ b/pkg/metadata/java_process.go
@@ -1,0 +1,43 @@
+// Copyright 2022-2023 The Parca Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package metadata
+
+import (
+	"fmt"
+
+	"github.com/go-kit/log"
+	"github.com/prometheus/common/model"
+
+	"github.com/parca-dev/parca-agent/pkg/hsperfdata"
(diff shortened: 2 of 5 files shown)


pkg/hsperfdata/hsperfdata.go
```diff
@@ -0,0 +1,99 @@
...
+	"os"
+	"path/filepath"
+	"strconv"
+)
+
+//HsperfdataPath returns the path of the hsperfdata file for a given pid,
+//it searches in all hsperfdata user directories, pids are assumed to be
+//unique regardless of username.
+//TODO(vthakkar): Add tests to makes sure above statement is true especially
+//with containers
```
kakkoyun, line 29:
```suggestion
// HsperfdataPath returns the path of the hsperfdata file for a given pid,
// it searches in all hsperfdata user directories, pids are assumed to be
// unique regardless of username.
// TODO(vthakkar): Add tests to makes sure above statement is true especially
// with containers
```
🔧

pkg/hsperfdata/hsperfdata.go
```diff
@@ -0,0 +1,99 @@
...
+		}
+		filePaths[pid] = filePath
+	}
+
+	return filePaths, nil
+}
+
+// UserdataPaths returns all the java process hsperfdata paths belonging
+// to the given user
+func UserdataPaths(user string) (map[string]string, error) {
```
kakkoyun, line 67:
```suggestion
func UserDataPaths(user string) (map[string]string, error) {
```
🔧 I think this has a different pattern than `hsperfdata`, so better to stick to conventions 😄

pkg/hsperfdata/hsperfdata.go
```diff
@@ -0,0 +1,99 @@
...
+		if _, err := strconv.Atoi(f.Name()); err == nil {
+			filePaths[f.Name()] = filepath.Join(dir, f.Name())
+		}
+	}
+
+	return filePaths, nil
+}
+
+// AlldataPaths returns all users' hsperfdata path
+func AlldataPaths() (map[string]string, error) {
```
kakkoyun, line 85:
```suggestion
func AlldataPaths() (map[string]string, error) {
```

🔧 ditto

kakkoyun (commented):
❤️ 💚 💙 💛 💜

kakkoyun:
@v-thakkar, I took the liberty to copy the commit message to the PR description. Just to make my life easier 😄

pkg/hsperfdata/hsperfdata.go
v-thakkar: Yeah, it's a draft PR so no lint checks or code cleanups were done :D I generally do it after all the code works. But should probably change my editor setup, just so it's being done on save.

v-thakkar:
JFYI, haven't removed the 'draft' status just yet as there is a potential bug in the perf package's function: `FindNSPIDs`. Performing some manual tests.

v-thakkar:
Actually don't merge it yet, I ran it for a few hours and for some reason meta label values for 'java' doesn't seem to be stable. Checking it.

pkg/metadata/java_process.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+func JavaProcess(cache *hsperfdata.Cache) Provider {
+	return &StatelessProvider{"java process", func(pid int) (model.LabelSet, error) {
+		java, err := cache.IsJavaProcess(pid)
+		if err != nil {
+			return nil, fmt.Errorf("failed to determine if PID %d belongs to a java process: %w", pid, err)
+		}
+
+		return model.LabelSet{
```
kakkoyun, line 32: 🌱 Maybe we should only add the label if it's a java process. What do you think?

pkg/hsperfdata/hsperfdata.go
```diff
@@ -0,0 +1,122 @@
...
+		nsPID:  map[int]int{},
+	}
+}
+
+// IsJavaProcess returns true if the hsperfdata file exists for a given pid.
+// It first searches in all hsperfdata user directories for the processes
+// running on host and then searches in /proc/{pid}/root/tmp for processes
+// running in containers. Note that pids are assumed to be unique regardless
+// of username.
+
```
kakkoyun, line 61:
```suggestion
```

pkg/hsperfdata/hsperfdata.go
```diff
@@ -0,0 +1,122 @@
...
+type realfs struct{}
+
+func (f *realfs) Open(name string) (fs.File, error) {
+	return os.Open(name)
+}
+
+func NewCache(logger log.Logger) *Cache {
+	return &Cache{
+		pids:   make(map[int]bool),
```
kakkoyun, line 49:
```suggestion
		pids:   make(map[int]struct{}),
```
⛏ nitpick: We can reduce the memory food print just by using an empty struct to represent the existence of the pid.

pkg/hsperfdata/hsperfdata.go
```diff
@@ -0,0 +1,122 @@
...
+
+// IsJavaProcess returns true if the hsperfdata file exists for a given pid.
+// It first searches in all hsperfdata user directories for the processes
+// running on host and then searches in /proc/{pid}/root/tmp for processes
+// running in containers. Note that pids are assumed to be unique regardless
+// of username.
+
+func (c *Cache) IsJavaProcess(pid int) (bool, error) {
+	// Check if the pid is in the cache
+	c.mu.Lock()
```
kakkoyun, line 64: Since we have a lot of filesystem access in this function, I think we should create an accessor function for the cache like `Get` and use the lock in that function and not block the filesystem access.

Just food for thought: If we want to track in-flight requests and prevent duplicate processing of the same PID, we may have another map.

kakkoyun (commented):
Looks great. 

I have some minor suggestions.

pkg/metadata/java_process.go
v-thakkar: Yes, I was using java:false for some debugging and forgot to change it back.

v-thakkar:
Oops, seems like I force-pushed the one with the wrong fixup. AFK now but will push the correct one tomorrow morning.

pkg/hsperfdata/hsperfdata.go
v-thakkar: Ok, I ended up using a singleflight as discussed. Not extremely happy with the error handling for the result but that's the best I could make work. Let me know in case you've a more sophisticated suggestion for it :smile:

v-thakkar:
Ran this successfully for a few hours. Metadata values are stable now.

pkg/hsperfdata/hsperfdata.go
```diff
@@ -0,0 +1,144 @@
...
+				if name := f.Name(); strings.HasPrefix(name, "hsperfdata") {
+					if _, err := c.fs.Open(filepath.Join(perfdataFiles, name, strconv.Itoa(nsPid))); err == nil {
+						c.pids[pid] = struct{}{}
+						return true, nil
+					}
+				}
+			}
+		}
+
+		c.pids[pid] = struct{}{}
```
javierhonduco, line 129: This case is for when we can't find the hsperf file, right? In this case why do we need to add this pid to the "is pid JAVA" cache?

pkg/hsperfdata/hsperfdata.go
```diff
@@ -0,0 +1,144 @@
...
+	ch := c.group.DoChan(strconv.Itoa(pid), func() (interface{}, error) {
+		// List all directories that match the pattern /tmp/hsperfdata_*
+		dirs, err := filepath.Glob(hsperfdata)
+		if err != nil {
+			return false, fmt.Errorf("failed to list directories: %w", err)
+		}
+
+		// Loop over all directories and search for the hsperfdata file for the given pid
+		for _, dir := range dirs {
+			hsperfdataPath := filepath.Join(dir, strconv.Itoa(pid))
```
javierhonduco, line 88: Curious why don't we directly traverse `/proc/<pid>/root/hsperfdata` directly as it's done below. Is it for some edge case or optimisation?

pkg/hsperfdata/hsperfdata.go
```diff
@@ -0,0 +1,144 @@
...
+		return false, nil
+	})
+
+	result := <-ch
+	if result.Err != nil {
+		return false, result.Err
+	}
+
+	val, ok := result.Val.(bool)
+	if !ok {
```
javierhonduco, line 139: Nice error handling!

pkg/hsperfdata/hsperfdata.go
```diff
@@ -0,0 +1,144 @@
...
+		// List all directories that match the pattern /tmp/hsperfdata_*
+		dirs, err := filepath.Glob(hsperfdata)
+		if err != nil {
+			return false, fmt.Errorf("failed to list directories: %w", err)
+		}
+
+		// Loop over all directories and search for the hsperfdata file for the given pid
+		for _, dir := range dirs {
+			hsperfdataPath := filepath.Join(dir, strconv.Itoa(pid))
+			if _, err := c.fs.Open(hsperfdataPath); err == nil {
```
javierhonduco, line 89: Should we close the file handles? While the Go runtime will close them when running the finalizer I think an explicit close might result in clearer code and less work for the GC

pkg/hsperfdata/hsperfdata.go
v-thakkar: It's the optimization. That way is mainly for the processes running in containers. This is a fast path for the processes running on hosts. We don't need to go for the longer route of reading /proc/%d/status and finding nspids etc if the process is running on the host.

pkg/hsperfdata/hsperfdata.go
v-thakkar: Oops, good catch. That was some personal debugging leftover :see_no_evil: Will wait for Kemal's review and push the fixes after that. Thanks for the review!

pkg/hsperfdata/hsperfdata.go
javierhonduco: That makes sense, thanks for explaining. Perhaps a code comment to clarify this could be useful 😄

pkg/hsperfdata/hsperfdata.go
```diff
@@ -0,0 +1,144 @@
...
+	return &Cache{
+		pids:   make(map[int]struct{}),
+		fs:     &realfs{},
+		logger: logger,
+		nsPID:  map[int]int{},
+	}
+}
+
+func (c *Cache) Get(pid int) bool {
+	c.mu.Lock()
```
kakkoyun, line 60: ❤️

pkg/hsperfdata/hsperfdata.go
kakkoyun: Exactly this ☝️

kakkoyun (commented):
LGTM ❤️ 💚 💙 💛 💜

Awesome work! There are a few minor things left and then we can merge.

pkg/hsperfdata/hsperfdata.go
```diff
@@ -0,0 +1,146 @@
...
+		if !found {
+			nsPids, err := perf.FindNSPIDs(c.fs, pid)
+			if err != nil {
+				if os.IsNotExist(err) {
+					return false, fmt.Errorf("%w when reading status", perf.ErrProcNotFound)
+				}
+				return false, err
+			}
+
+			c.nsPID[pid] = nsPids[len(nsPids)-1]
```
javierhonduco, line 109: Could you add a comment explaining that we expect to find the PID in the non-root PID namespace in the last element?

Another thing I am thinking right now, is what about more than 1 level of nesting, would we want to cover that?

javierhonduco (approved):
LGTM! Good stuff!

pkg/hsperfdata/hsperfdata.go
v-thakkar: Yes, ideally we want to cover more than 1 level of nesting. But currently, we don't handle it in perf-map-agent support either. I am planning to add it in ~2 weeks once the first version of Java support is out.

I have an idea for supporting it by using the /proc/<pid>/ns directory to read the file descriptors corresponding to each namespace of the process and then using the setns system call to switch to each namespace in turn and search for the hsperfdata file. But it needs more testing. (or maybe there is a simpler way to do this - either way let's discuss offline) :)

P.S. Added the comment. :)

v-thakkar:
Merging this now but happy to improve the added comment around how we're handling nsPID map in the follow-up PR, in case there are any more suggestions. Also as discussed on Discord, we'll handle the case of nested namespaces as part of the future work.

PR #1287, opened 2023-02-01
Frame pointer detection caching

javierhonduco:
As of right now it uses 50% of CPU cycles in Demo. We only saw it there as
it seems that the disks we use there are very slow.

Test Plan
=========
Ran the agent for a while with zero issues. Let's push this change to
the Demo environment and ensure that we are getting the performance wins
we expect.

Running locally:

image

### DIFF
--- a/pkg/stack/unwind/executable.go
+++ b/pkg/stack/unwind/executable.go
@@ -16,12 +16,84 @@ package unwind
 
 import (
 	"debug/elf"
+	"errors"
 	"fmt"
+	"os"
 	"strings"
+	"syscall"
 
+	burrow "github.com/goburrow/cache"
 	"github.com/hashicorp/go-version"
 	"github.com/xyproto/ainur"
 )
+
+type FramePointerCache struct {
+	cache burrow.Cache
+}
+
+// The inode value can be recycled (this behavior is filesystem specific)
+// and it's only guaranteed to be unique within each filesystem. By adding
+// the change time, which gets updated every time the file or its medatadata
+// are modified + the inode we significantly reduce the chances of collisions.
+//
+// Replacing the file in-place might result in the same inode being used, but
+// the change time will most likely be different.
+type framePointerCacheKey struct {
+	inode        uint64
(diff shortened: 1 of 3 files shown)


pkg/stack/unwind/executable.go
```diff
@@ -16,13 +16,73 @@ package unwind
...
+type FramePointerCache struct {
+	cache burrow.Cache
+}
+
+// the inode value can be recycled (this is implementation specific)
+// and it's only unique across filesystems. By adding the creation time
+// to the cache key we significantly reduce the chances of producing wrong
+// results.
+type framePointerCacheKey struct {
+	inode        uint64
```
javierhonduco, line 39: Didn't add the path as it's more storage than we need, but also because I don't think it gives us any advantages. Files can be renamed, symlinks should be handled, etc

pkg/stack/unwind/executable.go
```diff
@@ -16,13 +16,73 @@ package unwind
...
+		return false, err
+	}
+
+	fpc.cache.Put(cacheKey, val)
+	return false, nil
+}
+
+func NewHasFramePointersCache() FramePointerCache {
+	return FramePointerCache{
+		cache: burrow.New(burrow.WithMaximumSize(10_000)), // 3 * 64bit * 10k (excluding metadata)
```
javierhonduco, line 82: Note to self: check the expiry policies and select one that works for us

brancz (approved):
Looks good to me. I think it’s worth commenting extra what the case is that we would miss, eg. the same file modified in place, which I’ve never heard of and would find questionable at best, but I suppose theoretically possible?

pkg/stack/unwind/executable.go
javierhonduco: Sorry, this was just a quick draft, it was the size of the value of the hashmap alone, there is 1 `uint64` and 2 `int64` from `syscall.Timespec`. I'm going to update this with more accurate numbers

javierhonduco:
Just updated the comments to be more accurate

PR #1292, opened 2023-02-02
Ignore contents of the section if its type is SHT_NOBITS

kakkoyun:

### DIFF
--- a/pkg/elfwriter/helpers.go
+++ b/pkg/elfwriter/helpers.go
@@ -23,6 +23,15 @@ import (
 	"runtime/debug"
 )
 
+type zeroReader struct{}
+
+func (*zeroReader) ReadAt(p []byte, off int64) (n int, err error) {
+	for i := range p {
+		p[i] = 0
+	}
+	return len(p), nil
+}
+
 func copyCompressed(w io.Writer, r io.Reader) (uint64, uint64, error) {
 	if r == nil {
 		return 0, 0, errors.New("reader is nil")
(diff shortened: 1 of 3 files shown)


brancz (approved):

pkg/elfwriter/helpers.go
```diff
@@ -23,6 +23,15 @@ import (
 	"runtime/debug"
 )
 
+type zeroReader struct{}
+
+func (*zeroReader) ReadAt(p []byte, off int64) (_ int, _ error) {
```
javierhonduco, line 28:
```suggestion
func (*zeroReader) ReadAt(p []byte, off int64) (int, error) {
```

nit

pkg/elfwriter/helpers.go
```diff
@@ -23,6 +23,15 @@ import (
...
+	for i := range p {
+		p[i] = 0
```
javierhonduco, line 30: I find it a bit odd that a reader is modifying a slice, but I guess we do this to avoid allocations, right?

javierhonduco (approved):

javierhonduco:
I am a bit late to the party but perhaps we could add more context to the PR linking to the Go commit that changed this behaviour

kakkoyun:
> I am a bit late to the party but perhaps we could add more context to the PR linking to the Go commit that changed this behaviour

Sorry for hastily merging it. Will add the link and do the fixes later.

pkg/elfwriter/helpers.go
kakkoyun: It's more to avoid using wrong bytes if the slice somehow has data.

pkg/elfwriter/helpers.go
kakkoyun: 😅 Good catch.

Issue #688, opened 2022-08-08
metadata: Implement a more scalable solution for Kubernetes metadata discovery


kakkoyun:
We talked directly to the Kubernetes API using a list watcher to discover metadata labels for containers that we discover. However, this strategy is scalable until a limit of nodes in the cluster. One thing we can try is to talk to the kubelet on the node instead of Kubernetes API. But you can't have watchers on the kubelet API, so we need to implement polling a hammer the kubelet API which is not a perfect solution. So what we can try is to use container runtime hooks to enrich the discovered binary metadaata.

https://github.com/kinvolk/inspektor-gadget/tree/main/gadget-container/hooks

We only need to invest in this when we have scalability problems regarding Kubernetes metadata discovery but I'd like to add this for future reference.

cc @brancz

PR #1303, opened 2023-02-06
pkg/metadata: Add PodHosts provider to provide pod_ip and pod

zdyj3170101136:
it is not always ok to let parca-agent query k8s directly to connect pid and pod。

so we should provide pid's pod name if possible.

### DIFF
--- a/pkg/metadata/pod_hosts.go
+++ b/pkg/metadata/pod_hosts.go
@@ -1,3 +1,16 @@
+// Copyright 2023 The Parca Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package metadata
 
 import (
@@ -11,9 +24,7 @@ import (
 	"github.com/prometheus/common/model"
 )
 
-var (
-	hostname string
-)
+var hostname string
 
 func init() {
 	local, err := getExternalIpAndHost(1)
@@ -24,7 +35,7 @@ func init() {
 }
 
 // PodHosts provide host ip in default
-// and will provide pod_ip and pod if pid is a pod
+// and will provide pod_ip and pod if pid is a pod.
 func PodHosts() Provider {
 	return &StatelessProvider{"hosts", func(pid int) (model.LabelSet, error) {
 		e, err := getExternalIpAndHost(pid)
--- a/pkg/metadata/pod_hosts.go
+++ b/pkg/metadata/pod_hosts.go
@@ -20,25 +20,29 @@ import (
 	"io"
 	"os"
 	"strings"
+	"sync"
 
 	"github.com/prometheus/common/model"
 )
 
-var hostname string
-
-func init() {
-	local, err := getExternalIpAndHost(1)
-	if err != nil {
-		panic(err)
-	}
-	hostname = local.hostname
-}
+var (
+	initLocalHostname sync.Once
+	hostname          string
+)
 
 // PodHosts provide host ip in default
 // and will provide pod_ip and pod if pid is a pod.
 func PodHosts() Provider {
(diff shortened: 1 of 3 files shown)


pkg/metadata/pod_hosts.go
```diff
@@ -0,0 +1,98 @@
...
+	ip       string
+	hostname string
+}
+
+func getExternalIPAndHost(pid int) (*hostEntry, error) {
+	data, err := os.ReadFile(fmt.Sprintf("/proc/%d/root/etc/hosts", pid))
+	if err != nil {
+		return nil, err
+	}
+	hostEntrys, err := parseHosts(bytes.NewReader(data))
```
kakkoyun, line 70:
```suggestion
	hostEntries, err := parseHosts(bytes.NewReader(data))
```
⛏ Spelling

pkg/metadata/pod_hosts.go
```diff
@@ -0,0 +1,98 @@
...
+
+// PodHosts provide host ip in default
+// and will provide pod_ip and pod if pid is a pod.
+func PodHosts() Provider {
+	return &StatelessProvider{"hosts", func(pid int) (model.LabelSet, error) {
+		initLocalHostname.Do(func() {
+			local, err := getExternalIPAndHost(1)
+			if err != nil {
+				// this should not happen
+				panic(err)
```
kakkoyun, line 41: 🔧  We can return the error. No need to panic. Metadata providers should work in the best-effort mode.

pkg/metadata/pod_hosts.go
```diff
@@ -0,0 +1,98 @@
...
+		return nil, err
+	}
+	hostEntrys, err := parseHosts(bytes.NewReader(data))
+	if err != nil {
+		return nil, err
+	}
+	if len(hostEntrys) == 0 {
+		return nil, fmt.Errorf("read hosts file result nil, raw:%s", string(data))
+	}
+	return &hostEntrys[len(hostEntrys)-1], nil
```
kakkoyun, line 77: Why do we return the pointer here? What's the advantage?

For consistency, either we should stick to the pointers even in the slice or the values. I'd vote for the values since we don't change them at all.

pkg/metadata/pod_hosts.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func parseHosts(r io.Reader) ([]hostEntry, error) {
```
kakkoyun, line 80:
```suggestion
func parseHosts(r io.Reader, entries []hostEntry)  error {
```

This function can take the resulting slice from the caller to avoid additional escape to the heap.

kakkoyun (commented):
This is a great additon. Thanks for the PR.

I have a couple of suggestions, and then we can merge the PR.

pkg/metadata/pod_hosts.go
zdyj3170101136: Because if return pointer,
Then the return statement could write like `return nil, err ` directly.

I will fix it.

pkg/metadata/pod_hosts.go
zdyj3170101136: Do you mean prealloc entries?
Like parseHosts(r, make([]entries, 0, 3))?

But it is not easy to know the entries cap.

zdyj3170101136:
> 这是一个很好的补充。感谢您的公关。
> 
> 我有几个建议，然后我们可以合并 PR。

in our company, it is not allowed to query k8s directly in agent.
the people who run k8s provide an api about pod name and it's info.

in agent, i use this to connect pid and pod.
and in parca server, i use the pod name to query the api, connnect pid and pod info.

kakkoyun:
> > 这是一个很好的补充。感谢您的公关。
> > 我有几个建议，然后我们可以合并 PR。
> 
> in our company, it is not allowed to query k8s directly in agent. the people who run k8s provide an api about pod name and it's info.
> 
> in agent, i use this to connect pid and pod. and in parca server, i use the pod name to query the api, connnect pid and pod info.

I totally get it. We plan to talk directly to `kubelet` to discover all the metadata. (this is somewhat related https://github.com/parca-dev/parca-agent/issues/688). We'll get to it eventually.

pkg/metadata/pod_hosts.go
kakkoyun: Never mind, it's a micro-optimization.
But this is what I've meant 👉 https://blog.devgenius.io/in-depth-analysis-of-golang-memory-escape-edfbfb856913

kakkoyun (approved):

kakkoyun (approved):

brancz:
Clever!

PR #1307, opened 2023-02-06
ci: More unsafe directory fixes

javierhonduco:
To fix the broken CI (https://github.com/parca-dev/parca-agent/pull/1287)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -53,6 +53,9 @@ jobs:
       - name: Check out the code
         uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
 
+      - name: Add repository directory to the git global config as a safe directory
+        run: git config --global --add safe.directory /__w/parca-agent/parca-agent
+
       - name: Set up Go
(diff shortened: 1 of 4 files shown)


kakkoyun (approved):

.github/workflows/build.yml
```diff
@@ -53,6 +53,9 @@ jobs:
       - name: Check out the code
         uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
 
+      - name: Add repository directory to the git global config as a safe directory
+        run: git config --global --add safe.directory /__w/parca-agent/parca-agent
```
maxbrunet, line 57: This one should not be necessary, I've only seen affecting job containers so far

maxbrunet (commented):
Thanks for doing that, as Git get upgraded in the container images we see more of these, I'm going to add a few comments for the reason

.github/workflows/build.yml
maxbrunet: The snap workflow can use one tho

javierhonduco: Thanks!!

PR #1312, opened 2023-02-07
Improve performance of BPF map buffer writes

javierhonduco:
Using `binary.Write` to write to memory buffers we use to update BPF maps is not only using quite a bit of CPU cycles but also incurs quite a bit of small allocations. In particular, [every single write performs an allocation](https://github.com/golang/go/blob/02d8ebda8398342516a24a3e8435d527ed156cab/src/encoding/binary/binary.go#L345).

This commit reduces allocations to 0 and CPU cycles by ~14x by writing to the buffer without performing any allocations. We do this with a new type, defined over a byte slice, which directly writes to the buffer using the lower-level `binary` APIs.

This also introduces two other advantages:
- The underlying array backing the slice will never grow in case of bugs. We want the buffers to be static and of exactly the size we expect in BPF.
- We can zero the buffers more efficiently. We need to do this to ensure that there is no stale information in the buffers we reuse.

This new type has to be used with care as we expect the `Slice` method to request a size equal or smaller to the available capacity.

Test Plan
=========

Added unit tests, ran locally for a while with no issues and `binary.Write` barely showing up in the profiles. The new methods don't take much CPU either.

```
$ go test -v ./pkg/profiler/ -bench=. -count=30 > bench.txt
```

```
$ benchstat bench.txt
name                          time/op
EfficientBufferSliceWrite-12  2.18ns ± 4%
BinaryWrite-12                31.3ns ±26%

name                          alloc/op
EfficientBufferSliceWrite-12   8.00B ± 0%
BinaryWrite-12                 16.0B ± 0%

name                          allocs/op
EfficientBufferSliceWrite-12    0.00
BinaryWrite-12                  1.00 ± 0%
```

### DIFF
--- a/pkg/profiler/buffer.go
+++ b/pkg/profiler/buffer.go
@@ -1,0 +1,69 @@
+// Copyright 2022-2023 The Parca Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package profiler
+
+import "encoding/binary"
+
+type EfficientBuffer []byte
+
+// Slice returns a slice re-sliced from the original EfficientBuffer.
+// This is useful to efficiently write byte by byte, for example, when
+// setting BPF maps without incurring in extra allocations in the writing
+// methods, or changing the capacity of the underlying memory buffer.
+//
+// Callers are responsible to ensure that there is enough capacity left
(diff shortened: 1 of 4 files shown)


pkg/profiler/buffer.go
```diff
@@ -0,0 +1,59 @@
...
+	newSize := len(*eb) + size
+	subSlice := (*eb)[len(*eb):newSize]
+	// Extend its length.
+	*eb = (*eb)[:newSize]
+	return subSlice
+}
+
+func (eb *EfficientBuffer) PutUint64(v uint64) {
+	binary.LittleEndian.PutUint64((*eb)[:8], v)
+	*eb = (*eb)[8:]
```
brancz, line 38: This is clever, but different from what people are likely used to from the `binary` package. I would suggest to write a function comment that notes that the function writes the uint64 in little endian _and_ advances the buffer.

brancz (approved):
I think I would have personally wrapped the byte slice in a struct and kept a current index instead (I just find that a little easier to reason about), but this works perfectly well!

kakkoyun (commented):
❤️ 💚 💙 💛 💜

PR #1316, opened 2023-02-08
pkg/profiler: Batch upload all prof's objFile

zdyj3170101136:
#### problem
sometimes i find it take a lot of time between agent get profile and server get debuginfo.

#### reason
the agent upload each pid profile's objectfile to parca.
and the debuginfomanager use singleflight to make sure that only one debuginfoManager.EnsureUploaded is running.

if the exists api cost 10x ms, it will make the agent only upload one pid's objectfile.

#### what this pr do?
it upload all pid's objectfile to parca.
which makesure that all pid's objectfile will uploaded.


kakkoyun (approved):
LGTM. Thanks ❤️

PR #1317, opened 2023-02-08
pkg/profiler/cpu: Ignore non-existent keys

javierhonduco:
Trying to delete a key that doesn't exist fails with `ENOENT`. This might happen due to race conditions. For example, stacks keep on being written while we list and remove them.

In the future, we might want to have a better mechanism to prevent races from occurring in the first place, but let's ensure that we progress and continue removing items from the map if a key is not present.

If we don't do this, we might leave the map filled with older data, and can produce broken profiles.

This commit also moves the common listing+removing functionality to its own function

Test Plan
=========

Run the agent for >1h with DWARF unwinding and several hot processes without issues:

image


kakkoyun (approved):
LGTM  ❤️ 💚 💙 💛 💜

PR #1318, opened 2023-02-08
Add synthetic unwind row entry for the function end

kakkoyun:
Context
=======

DWARF unwind information specifies the low and high PC for every
function it covers. Within a function, it has entries for each
instruction boundary. These entries have the program counter they
start covering, but not its end.

This is not a problem for any instruction boundary except the last one,
but while we have that information when parsing the DWARF info, we never
add it anywhere in the BPF representation.

While this was not a problem for a large majority of the stacks we need
to unwind it introduce two issues:

- We know we've reached the bottom frame when a PC doesn't have unwind
  info + frame pointer == 0. Normally the bottom frame has a PC that's
not covered by the (lowest, highest) PCs that the unwind info has, but
as far as I know there's no guarantee of this. If we get the last item
of a chunk, we have no way to know if the PC we are interested in is
covered in the unwind info, or not;
- The other problem is that we currently have, related to the last point
  above, is that we have to chunk some unwind tables when they are too
large for the maximum size of a shard or for the current free space in a
shard. When this happens we might split a function in two. This causes
the same issues as above at every split point.

Proposed solution
=================

This commit adds a fake or synthetic unwind row to indicate the end of a
function. We then handle this appropriately in the unwinder.

Unfortunately we will increase the unwind table size, I expect by ~5-10% on
average, but haven't measured properly. While this is not ideal, we want to
ensure that the statistics we get from the unwinder as are correct as possible.

In the future we might remove some of these markers that aren't totally
necessary (e.g. if we can prove that there can't be any other PC in between
a pair of consecutive rows from different functions. This might be hard
to get 100% right on x86, but could be a possibility).

I was planning on slightly reducing the size of the unwind row, which
should help a bit with the recent size increase from this commit.

Other changes
=============

- Made the code that deals with chunking and shard allocation cleaner an
more correct.
- The `eh-frame` utility accepts a new argument to filter unwind info
  for a particular function that contains the given PC.

Test Plan
=========

image

+ no errors on `sudo bpftool prog tracelog | grep err` 

Co-authored-by: Kemal Akkoyun <kakkoyun@users.noreply.github.com>
Co-authored-by: Javier Honduvilla Coto <javierhonduco@gmail.com>

### DIFF
--- a/bpf/cpu/cpu.bpf.c
+++ b/bpf/cpu/cpu.bpf.c
@@ -58,6 +58,7 @@ _Static_assert(1 << MAX_BINARY_SEARCH_DEPTH >= MAX_UNWIND_TABLE_SIZE, "Unwind ta
 #define CFA_TYPE_RBP 1
 #define CFA_TYPE_RSP 2
 #define CFA_TYPE_EXPRESSION 3
+#define CFA_TYPE_END_OF_FDE_MARKER 4
 
 // Values for the unwind table's frame pointer type.
 #define RBP_TYPE_UNCHANGED 0
@@ -66,6 +67,7 @@ _Static_assert(1 << MAX_BINARY_SEARCH_DEPTH >= MAX_UNWIND_TABLE_SIZE, "Unwind ta
 #define RBP_TYPE_EXPRESSION 3
 
 // Binary search error codes.
+#define BINARY_SEARCH_DEFAULT 0xFAFAFAFA
 #define BINARY_SEARCH_NOT_FOUND 0xFABADA
 #define BINARY_SEARCH_SHOULD_NEVER_HAPPEN 0xDEADBEEF
 #define BINARY_SEARCH_EXHAUSTED_ITERATIONS 0xBADFAD
@@ -121,7 +123,6 @@ typedef struct shard_info {
 
 // Unwind table shards for an executable mapping.
 typedef struct stack_unwind_table_shards {
-  u64 len;
   shard_info_t shards[MAX_UNWIND_TABLE_CHUNKS];
 } stack_unwind_table_shards_t;
 
@@ -331,7 +332,7 @@ static __always_inline void *bpf_map_lookup_or_try_init(void *map, const void *k
 // Binary search the unwind table to find the row index containing the unwind
 // information for a given program counter (pc).
 static u64 find_offset_for_pc(stack_unwind_table_t *table, u64 pc, u64 left, u64 right) {
-  u64 found = BINARY_SEARCH_NOT_FOUND;
+  u64 found = BINARY_SEARCH_DEFAULT;
 
   for (int i = 0; i < MAX_BINARY_SEARCH_DEPTH; i++) {
     // TODO(javierhonduco): ensure that this condition is right as we use
@@ -351,9 +352,8 @@ static u64 find_offset_for_pc(stack_unwind_table_t *table, u64 pc, u64 left, u64
     }
 
     // Debug logs.
-    // bpf_printk("\t-> fetched PC %llx, target PC %llx (iteration %d/%d, mid:
-    // %d, left:%d, right:%d)", ctx->table->rows[mid].pc, ctx->pc, index,
-    // MAX_BINARY_SEARCH_DEPTH, mid, ctx->left, ctx->right);
+    // bpf_printk("\t-> fetched PC %llx, target PC %llx (iteration %d/%d, mid: %d, left:%d, right:%d)", table->rows[mid].pc, pc, i, MAX_BINARY_SEARCH_DEPTH,
+    // mid, left, right);
     if (table->rows[mid].pc <= pc) {
       found = mid;
       left = mid + 1;
@@ -464,19 +464,20 @@ static __always_inline enum find_unwind_table_return find_unwind_table(shard_inf
     return FIND_UNWIND_SHARD_NOT_FOUND;
   }
 
+  u64 adjusted_pc = pc - load_address;
   for (int i = 0; i < MAX_UNWIND_TABLE_CHUNKS; i++) {
-    if (i > shards->len) {
-      return FIND_UNWIND_SHARD_EXHAUSTED_SEARCH;
+    // Reached last chunk.
+    if (shards->shards[i].low_pc == 0) {
+      break;
     }
-
-    if (shards->shards[i].low_pc <= pc - load_address && pc - load_address <= shards->shards[i].high_pc) {
+    if (shards->shards[i].low_pc <= adjusted_pc && adjusted_pc <= shards->shards[i].high_pc) {
       bpf_printk("[info] found shard");
       *shard_info = &shards->shards[i];
       return FIND_UNWIND_SUCCESS;
     }
   }
 
-  bpf_printk("[error] could not find the right shard...");
+  bpf_printk("[error] could not find shard");
   return FIND_UNWIND_SHARD_NOT_FOUND;
 }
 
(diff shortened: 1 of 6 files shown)


bpf/cpu/cpu.bpf.c
```diff
@@ -121,7 +123,6 @@ typedef struct shard_info {
 
 // Unwind table shards for an executable mapping.
 typedef struct stack_unwind_table_shards {
-  u64 len;
```
javierhonduco, line 124: We don't need this anymore. We traverse all the chunks until one matches. Calculating the total amount of chunks ahead of time complicates things and doesn't yield any tangible benefit.

bpf/cpu/cpu.bpf.c
```diff
@@ -655,7 +668,7 @@ int walk_user_stacktrace_impl(struct bpf_perf_event_data *ctx) {
 
       previous_rsp = unwind_state->sp + 8 + ((((unwind_state->ip & 15) >= threshold)) << 3);
     } else {
-      bpf_printk("\t[error] register %d not valid (expected $rbp or $rsp)", found_cfa_type);
+      bpf_printk("\t[warn] register %d not valid (expected $rbp or $rsp)", found_cfa_type);
```
javierhonduco, line 671: "downgraded" these two logs as they are well-known limitations of our unwinder. They happen rarely enough too.

pkg/profiler/cpu/maps.go
```diff
@@ -89,11 +89,10 @@ const (
 		} shard_info_t;
 
 		typedef struct stack_unwind_table_shards {
-			u64 len;
 			shard_info_t shards[MAX_UNWIND_TABLE_CHUNKS];
 		} stack_unwind_table_shards_t;
 	*/
-	unwindShardsSizeBytes     = 8 + (maxUnwindTableChunks * 8 * 5)
```
javierhonduco, line 96: One less 64 bits word.

pkg/profiler/cpu/maps.go
```diff
@@ -799,27 +818,56 @@ func (m *bpfMaps) setUnwindTableForMapping(buf *profiler.EfficientBuffer, pid in
...
-		// available space.
-		if m.highIndex%maxUnwindTableSize == 0 && uint64(len(ut)) > m.availableEntries() {
-			numShards++
-		}
+		chunkIndex := 0
 
-		// .len
-		if err := binary.Write(unwindShardsValBuf, m.byteOrder, uint64(numShards)); err != nil {
-			return fmt.Errorf("write shards .len bytes: %w", err)
-		}
```
javierhonduco, line 818: The previous logic was slightly incorrect, besides being quite repetitive. Now things should be improved!

bpf/cpu/cpu.bpf.c
```diff
@@ -517,7 +518,7 @@ static __always_inline void add_stack(struct bpf_perf_event_data *ctx, u64 pid_t
     }
 
   } else if (method == STACK_WALKING_METHOD_FP) {
-    bpf_printk("[info] fp unwinding %d", DISABLE_BPF_HELPER_FP_UNWINDER);
+    // bpf_printk("[info] fp unwinding %d", DISABLE_BPF_HELPER_FP_UNWINDER);
```
kakkoyun, line 521: ❓ Was this noisy?

bpf/cpu/cpu.bpf.c
```diff
@@ -58,6 +58,7 @@ _Static_assert(1 << MAX_BINARY_SEARCH_DEPTH >= MAX_UNWIND_TABLE_SIZE, "Unwind ta
 #define CFA_TYPE_RBP 1
 #define CFA_TYPE_RSP 2
 #define CFA_TYPE_EXPRESSION 3
+#define CFA_TYPE_END_OF_FDE_MARKER 4
```
kakkoyun, line 61: 😃 👍

cmd/eh-frame/main.go
```diff
@@ -27,6 +27,7 @@ import (
 type flags struct {
 	Executable string `kong:"help='The executable to print the .eh_unwind tables for.'"`
 	Compact    bool   `kong:"help='Whether to use the compact format.'"`
+	RelativePc uint64 `kong:"help=''"`
```
kakkoyun, line 30: ⛏ nitpick
```suggestion
	RelativePC uint64 `kong:"help=''"`
```

pkg/profiler/cpu/maps.go
```diff
@@ -847,8 +890,8 @@ func (m *bpfMaps) setUnwindTableForMapping(buf *profiler.EfficientBuffer, pid in
 
 			// Dealing with the first chunk, we must add the lowest known PC.
 			minPc := currentChunk[0].Pc()
```
kakkoyun, line 892: ⛏ nitpick:
```suggestion
			minPC := currentChunk[0].PC()
```

pkg/stack/unwind/compact_unwind_table.go
```diff
@@ -73,6 +74,10 @@ func (cutr *CompactUnwindTableRow) RbpOffset() int16 {
 	return cutr.rbpOffset
 }
 
+func (cutr *CompactUnwindTableRow) IsEndofFDEMarker() bool {
```
kakkoyun, line 77: 🔧 
```suggestion
func (cutr *CompactUnwindTableRow) IsEndOfFDEMarker() bool {
```

pkg/stack/unwind/compact_unwind_table_test.go
```diff
@@ -22,6 +22,14 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestIsEndofFDEMarkerWorks(t *testing.T) {
```
kakkoyun, line 25:
```suggestion
func TestIsEndOfFDEMarkerWorks(t *testing.T) {
```
🔧

kakkoyun (commented):
Looks good to me.   ❤️ 💚 💙 💛 💜 

I have pointed out some stylistic things, but they're not blockers. We can handle them on subsequent PRs.

kakkoyun:
@javierhonduco I can't officially approve this 😅 Feel free to proceed as you see fit!

bpf/cpu/cpu.bpf.c
javierhonduco: Very! Every single time we fetch a stack using the fp unwinding (e.g. Go processes) we would get a log

cmd/eh-frame/main.go
javierhonduco: No strong opinions here, but I thought [we discussed](https://github.com/parca-dev/parca-agent/pull/1004#discussion_r1017847289) that we preferred camelcase even for acronyms. Which one would you prefer that we use?

pkg/stack/unwind/compact_unwind_table.go
javierhonduco: Nice eye 🦅

javierhonduco:
That's funny 😄

cmd/eh-frame/main.go
kakkoyun: Oops, sorry 😅 This keeps triggering my alerts because it's not the idiomatic way to name abbreviations in Go.
Thanks for baring with my nitpicks 😁

javierhonduco: No worries! If we decide to change this we can perhaps mass-change all the offences in another PR?

kakkoyun: Sounds good to me. 

But again, I don't have any problems if you don't get frustrated with my autonomous PR comments (some part of my brain acts as a linter 😁). Feel free to save the above comment as a template 😜

PR #1320, opened 2023-02-09
pkg/metadata: Use os.Readlink replace process.Path() and use right elf…

zdyj3170101136:
function process.Path use filepath.EvalSymlinks to get filepath, 
this will return error, if file not existed in / directory but in absoulute directory.

readlink:
```
[root@iZuf6b8c1etgx8ypqeldc1Z 6504]# readlink ./exe
/usr/bin/cilium-agent
[root@iZuf6b8c1etgx8ypqeldc1Z 6504]# pwd
/proc/6504
```

 filepath.EvalSymlinks return error:
```
lstat /usr/bin/cilium-agent: no such file or directory
```

### DIFF
--- a/pkg/metadata/compiler.go
+++ b/pkg/metadata/compiler.go
@@ -17,10 +17,11 @@ package metadata
 import (
 	"debug/elf"
 	"fmt"
+	"os"
+	"path/filepath"
 	"sync"
 
 	burrow "github.com/goburrow/cache"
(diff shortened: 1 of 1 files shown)


kakkoyun (approved):
LGTM ❤️ 

We should converge on using prometheus/procfs and get rid of all other, so this one also helps that purpose.

PR #1336, opened 2023-02-14
bpf: Do not reuse kernel stack hash

javierhonduco:
When using `BPF_F_REUSE_STACKID` if a different stack hashes to the same hash, the old stack will be replaced with a new one. If a collision like that happens, we will get the wrong stack when finding the right kernel stack for its hash.

Right now we only have up to `MAX_STACK_TRACES_ENTRIES=1024` unique stacks, which is makes hash collitions incredibly common. This will be changed in a forthcoming commit.


kakkoyun (approved):
LGTM  ❤️ 💚 💙 💛 💜

PR #1337, opened 2023-02-14
bpf: Increase the number unique stacks

javierhonduco:
Right now, the current value is ~1k. We will have collisions very quickly. With 500 elements alone, we'll have a probability of almost 1 of a hash collision (see [0]), so let's bump it to 64k.

[0] : https://preshing.com/20110504/hash-collision-probabilities


kakkoyun (approved):
LGTM  ❤️ 💚 💙 💛 💜

Issue #1304, opened 2023-02-06
Set versions of C dependencies and compile with frame pointers


javierhonduco:
We depend on `libbpf`, which links `zlib` and `libelf`. Let's make sure that the versions of these packages we use are defined by us, and are not different across machines depending on the available version provided by the OS. While we are at it, let's compile them with frame pointers, too.

I do this in `rbperf` for `zlib` and `libelf` to ensure consistent builds and to make troubleshooting regressions in behaviour and performance easier with [this script](https://github.com/javierhonduco/rbperf/blob/main/tools/build_deps).

cc @kakkoyun

maxbrunet:
_\*whispers* Nix..._

kakkoyun:
> _*whispers* Nix..._

I'd be happy to use Nix, as the base image of choice, for our build pipeline as long as it checks all our boxes.
As far as I know, Nix means more reproducible builds, even locally.

We still have this project for these types of aspirations 🤗 https://github.com/parca-dev/cross-builder

maxbrunet:
Nix is not a base image, it is a build tool and package manager. It would replace goreleaser, podman, git-submodules, and probably parts of the Makefile and GHA workflows. If done correctly, it should check all the boxes very strictly... its main caveat being having to learn the language itself. If that's something you are willing to try, I'll try to find time to build a POC. I would need to do some research on how to structure it and other best practices (it would a first for me too).

`cross-builder` would be unnecessary and this GitHub Action would be used in the workflows: [install-nix](https://github.com/marketplace/actions/install-nix). The container image could be use as an alternative to installing it locally.

kakkoyun:
https://github.com/parca-dev/parca-agent/issues/1304 add a way to download the dependencies and build locally for the current host.

We need to extend the script to cross-build against multiple architectures and configure goreleaser accordingly.

kakkoyun:
Also these configuration variable should be extended with `-fno-omit-frame-pointer -mno-omit-leaf-frame-pointer` https://github.com/parca-dev/parca-agent/blob/bd6a25af07744bef33f85d7ae51cfadc3ce93f5f/.goreleaser.yml#L27

https://github.com/parca-dev/parca-agent/blob/bd6a25af07744bef33f85d7ae51cfadc3ce93f5f/.goreleaser.yml#L59

PR #1346, opened 2023-02-15
build: Compile libbpf with frame pointers

javierhonduco:
Test Plan
=========

```
$ make libbpf
$ ar x dist/libbpf/amd64/libbpf.a
$ gdb linker.o
```
**Before**

```
Dump of assembler code for function bpf_linker__new:
   0x00000000000000fa <+10>:	sub    $0x28,%rsp
   0x00000000000000fe <+14>:	mov    %rdi,%rbp
   0x0000000000000101 <+17>:	test   %rsi,%rsi
```

**After**

```
Dump of assembler code for function bpf_linker__new:
   0x00000000000000f0 <+0>:	push   %rbp
   0x00000000000000f1 <+1>:	mov    %rsp,%rbp
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,8 +78,7 @@ CGO_LDFLAGS_DYN = -L$(abspath $(LIBBPF_DIR)) -fuse-ld=ld -lelf -lz -lbpf
 # CGO_FFLAGS ?=
 
 # libbpf build flags:
-# CFLAGS = -g -O2 -Wall -fpie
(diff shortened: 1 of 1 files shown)


javierhonduco:
This fixes 1/3 libraries in https://github.com/parca-dev/parca-agent/issues/1304, but probably this is the one that would yield the most benefit, as I don't think the elfutils and zlib paths would run after loading the BPF program

kakkoyun (approved):

PR #1347, opened 2023-02-15
.eh_frame: Ensure we give a fairer chance of being profiled

javierhonduco:
When we run out of unwind shards, we wipe out the whole state, and reconstruct the unwind info from scratch.

This has two big problems:
- First, we aren't being very fair. The chances of the same profiled processes being profiled again is high as we iterate through procfs in order. We are only trying to make this slightly less bad in this PR. More details below.
- Second, and this is the most important bit, we want to collect enough signal from the processes once we pay the relatively costly price of generating the unwind info for them.

This commits waits up to some defined number of profiling rounds before wiping the state (this would be aking to a free() in a bump allocator).

This is not that important with the current configuration, as the likelihood of using 50 shards is not big, but it'll be key to implement the adative sizing of the unwind shards depending on the available memory, to make DWARF unwinding possible in small boxes.

Considerations
==============

This change doesn't tackle complete fairness nor being efficient about which tables we generate. The latter will be something I want to focus on in the next few days.

There's a known race condition that might happen more often with this logic. This will be tackled in another PR.

Test Plan
=========
Ran on my machine for 2h without issues.

### DIFF
--- a/pkg/profiler/cpu/cpu.go
+++ b/pkg/profiler/cpu/cpu.go
@@ -23,6 +23,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"math/rand"
 	"os"
 	"regexp"
 	"runtime"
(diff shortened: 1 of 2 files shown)


brancz (approved):
Random tends to actually be surprisingly good unless you want to go down evolutionary algorithms… nice catch

kakkoyun (approved):
LGTM ❤️ 💚 💙 💛 💜

PR #1348, opened 2023-02-15
metadata: Fix association problems of children processes

### LINKED ISSUE #1258: Metadata association problems with children processes, part 2
Follow up to https://github.com/parca-dev/parca-agent/issues/1157

I've tested parca-agent 0.12.0 with parca 0.15.0. Unfortunately I still see problems when trying to profile RabbitMQ containers.

This is what I see when I filter by `namespace`:
namespace

And this is what I see when I filter by `comm`:
comm

All those processes run in the `cqv2` namespace, that I filter by. It's hard to show but there are some sporadic dots on the graph when filtering by namespace.

If you want to test on your own, you should be able to deploy RabbitMQ like this:
1. deploy RabbitMQ Operator
```
kubectl apply -f https://github.com/rabbitmq/cluster-operator/releases/latest/download/cluster-operator.yml
```
2. Deploy this manifest:
```
apiVersion: rabbitmq.com/v1beta1
kind: RabbitmqCluster
metadata:
  name: rabbitmq
spec:
  replicas: 1
  image: rabbitmq:management
  resources:
    requests:
      cpu: 1
      memory: 1Gi
    limits:
      cpu: 1
      memory: 1Gi
  persistence:
    storage: 0
  rabbitmq:
    envConfig: |
      SERVER_ADDITIONAL_ERL_ARGS="+JPperf true"
```

kakkoyun:
Our previous strategy, relying on the `namespaces` for associating the adjacent processes, did not work out entirely. You can see below that the entry point PIDs in a container and the children processes do not share the same `pid` and `pid_for_children` namespaces; thus, we change the strategy with this PR.

CleanShot 2023-02-15 at 15 22 49

The proposed PR does the following: 
- Only stores the `EntryPID` for each target group (a.k.a `systemd` unit, `container`)
- For each received process ID, it builds a tree for tracking the hierarchy of the processes. And looks up the ancestry chain to find out the `EntryPID` of the group that it belongs to associate the metadata.

There are some downsides attached to the current strategy:
- First, we need to clean up the process tree we build. The mechanism to handle this will be introduced as part of a greater process discovery effort in a subsequent PR. We do not handle it right now because it is hard to get it right. And needs diligent testing. To be able to unblock our users, I kept it brief. This should only create a big problem if the environment has a lot of volatile processes that we profile.
- Moreover, `procfs.Proc.Stat` calls are relatively costly; we might want to cache the results aggressively, but this needs more consideration (see above, also check https://github.com/parca-dev/parca-agent/pull/1348#discussion_r1108996589).

Fixes #1258 

References:
- https://man7.org/linux/man-pages/man5/proc.5.html
- https://github.com/kakkoyun/subshells

# TODO
- [x] Unit tests

### Test Plan

Running in on a local minikube cluster and testing against demo program and `rabbitmq`.

#### Subshell Tests

1. deploy subshells demo app [https://github.com/kakkoyun/subshells](https://github.com/kakkoyun/subshells)
```
kubectl run subshells --image ghcr.io/kakkoyun/subshells:v0.3.1  
```

CleanShot 2023-02-16 at 11 49 23
CleanShot 2023-02-16 at 11 50 26
CleanShot 2023-02-16 at 11 50 55

#### RabbitMQ Tests
1. deploy RabbitMQ Operator
```
kubectl apply -f https://github.com/rabbitmq/cluster-operator/releases/latest/download/cluster-operator.yml
```
3. Deploy this manifest:
```
apiVersion: rabbitmq.com/v1beta1
kind: RabbitmqCluster
metadata:
  name: rabbitmq
spec:
  replicas: 1
  image: rabbitmq:management
  resources:
    requests:
      cpu: 1
      memory: 1Gi
    limits:
      cpu: 1
      memory: 1Gi
  persistence:
    storage: 0
  rabbitmq:
    envConfig: |
      SERVER_ADDITIONAL_ERL_ARGS="+JPperf true"
```
CleanShot 2023-02-16 at 11 46 42
CleanShot 2023-02-16 at 11 45 15
CleanShot 2023-02-16 at 11 44 15


brancz:
I realize this is WIP, just for when you’re ready to open this, can you make sure to write in the commit/PR description what the stretegy is that we’re using not and some of the different cases (eg why/how containers work, why/how systemd units work).

Either way, really awesome that you’re tackling this!! This is a real problem with users, really good that we’re doing this!

brancz:
Thanks for the thorough PR description, definitely helped me understand much better what's happening!

> Moreover, procfs.Proc.Stat calls are relatively costly; we might want to cache the results aggressively, but this needs more consideration (see above).

Is this really a problem? I thought process metadata is only ever resolved once, that should amortize that cost no?

Something that I didn't understand from the screenshot was, how did we go from the process with the PID 4495 to PID 4473?

pkg/process/process.go
```diff
@@ -0,0 +1,143 @@
...
+import (
+	"fmt"
+
+	"github.com/prometheus/procfs"
+
+	"github.com/parca-dev/parca-agent/pkg/cgroup"
+)
+
+// key is a unique identifier for a process.
+// TODO(kakkoyun): This is not unique enough. We will need to add more fields.
```
brancz, line 26: I think what you've done is fine since there is a case for when the start time has changed. It doesn't necessarily have to be part of the key as long as it's otherwise handled.

pkg/process/process.go
```diff
@@ -0,0 +1,143 @@
...
+		if !ok {
+			break
+		}
+		ancestors = append(ancestors, &parent.Proc)
+		proc = parent
+	}
+	return ancestors
+}
+
+// TODO(kakkoyun): This is an ever growing map. Introduce a mechanism to prune it.
```
brancz, line 63: I think it would be better to not have this be a package global, but rather a struct that's instantiated once and passed around. Then lifecycle management with a long-running go-routine is also easier.

kakkoyun:
> Is this really a problem? I thought process metadata is only ever resolved once, that should amortize that cost no?

Yes, you're right. This comment is out of place because I didn't commit the version that made me write this description.

I used the `starttime` as part of the key in the aforementioned uncommitted version. 

Since it's not part of the PR, I'll fix the description.

brancz:
makes sense, probably just a manual walking of parent PIDs with the rabbitmq example would already be enough, I think I get how it works, but helps to see the concrete manual steps

pkg/process/process.go
kakkoyun: I'll do this. This was hard to commit for me anyway 😅

kakkoyun:
@brancz Could you please have another look?

pkg/process/process.go
```diff
@@ -85,38 +93,45 @@ func FindAllAncestorProcessIDsInSameCgroup(pid int) ([]int, error) {
...
+
+	// Tree is updated. Try to find the process again.
+	p, ok := t.tree[k]
+	if !ok {
+		return nil, fmt.Errorf("failed to find the process for PID %d", pid)
+	}
+	return findAncestorPIDsInSameCgroup(p)
+}
+
+func (t *Tree) populate() error {
```
brancz, line 108: Until we prove that this is a perf problem, what do you think of just completely re-building the tree and returning it in this function? Then we don't have to worry about pruning as we rebuild from scratch anyway.

Since we profile every 10s only, I would also throttle rebuilding to at most every 10s (I would also only add this optimization if we see it necessary though).

pkg/process/process.go
```diff
@@ -0,0 +1,158 @@
...
+// NewTree returns a new process tree with current state of all the processes on the system.
+func NewTree() *Tree {
+	return &Tree{
+		// TODO(kakkoyun): This is an ever growing map. Introduce a mechanism to prune it.
+		tree: make(map[key]*process),
+	}
+}
+
+// FindAllAncestorProcessIDsInSameCgroup returns all ancestor process IDs for a given PID in the same cgroup.
+func (t *Tree) FindAllAncestorProcessIDsInSameCgroup(pid int) ([]int, error) {
```
brancz, line 75: Are we 100% certain that there is no race conditions possible here? I feel like there could be multiple go routines potentially doing work here no?

pkg/process/process.go
kakkoyun: We can try. I'm happy to run this for a couple of hours on my local cluster and attach the profile here.

pkg/process/process.go
kakkoyun: In theory, since we only have one profiler right now, there shouldn't be a race condition because we only have on `cpu.Run` at a time https://github.com/parca-dev/parca-agent/blob/68d2c8cf561d94ab765e1f93d9408938d0eae694/pkg/profiler/cpu/cpu.go#L395

However, with multiple profilers, this could be possible. But if it's the case, we have similar problems in several places in the code base.

I will test what you proposed above. With that, this could be ok.

pkg/process/process.go
kakkoyun: CleanShot 2023-02-16 at 21 31 34

https://pprof.me/2ba7315/

It's taking the majority of the CPU time with these changes https://github.com/parca-dev/parca-agent/pull/1348/commits/20ac55d6c10d6d518b79f24cca88c83655ea3e10

I'll revert and introduce a lock for shared resource.

pkg/process/process.go
kakkoyun: Ah also we have a lock on the call site https://github.com/parca-dev/parca-agent/blob/20ac55d6c10d6d518b79f24cca88c83655ea3e10/pkg/metadata/service_discovery.go#L35-L36

kakkoyun: see 432b2ff4139ee41590b3795e7b858f2c086e0767

pkg/metadata/service_discovery.go
```diff
@@ -15,21 +15,54 @@
...
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/prometheus/common/model"
 
 	"github.com/parca-dev/parca-agent/pkg/discovery"
+	"github.com/parca-dev/parca-agent/pkg/process"
 )
 
+type serviceDiscoveryProvider struct {
+	StatefulProvider
```
brancz, line 30: I find this hides a lot of context, I would prefer if we got rid of this and StatelessProvider entirely and rather have tiny amounts of duplication than this abstraction that hides very important context.

pkg/metadata/service_discovery.go
```diff
@@ -15,21 +15,54 @@
...
+	if p.state == nil {
+		return nil, errors.New("state not initialized")
+	}
+
+	pids, err := p.tree.FindAllAncestorProcessIDsInSameCgroup(pid)
+	if err != nil {
+		return nil, err
+	}
+
+	p.mtx.RLock()
```
brancz, line 45: This function writes state, so this should be a `p.mtx.Lock()`, no?

pkg/process/process.go
```diff
@@ -0,0 +1,234 @@
...
+}
+
+// NewTree returns a new process tree with current state of all the processes on the system.
+func NewTree(ctx context.Context, resetDuration time.Duration) *Tree {
+	t := &Tree{
+		tree:        make(map[key]*process),
+		disappeared: make(map[key]struct{}),
+		mtx:         &sync.RWMutex{},
+	}
+	go func() {
```
brancz, line 82: If we do this the caller has no control over the lifetime of this goroutine. Let's move this to a `.Run(ctx)` function which the caller needs to call, and the caller decides whether to run it in a goroutine. I think this should probably even be a top-level goroutine managed by `run.Group`.

pkg/process/process.go
```diff
@@ -0,0 +1,234 @@
...
+	return findAncestorPIDsInSameCgroup(p)
+}
+
+func (t *Tree) populate() error {
+	procs, err := procfs.AllProcs()
+	if err != nil {
+		return fmt.Errorf("failed to get all processes: %w", err)
+	}
+
+	t.mtx.Lock()
```
brancz, line 186: It's rather unpredictable how long we'll hold this lock for, so I think we should only obtain a read lock here and return the changes that need to happen to the map, and then only obtain the lock for the short moment where we populate the map with the new changes.

I would also suggest that this function doesn't actually obtain the lock at all, and rather just returns the difference. It's easier to reason about locking behavior when all the locking/unlocking happens in one function.

Another benefit of moving the write-lock to the caller of this function is that the subsequent read that we currently hold a read lock for can just be done as part of the write-lock.

pkg/process/process.go
```diff
@@ -0,0 +1,234 @@
...
+
+	tree *Tree
+
+	parent int
+	// name of the field is the same as the one in kernel struct.
+	starttime uint64
+}
+
+func (p *process) ancestors() []*procfs.Proc {
+	p.tree.mtx.RLock()
```
brancz, line 55: Since only the accesses need to be locked, let's move the locking into a `Get` function bound to the tree, and then just call `p.tree.Get(k)` in this function.

pkg/process/process.go
```diff
@@ -0,0 +1,234 @@
...
+		}
+	}()
+	return t
+}
+
+// FindAllAncestorProcessIDsInSameCgroup returns all ancestor process IDs for a given PID in the same cgroup.
+func (t *Tree) FindAllAncestorProcessIDsInSameCgroup(pid int) ([]int, error) {
+	// Fast path. Find the process if it exists in the process tree.
+	k := key{pid: pid}
+	t.mtx.RLock()
```
brancz, line 144: Can we move the part from here to the read-unlock to a separate function, that returns `[]int, needsRefresh`? We also only need to lock around the `t.tree[k]` access, the procfs accesses can theoretically block arbitrarily long but it actually doesn't matter for the read lock.

pkg/process/process.go
brancz: Thank you for trying this! Shame, but I guess it makes sense.

pkg/metadata/service_discovery.go
brancz: Just for clarity, I'm fine with just not using StatefulProvider in the process discovery for this PR and we do follow ups for the other providers.

pkg/process/process.go
brancz: Actually this could use the same `Get` function suggested elsewhere.

pkg/metadata/service_discovery.go
kakkoyun: 👍 Sounds good. It's only used here. I'll do the refactoring.

pkg/metadata/service_discovery.go
kakkoyun: In only reads the state. Do I miss something?

pkg/process/process.go
kakkoyun: Good call 👍

pkg/process/process.go
kakkoyun: It actually uses the same context and terminates the goroutine loop.
However, I like the idea of introducing a `Run(ctxt)` function. I'll do it.

pkg/metadata/service_discovery.go
brancz: You're totally right. It was the stateful provider that was hiding the writing which confused me.

brancz:
Alright. I performed the same tests as @kakkoyun and I think this is ready for review.

PR #1353, opened 2023-02-16
pkg/profiler/cpu: List all available threads rather than processes

javierhonduco:
While the procfs library we use does everything properly, we weren't getting all the task PIDs.

As it turns out, while accessing `/proc/$TASK_PID` works, listing `/proc` won't show threads...


Sylfrena (approved):

kakkoyun (commented):
❤️ 💚 💙 💛 💜

PR #1358, opened 2023-02-16
build: Upgrade LLVM from 11 to 14

javierhonduco:
LLVM 11 is ancient, let's upgrade to 14. Additionally, this fixes the problem that https://github.com/parca-dev/parca-agent/pull/1357 was trying to fix. I suggest we land both because the code is a bit cleaner and it would be good to use a more modern compiler 😄


brancz (approved):

kakkoyun (changes requested):
LGTM ❤️ 💚 💙 💛 💜

For completeness, let's upgrade everything at once.
I believe some documents mention the minimum required clang version. Let's update those as well.

Also, if we have other version specifications in our scripts/tooling, updating those would be great.

PR #1361, opened 2023-02-16
deploy/manifests: Add config for recently introduced flags

kakkoyun:


deploy/dev.jsonnet
```diff
@@ -25,6 +25,7 @@ function(serverVersion='v0.4.2')
     logLevel: 'debug',
     insecure: true,
     insecureSkipVerify: true,
+    profilingCPUSamplingFrequency: 100,
```
brancz, line 28: I think we should stay away from 100, if you want to use a non 19 value use 23 I would say (the next prime number)

kakkoyun: I pick a higher prime. This is just for local testing, by the way.

brancz (approved):

PR #1364, opened 2023-02-17
perf: automatically set GOMAXPROCS to match Linux container CPU quota

maxbrunet:
Same as parca-dev/parca#2601, since the agents likely runs in a container, there may be a CPU limit, better be aware of it. If there is none, it does not hurt, it is a no-op.


kakkoyun (approved):
LGTM 🚀

PR #1367, opened 2023-02-20
native unwinding: Ensure we discard failed stacks when using frame pointers

javierhonduco:
We currently unwind user stacks with frame pointers using the `bpf_get_stackid` BPF helper.

While its manpage asserts that errors will be returned on failure, it doesn't clarify what it considers a failure. As it turns out, there's no validation of whether the last frame was found (frame pointer == 0). It will only err when there's a failure inserting the stack in the map due to collisions, for example.

By adding a check to see if we can find the bottom frame using frame pointers, not only we ensure that the stacks are correct, but also can lazily generate the DWARF-derived unwind tables. This will be implemented in another commit.

Test Plan
=========

Ran the Agent for 1h with no issues. All the stacks look correct, no more truncated stacks while generating the DWARF derived unwind info.
image


bpf/cpu/cpu.bpf.c
```diff
@@ -935,37 +965,41 @@ int profile_cpu(struct bpf_perf_event_data *ctx) {
...
     walk_user_stacktrace(ctx);
+    return 0;
+  }
+
+  if (has_fp(unwind_state->bp)) {
+    add_stack(ctx, pid_tgid, STACK_WALKING_METHOD_FP, NULL);
+    return 0;
   }
 
+  // Debugging.
```
javierhonduco, line 995: Leaving this as it's somewhat useful while implementing the on-demand unwind tables generation. Will do this in another PR in few days

bpf/cpu/cpu.bpf.c
```diff
@@ -935,37 +965,41 @@ int profile_cpu(struct bpf_perf_event_data *ctx) {
...
+  if (unwind_state == NULL) {
+    // This should never happen.
+    return 0;
+  }
 
+  if (has_unwind_information(user_pid)) {
     shard_info_t *shard = NULL;
     find_unwind_table(&shard, user_pid, unwind_state->ip, NULL);
     if (shard == NULL) {
-      bpf_printk("IP not covered. In kernel space / JIT / bug? IP %llx)", unwind_state->ip);
```
javierhonduco, line 956: This is no longer valid because we are retrieving now the user registers saved when switching to kernel code during a syscall execution

v-thakkar:
This is super interesting. Just for my understanding, can you please put a screenshot of the stacks before this change? 

Also, seems like `bpf_get_stackid`  should ideally check this. Do you plan to propose a change in the mainline? Will be nice I think.

javierhonduco:
This is how an application without frame pointers looks when running CPython and CRuby:

**before**
image

**after**

(without DWARF-based unwinding)

image

There are some single frames that appear because even though `rbp=0` we haven't reached the bottom of the stack. This is not compliant with the spec and we'll have to add some special handling for this in the future, but at least it's single frames that happen less often than the truncated stacks in the screenshot above

(with DWARF-based unwinding)

image



> Also, seems like bpf_get_stackid should ideally check this. Do you plan to propose a change in the mainline? Will be nice I think.

Yes, it's something that I've thought about and also Sumera brought up. This will be nice to have but it will break the current assumptions that users have, so it'll probably take a bit longer to get it landed upstream. That being said it will be nice to try to tackle this in some months.

v-thakkar (approved):
Thanks for the additional screenshots. Overall, I like that we're simultaneously making sure that fp-based unwinding works well along with DWARF based unwinding. 

Changes look good to me, nice work both of you :tada:

PR #1375, opened 2023-02-22
.eh_frame: Enable DWARF-based unwinder by default

javierhonduco:
This commit:

- enables DWARF-unwinding by default, adding a flag to opt-out `--disable-dwarf-unwinding`;
- adds on-demand unwind table generation using the information provided by the BPF program when it can't unwind the stack using FPs and the unwind information isn't found;
- adds a new flag, `--dwarf-unwinding-use-polling` to produce the unwind tables ahead of time, if desired;

Notes
=====

I expect the CPU usage to be slightly lower thanks to not generating unwind info ahead of times for processes that perhaps will use no CPU.

Some stacks fail due to not following the ABI. This needs more investigation but seems related to the way Go grows goroutines stacks. This might cause many events to produce unwind tables to be sent, but this should not be a problem as the Go binary detection is cached as well as the unwind table generation. This code path should use very little CPU.

That being said, will take a look at profiles in Prod and Demo to ensure performance is within what we expect.

Test Plan
=========

Ran for 2h on my machine without issues or errors in the BPF logs. 

image

### DIFF
--- a/cmd/parca-agent/main.go
+++ b/cmd/parca-agent/main.go
@@ -131,8 +131,9 @@ type flags struct {
 	// Hidden debug flags (only for debugging):
 	DebugProcessNames []string `kong:"help='Only attach profilers to specified processes. comm name will be used to match the given matchers. Accepts Go regex syntax (https://pkg.go.dev/regexp/syntax).',hidden=''"`
 
-	// These flags are experimental. Use them at your own peril.
-	ExperimentalEnableDWARFUnwinding bool `kong:"help='Unwind stack using .eh_frame information.',hidden=''"`
+	// Native unwinding flags.
+	DisableDWARFUnwinding    bool `kong:"help='Do not unwind using .eh_frame information.'"`
+	DWARFUnwindingUsePolling bool `kong:"help='Poll procfs to generate the unwind information instead of generating them on demand.'"`
 }
 
 var _ Profiler = &profiler.NoopProfiler{}
@@ -163,8 +164,8 @@ func main() {
 		os.Exit(1)
 	}
 
-	if flags.ExperimentalEnableDWARFUnwinding && runtime.GOARCH == "arm64" {
-		level.Error(logger).Log("msg", "DWARF-unwinder support for ARM64 is currently in progress. See https://github.com/parca-dev/parca-agent/issues/1209")
+	if runtime.GOARCH == "arm64" {
+		level.Error(logger).Log("msg", "ARM64 support is currently in progress. See https://github.com/parca-dev/parca-agent/discussions/1376")
 		os.Exit(1)
 	}
 
@@ -185,15 +186,15 @@ func main() {
 
 	// Memlock rlimit 0 means no limit.
 	if flags.MemlockRlimit != 0 {
-		if flags.ExperimentalEnableDWARFUnwinding {
-			if flags.MemlockRlimit < defaultMemlockRLimitWithDWARFUnwinding {
-				level.Warn(logger).Log("msg", "memlock rlimit is too low for DWARF unwinding. Setting it to the minimum required value", "min", defaultMemlockRLimitWithDWARFUnwinding)
-				flags.MemlockRlimit = defaultMemlockRLimitWithDWARFUnwinding
-			}
-		} else {
+		if flags.DisableDWARFUnwinding {
 			if flags.MemlockRlimit < defaultMemlockRLimit {
 				level.Warn(logger).Log("msg", "memlock rlimit is too low. Setting it to the minimum required value", "min", defaultMemlockRLimit)
 				flags.MemlockRlimit = defaultMemlockRLimit
+			}
+		} else {
+			if flags.MemlockRlimit < defaultMemlockRLimitWithDWARFUnwinding {
+				level.Warn(logger).Log("msg", "memlock rlimit is too low for DWARF unwinding. Setting it to the minimum required value", "min", defaultMemlockRLimitWithDWARFUnwinding)
+				flags.MemlockRlimit = defaultMemlockRLimitWithDWARFUnwinding
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


cmd/parca-agent/main.go
```diff
@@ -163,7 +164,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	if flags.ExperimentalEnableDWARFUnwinding && runtime.GOARCH == "arm64" {
+	if !flags.DisableDWARFUnwinding && runtime.GOARCH == "arm64" {
```
brancz, line 167: Perhaps we can suggest to the user what to do? In this case, we would recommend to add the `--disable-dwarf-unwinding` flag, right?

Or should we detect we're on ARM64 and turn it off? I kind of like this because it means when people deploy it fleet-wide (which is typical on Kubernetes) across amd64 and arm64 machines they can deploy them with consistent flags.

pkg/profiler/cpu/cpu.go
```diff
@@ -269,6 +272,76 @@ func loadBpfProgram(logger log.Logger, debugEnabled bool, memlockRlimit uint64)
...
+//
+// Waiting for as long as `duration` is important because `PersistUnwindTable`
+// must be called to write the in-flight shard to the BPF map. This has been
+// a hot path in the CPU profiles we take in Demo when we persisted the unwind
+// tables after adding every pid.
+func onDemandUnwindInfoBatcher(eventsChannel <-chan []byte, duration time.Duration, callback func([]int)) {
+	batch := make([]int, 0)
+	timerOn := false
+	timer := &time.Timer{}
+	for {
```
brancz, line 323: this leaks a goroutine, we would want to pass the parent context in here, and return when the `<-ctx.Done()`

pkg/profiler/cpu/cpu.go
```diff
@@ -269,6 +272,76 @@ func loadBpfProgram(logger log.Logger, debugEnabled bool, memlockRlimit uint64)
...
+// frame pointers and unwind information are not present.
+//
+// Waiting for as long as `duration` is important because `PersistUnwindTable`
+// must be called to write the in-flight shard to the BPF map. This has been
+// a hot path in the CPU profiles we take in Demo when we persisted the unwind
+// tables after adding every pid.
+func onDemandUnwindInfoBatcher(eventsChannel <-chan []byte, duration time.Duration, callback func([]int)) {
+	batch := make([]int, 0)
+	timerOn := false
+	timer := &time.Timer{}
```
brancz, line 322: I suspect this is to protect from `timer` not being truly set yet and for example on line 336 on the first iteration it would result in a `nil` panic. I would have instead expected to see just a variable declaration and a first `select` statement outside of the `for` loop.

pkg/profiler/cpu/cpu.go
```diff
@@ -269,6 +272,76 @@ func loadBpfProgram(logger log.Logger, debugEnabled bool, memlockRlimit uint64)
...
+	for {
+		select {
+		case receivedBytes := <-eventsChannel:
+			if !timerOn {
+				timerOn = true
```
brancz, line 327: I think this deserves a code comment to explain that we are still below the deadline set by the first receivedBytes of the batch, or vice versa, we're starting a new batch. And we only want to set a deadline once we've received a first set of bytes.

pkg/profiler/cpu/cpu.go
```diff
@@ -269,6 +272,76 @@ func loadBpfProgram(logger log.Logger, debugEnabled bool, memlockRlimit uint64)
...
+			}
+			if len(receivedBytes) == 0 {
+				continue
+			}
+			pid := int(binary.LittleEndian.Uint32(receivedBytes))
+			batch = append(batch, pid)
+
+		case <-timer.C:
+			callback(batch)
+			batch = make([]int, 0)
```
brancz, line 338: probably doesn't matter perf wise, but I would have expected to see a batch reuse instead

```
batch = batch[:0]
```

pkg/profiler/cpu/cpu.go
brancz: Alternatively, this could be done by closing the channel that's received from.

cmd/parca-agent/main.go
javierhonduco: Actually right now we should probably just exit on ARM64. With the improved frame pointer detection the Agent won't work on this architecture.

We need to add support for a bunch of things, including kernel tests for ARM64 and integration tests to ensure we don't regress

pkg/profiler/cpu/cpu.go
javierhonduco: Right! There will be one goroutine leaked if we stop the profiler, let me add that logic. Will be important if we add multiple concurrent profilers and we multiplex around them

pkg/profiler/cpu/cpu.go
javierhonduco: Exactly, yeah! Thought about doing it in two selects, but found it less clear. No strong opinions though!

pkg/profiler/cpu/cpu.go
javierhonduco: Was afraid of a very large initial batch and then not allowing us to free the underlying buffer, but that's unlikely to happen :D

cmd/parca-agent/main.go
brancz: hmm .. is there anything best effort we can do? We already know of some people who run the agent without DWARF unwinding enabled on arm64...

pkg/profiler/cpu/cpu.go
```diff
@@ -269,6 +272,81 @@ func loadBpfProgram(logger log.Logger, debugEnabled bool, memlockRlimit uint64)
...
+				continue
+			}
+			pid := int(binary.LittleEndian.Uint32(receivedBytes))
+			batch = append(batch, pid)
+		case <-timer.C:
+			callback(batch)
+			batch = batch[:0]
+			timerOn = false
+			timer.Stop()
+		case <-ctx.Done():
```
brancz, line 344: nit: this is usually the very first case

brancz (approved):

pkg/profiler/cpu/cpu.go
javierhonduco: Changed! :)

PR #1381, opened 2023-02-23
pkg/ksym: consistence with perf

zdyj3170101136:


javierhonduco:
Hi @zdyj317010113! Thanks for your contribution! Could you add more context on which problems this PR aims to solve? Will help us to understand why it might be necessary 😄

pkg/ksym/ksym.go
```diff
@@ -204,30 +206,38 @@ func (c *cache) loadKsyms() error {
...
+		name = strings.TrimFunc(name, unicode.IsSpace)
+		for i, c := range name {
+			if unicode.IsSpace(c) {
+				name = name[:i]
+				break
+			}
 		}
 
-		symbol = string(line[19:endIndex])
-		c.kernelSymbols = append(c.kernelSymbols, ksym{address: address, name: symbol})
```
javierhonduco, line 223: Ideally, this structure should be as small as possible as we keep it in memory, on my machine it's almost 30MBs and according to heap profiles, one of the largest sources of retained memory

zdyj3170101136:
> Hi @zdyj317010113! Thanks for your contribution! Could you add more context on which problems this PR aims to solve? Will help us to understand why it might be necessary 😄

make the output consistent with perf.
- the kernel symbol have priority, like global symbol have priority over local symbol, we use symbolSearcher to do that.
- the agent would take 'a' type of symbol as function symbol, which is wrong. perf only process w or t symbol.

pkg/ksym/ksym.go
zdyj3170101136: 截屏2023-02-23 下午7 39 56

in my machine, this would cost 13 MB of memory, not too much.

zdyj3170101136: because we would filter nm type 'a' symbol, so it would contain less memory.

javierhonduco:
Would it be possible to filter the symbols as you are suggesting but still keep the previous structure as it uses less memory per entry?

zdyj3170101136:
> Would it be possible to filter the symbols as you are suggesting but still keep the previous structure as it uses less memory per entry?

it is not easy to use same structure cause the symbolsearch have many logic on checking elf type and compare.

but I could filter the elf.Object symbol  in advance

pkg/ksym/ksym.go
```diff
@@ -236,12 +250,8 @@ func (c *cache) resolveKsyms(addrs []uint64) []string {
 	result := make([]string, 0, len(addrs))
 
 	for _, addr := range addrs {
-		idx := sort.Search(len(c.kernelSymbols), func(i int) bool { return addr < c.kernelSymbols[i].address })
-		if idx < len(c.kernelSymbols) && idx > 0 {
-			result = append(result, c.kernelSymbols[idx-1].name)
-		} else {
-			result = append(result, "")
-		}
+		name, _ := c.searcher.Search(addr)
```
kakkoyun, line 253: Let's ignore the error.

kakkoyun:
@zdyj3170101136 This needs a rebase. And of course, @javierhonduco's review.

pkg/ksym/ksym.go
javierhonduco: What's type "a"? Can't find it here https://man7.org/linux/man-pages/man1/nm.1.html. I wonder if we should filter out the symbols that aren't necessary instead. That would fix the problem, no?

kakkoyun:
Probably needs a rebase.

kakkoyun:
@zdyj3170101136 Do you still want to continue working on this one?

Issue #1369, opened 2023-02-21
Don't unnecessarily strip debuginfos


brancz:
While debugging a particular situation where we pick up a debuginfo file from a debian dbgsym directory we still attempt to extract the debuginfos when they already don't contain the `.text` section, which is all we want to achieve.

We should first check whether an ELF file contains a `.text` section at all before performing the extraction.

palash25:
hi 👋🏽  i would like to work on this

brancz:
Go for it!

PR #1393, opened 2023-03-01
debuginfo: Don't strip ELF files if text section is missing

palash25:
Fixes: #1369


palash25:
I hope I understood the issue correctly. I was a bit confused whether it was asking to simply flush the file as is if `.text` was missing or return an error. I went with the former. Let me know if the solution is correct. cc @brancz 

I also wanted to write a unit test using an elf file that lacked a `.text` section but I couldn't understand how the testdata elf files were being generated even after taking a look at this tool https://github.com/kakkoyun/readelf-sections

pkg/debuginfo/extract.go
```diff
@@ -79,6 +79,13 @@ func (e *Extractor) Extract(ctx context.Context, dst io.WriteSeeker, src string)
 	if err != nil {
 		return fmt.Errorf("failed to initialize writer: %w", err)
 	}
+
+	// if ELF file doesn't contain `.text` section then there is no need to strip the file.
+	if !w.Writer.HasSection(".text") {
```
kakkoyun, line 84: 🔧 I think we don't need to add this method to the `Writer` interface. We can just check the raw ELF file to see whether it has the necessary sections and keep it as is. Since we already have an open file, we can just use `https://pkg.go.dev/debug/elf#NewFile` and quickly check the existence of the `.text` section and early return.

Also, I think doing this check even before the `Extract` call is better.

kakkoyun (commented):
Thanks for the PR ❤️  You're definitely on the right path.

I think we should restructure it a little bit. I've commented inline.

v-thakkar:
On the topic of writing a test, I haven't used the tool you linked above. But you can also use `llvm-strip` with -R option to have some testdata file with no `.text` section.

kakkoyun:
@palash25 This is so close to the finishing line. Let us know how we can help you.

palash25:
@kakkoyun i was just waiting for your reply, do the changes look good to you? i guess the only thing left is writing a unit test by generating tesdata like @v-thakkar suggested. I will try to do that tomorrow

kakkoyun:
> @kakkoyun i was just waiting for your reply, do the changes look good to you? i guess the only thing left is writing a unit test by generating tesdata like @v-thakkar suggested. I will try to do that tomorrow

The changes look good :) Thanks for the contribution. We can merge it after some tests.

pkg/debuginfo/manager_test.go
```diff
@@ -76,3 +76,36 @@ func BenchmarkEnsureUploadedInitiateUploadError(b *testing.B) {
...
+func TestHasTextSection(t *testing.T) {
+	testCases := []struct {
+		name              string
+		filepath          string
+		textSectionExists bool
+		wantErr           bool
+	}{
+		{
+			name:              "text section present",
+			filepath:          "./testdata/elf-file-with-text-section",
```
v-thakkar, line 89: For this, we don't need to add an extra file, let's reuse one of the existing test data files. Like redelf-sections

v-thakkar (changes requested):
Looking good. :tada: I just have one minor comment and we should be able to merge it after that.

pkg/debuginfo/manager_test.go
palash25: thanks for pointing it out, done.

PR #1397, opened 2023-03-01
bpf: Remove comment warning on helpers

javierhonduco:
We have kernel tests that will ensure that we don't add helpers that cause regressions.

Additionally, there are many other sources of BPF loading issues, beyond helper usage, such as using of global variables, etc


v-thakkar (approved):
LGTM :+1:

PR #1401, opened 2023-03-02
bpf: Print PID on unwind info requests

javierhonduco:
This is useful while debugging multi-threaded applications. Note that we are printing the userspace process ID, or kernel thread ID.

### DIFF
--- a/bpf/cpu/cpu.bpf.c
+++ b/bpf/cpu/cpu.bpf.c
@@ -1011,7 +1011,7 @@ int profile_cpu(struct bpf_perf_event_data *ctx) {
   // Request unwind table generation.
   char comm[20];
   bpf_get_current_comm(comm, 20);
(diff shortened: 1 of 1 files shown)


javierhonduco:
This breaks older kernels because of libbpf using `bpf_trace_vprintk`, but the vmtests did not run :(

kakkoyun (approved):
LGTM ❤️

javierhonduco:
Awesome, it failed now as expected 🎉

kakkoyun (commented):
LGTM, but which version of libbpf this changes to? Is it an upgrade or a downgrade?

javierhonduco:
My bad, that was not supposed to be part of this commit/PR

kakkoyun (approved):
LGTM 🥇

PR #1402, opened 2023-03-02
bpf: Bump max mappings per process to 250

javierhonduco:
While the current limit is enough for most applications, there are some behemoths out there, looking at you, gnome-shell, that bring more than 150 mappings.

Let's bump the maximum mappings so we can support many other processes. This increases our memory footprint and it's something we will be looking to optimise in the future.

Note that we still perform a linear search over the array of mappings. We'll also measure the impact of switching this to other searching algorithms and decide if it's worth changing it.

Test Plan
=========

Ran the agent for 30 mins without any issue, 100% of the firefox stacks were collected https://pprof.me/9a8a8c3

```
=============
Test results:
=============
- ✅ 5.4
- ✅ 5.10
- ✅ 5.18
- ✅ 5.19
```


brancz (approved):

kakkoyun (approved):
LGTM ❤️ 💚 💙 💛 💜

Issue #1398, opened 2023-03-01
Error opening executables to process unwind tables


javierhonduco:
```
elf.Open failed: bad magic number '[68 84 105 74]' in record at byte 0x0"
```

reported by one of our users. Perhaps it's a wrapper?

PR #1406, opened 2023-03-02
unwinding/native: Ignore jitdump sections

javierhonduco:
Reported by an user after seeing a bunch of errors in their production cluster such as:

```
elf.Open failed: bad magic number '[68 84 105 74]' in record at byte 0x0"
```

They graciously provided the mappings for this process, and this one stood out:

```
7f6c903cc000-7f6c903cd000 r-xp 00000000 08:01 403978                     /tmp/jit-11.dump
```

This commit also adds extra context to the logs to make it easier to debug these issues in the future.

### DIFF
--- a/pkg/stack/unwind/maps.go
+++ b/pkg/stack/unwind/maps.go
@@ -16,6 +16,7 @@ package unwind
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/prometheus/procfs"
 )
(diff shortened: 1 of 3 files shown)


pkg/stack/unwind/maps.go
```diff
@@ -48,6 +49,13 @@ func (pm *ExecutableMapping) IsJitted() bool {
 	return pm.Executable == ""
 }
 
+// IsJitDump returns whether the mapping looks like a jitdump[0] file.
+//
+// [0]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/perf/Documentation/jitdump-specification.txt
+func (pm *ExecutableMapping) IsJitDump() bool {
+	return strings.Contains(pm.Executable, "jit") && strings.HasSuffix(pm.Executable, ".dump")
```
javierhonduco, line 56: Started with `HasPrefix(..., "/jit-")`, but decided to make it a bit more loose

brancz (approved):
Probably good to get an opinion from @maxbrunet as well. But lgtm we could always do future improvements to this.

kakkoyun (approved):

javierhonduco:
Merging, but would love your feedback Maxime. Will reply in some days as I'll come back late next week

javierhonduco:
Should fix https://github.com/parca-dev/parca-agent/issues/1398

maxbrunet:
Here is the reason:

> Mapping must be PROT_EXEC to ensure it is captured by perf record even when not using -d option.

https://github.com/llvm/llvm-project/blob/llvmorg-15.0.6/llvm/lib/ExecutionEngine/PerfJITEvents/PerfJITEventListener.cpp#L345-L346

PR #1426, opened 2023-03-08
pkg/discovery: systemd discoverer with less allocs

marselester:
Hi! I would like to share some of the results I got with https://github.com/marselester/systemd. ~~There are still things I need to finish though...~~

The current version of `discovery.SystemdDiscoverer` makes use of 1.1 MB/op and 32,332 allocs/op to get a list of systemd services along with their PIDs. That accounts for 40% of CPU cycles in the agent, excluding the GC time, see https://pprof.me/e711789. The proposed change gets 13.96 KB/op and 30 allocs/op.

```
BenchmarkSystemd-2    	       1	1001314916 ns/op	 1163552 B/op	   32332 allocs/op
BenchmarkSystemd2-2   	       1	1003780290 ns/op	   14304 B/op	      30 allocs/op
```

I ran the benchmarks as follows:

- temporarily removed `pkg/discovery/kubernetes.go` or else tests won't run `github.com/parca-dev/parca-agent/pkg/cgroup: build constraints exclude all Go files in /vagrant/parca-agent/pkg/cgroup`
- placed the benchmarks into `pkg/discovery/systemd_test.go`
- `sudo go test -run=^$ -bench=. -benchmem .`

<details>

```go
// The benchmarks are not pretty, but at least they give a rough idea.

func BenchmarkSystemd(b *testing.B) {
	ctx := context.Background()
	conf := NewSystemdConfig()
	d, err := conf.NewDiscoverer(DiscovererOptions{})
	if err != nil {
		b.Fatal(err)
	}

	up := make(chan<- []*Group)
	ctx, cancel := context.WithTimeout(ctx, time.Second)
	b.Cleanup(cancel)

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		if err = d.Run(ctx, up); err != nil && err != context.DeadlineExceeded {
			b.Fatal(err)
		}
	}
}

func BenchmarkSystemd2(b *testing.B) {
	ctx := context.Background()
	conf := NewSystemd2Config()
	d, err := conf.NewDiscoverer(DiscovererOptions{})
	if err != nil {
		b.Fatal(err)
	}

	up := make(chan<- []*Group)
	ctx, cancel := context.WithTimeout(ctx, time.Second)
	b.Cleanup(cancel)

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		if err = d.Run(ctx, up); err != nil && err != context.DeadlineExceeded {
			b.Fatal(err)
		}
	}
}
```

</details>

[...]

PR #1453, opened 2023-03-14
pkg/discovery: Move github.com/marselester/systemd

marselester:
@brancz proposed to move https://github.com/marselester/systemd into the parca-agent repository due to a limited scope, see https://github.com/parca-dev/parca-agent/pull/1426.


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/parca-dev/parca-agent
 
-go 1.18
+go 1.20
```
marselester, line 3: The systemd package relies on `unsafe.String` introduced in Go 1.20.

kakkoyun (approved):
Awesome 🏆 🥇 💯

PR #1454, opened 2023-03-15
fix/objectfile: compute right cache key

zdyj3170101136:
the MappedObjectFile would cache the base address.
and the base address is different if start, limit, offset is changed.


pkg/objectfile/cache.go
```diff
@@ -80,5 +81,11 @@ func fromProcess(pid int, m *profile.Mapping) (*MappedObjectFile, error) {
 }
 
 func cacheKey(pid int, m *profile.Mapping) string {
-	return path.Join("/proc", strconv.FormatInt(int64(pid), 10), m.BuildID)
+	// use all filed needed in MappedObjectFile.computeBase to build a unique key
+	b := make([]byte, 0, 3*binary.MaxVarintLen64)
```
brancz, line 85: If we're allocating this anyway, might as well just allocate `3*8+len(m.BuildID)`.

pkg/objectfile/cache.go
```diff
@@ -80,5 +81,11 @@ func fromProcess(pid int, m *profile.Mapping) (*MappedObjectFile, error) {
...
+	b = binary.AppendUvarint(b, m.Start)
+	b = binary.AppendUvarint(b, m.Limit)
+	b = binary.AppendUvarint(b, m.Offset)
```
kakkoyun, line 88: Let's also add the process id.

brancz: Since this reflects the true address space isn’t that already unique enough on that machine?

pkg/objectfile/cache.go
```diff
@@ -80,5 +81,11 @@ func fromProcess(pid int, m *profile.Mapping) (*MappedObjectFile, error) {
 }
 
 func cacheKey(pid int, m *profile.Mapping) string {
```
kakkoyun, line 83: Let's move `pid` if it's not used. As the CI points out.

pkg/objectfile/cache.go
```diff
@@ -80,5 +81,14 @@ func fromProcess(pid int, m *profile.Mapping) (*MappedObjectFile, error) {
...
-	return path.Join("/proc", strconv.FormatInt(int64(pid), 10), m.BuildID)
+	b := make([]byte, 4*8)
+	// use all field needed in MappedObjectFile.computeBase to build a unique key
+	binary.BigEndian.PutUint64(b, m.Start)
+	binary.BigEndian.PutUint64(b, m.Limit)
```
brancz, line 87: This needs to be `b[8:]` and so on for the next lines as well

pkg/objectfile/cache.go
```diff
@@ -80,5 +81,14 @@ func fromProcess(pid int, m *profile.Mapping) (*MappedObjectFile, error) {
...
+	binary.BigEndian.PutUint64(b, m.Offset)
+
+	// use pid and file to build unique identify
+	// do not use buildID, cause it tend to be larger strings
+	binary.BigEndian.PutUint64(b, uint64(pid))
+	return string(b) + m.File
```
brancz, line 93: This causes a second allocation, let’s allocate only once when allocating the byte slice.

zdyj3170101136: how?

is this what you mean?
```
b = append(b, unsafeStringToSlice(m.File))
```

javierhonduco:
Now that we are revising this code I was wondering why we need to cache by PID. Wouldn't it be enough to cache using the buildid? Perhaps I am missing something!

zdyj3170101136:
> 现在我们正在修改这段代码，我想知道为什么我们需要通过 PID 进行缓存。使用 buildid 进行缓存还不够吗？也许我错过了什么！

just because Kakkoyun say the pid and file would cost less memory usage compared build id

javierhonduco:
> just because Kakkoyun say the pid and file would cost less memory usage compared build id

The problem is that the cache might be more efficient if we base it on buildids, no? pids come and go but buildids are stable

kakkoyun (approved):
Thanks for taking care of it 🙏 🤗

PR #1455, opened 2023-03-15
pkg/vdso: Support vdso mapping

zdyj3170101136:


zdyj3170101136:
@kakkoyun here

pkg/vdso/vdso.go
```diff
@@ -0,0 +1,85 @@
...
+		merr = multierr.Append(merr, err)
+		if err == nil {
+			break
+		}
+	}
+	if err != nil {
+		return nil, merr
+	}
+
+	//  Num:    Value          Size Type    Bind   Vis      Ndx Name
```
v-thakkar, line 56: Can we please add a comment describing what is the purpose of this information in the comment?

pkg/vdso/vdso.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	var (
+		elfFile *elf.File
+		merr    error
+		f       string
+	)
+	// find a file is enough
+	for _, vdso := range []string{"vdso.so", "vdso64.so"} {
+		f = fmt.Sprintf("/usr/lib/modules/%s/vdso/%s", kernelVersion, vdso)
+		elfFile, err = elf.Open(f)
```
v-thakkar, line 46: We should add an error handling code in case we fail to open the elf file. Also, the elf file will need to be closed after the usage.

pkg/vdso/vdso.go
```diff
@@ -0,0 +1,85 @@
...
+	//    10: ffffffffff700f50    22 FUNC    GLOBAL DEFAULT   13 __vdso_time@@LINUX_2.6
+	syms, err := elfFile.DynamicSymbols()
+	if err != nil {
+		return nil, err
+	}
+	return &Cache{searcher: symbolsearcher.New(syms), f: f}, nil
+}
+
+func (c *Cache) Resolve(addr uint64, m *profile.Mapping) (string, error) {
+	o, err := objectfile.Open(c.f, m)
```
v-thakkar, line 76: Same as above for this object file. We should close it explicitly using `defer`.

v-thakkar (changes requested):
Great work. Love that we're adding support for vdso mapping. I have a few minor comments. Once those are resolved along with lint errors, we should be able to merge this :tada:

pkg/vdso/vdso.go
zdyj3170101136: just a example of readelf --dyn-syms vdso64.so.

v-thakkar:
The changes look good to me @zdyj3170101136 . Seems like there is still a lint error in the added code, we can merge it once all checks are passed.

Btw, just in case you don't know, you can use `make go/lint` locally to check if linters are happy with your code. :smile:

v-thakkar (approved):

Issue #1429, opened 2023-03-08
Improve support for short-lived processes


javierhonduco:
Short-lived processes aren't well supported, this also affects processes that might run for several seconds, but that exit before collecting the metadata for the processes.

Besides this bug, we have another bug that results in samples being sent without mapping information, resulting in broken stacks

<img width="595" alt="image" src="https://user-images.githubusercontent.com/959128/223720171-60631609-60b1-4fa4-baf3-d8cbfa50d681.png">

Addressing the former should fix most of the instances we see as it would minimise the race condition window.

javierhonduco:
We've made a lot of progress here -- leaving closing it to @kakkoyun and @brancz

PR #1457, opened 2023-03-15
*: Populate process mappings on-demand

javierhonduco:
This commit starts the work on increasing the chances of success when profiling short-lived processes.

Among other things:
- populate the mappings cache on-demand to be able to normalize and find the build ids for processes that exit before we collect this information;
- we will now drop samples that contained frames that failed to normalize;
- reduced the unwind info batch deadline from 250 to 150 ms to increase the chance of profiling short-lived processes;

This commit doesn't tackle:
- symbolisation of short-lived processes. See comment in cpu.go;
- DWARF-based unwinding for processes that run for less than 150ms;
- improvements to how the mappings are cached;
- single stray frames that require more investigation;
- configurable batch deadlines;
- adding an eviction strategy for the different caches;

Test Plan
=========

Ran for 2h without issues.

**short-lived process: git**
```
$ while true; do git --no-pager log; done
```
image

**all processes but git**
image

### DIFF
--- a/pkg/address/normalizer.go
+++ b/pkg/address/normalizer.go
@@ -16,10 +16,10 @@ package address
 
 import (
 	"errors"
+	"fmt"
 	"os"
 
 	"github.com/go-kit/log"
-	"github.com/go-kit/log/level"
 	"github.com/google/pprof/profile"
 
 	"github.com/parca-dev/parca-agent/pkg/objectfile"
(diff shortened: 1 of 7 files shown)


javierhonduco:
First steps towards improving https://github.com/parca-dev/parca-agent/issues/1429

pkg/profiler/cpu/cpu.go
```diff
@@ -476,15 +472,57 @@ func (p *CPU) Run(ctx context.Context) error {
...
+				// On x86_64:
+				//	- unsafe.Sizeof(int(0)) = 8
+				//	- unsafe.Sizeof(uint32(0)) = 4
+				pid := int(int32(payload))
+				if payload&RequestUnwindInformation == RequestUnwindInformation && p.DwarfUnwindingWithoutPolling() {
+					requestUnwindInfoChannel <- pid
+				} else if payload&RequestProcessMappings == RequestProcessMappings {
+					// Populate mappings cache.
+					_, _ = p.processMappings.PopulateMappings(pid)
+					// TODO: even if we populate the mappings for the process we just got
```
brancz, line 504: I think this a perfectly acceptable trade-off. If this is truly common and important, I think one of two things would happen:

* The user uploads debuginfos themselves
* The process is spawned often enough that we eventually catch it

brancz (commented):
Nice work!!

pkg/profiler/cpu/cpu.go
javierhonduco: Right! That's what I thought too. For this iteration we can leave it as-is and then reevaluate things once we get feedback from our users

pkg/address/normalizer.go
```diff
@@ -45,30 +45,28 @@ func NewNormalizer(logger log.Logger, objCache ObjectFileCache) *normalizer {
...
+func (n *normalizer) Normalize(pid int, m *profile.Mapping, addr uint64) (uint64, error) {
 	if m == nil {
-		return addr
+		return 0, errors.New("mapping is nil")
 	}
 
-	logger := log.With(n.logger, "pid", pid, "buildID", m.BuildID)
-	if m.Unsymbolizable() {
-		return addr
+	if m.Unsymbolizable() { // TODO: re-evaluate this condition and return values.
```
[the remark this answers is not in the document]
javierhonduco: We might want to return an error, but wanted to do this later on

pkg/profiler/cpu/cpu.go
```diff
@@ -886,10 +924,14 @@ func (p *CPU) obtainProfiles(ctx context.Context) ([]*profiler.Profile, error) {
 						}
 					}
 
+					normalizedAddress, err := p.normalizer.Normalize(int(key.PID), m, addr)
+					if err != nil {
+						level.Debug(p.logger).Log("msg", "failed to normalise address", "pid", id.PID, "address", addr, "err", err)
+						continue
```
javierhonduco, line 930: Just realized that this is wrong, we have an inner loop but we want to skip the whole stack, so we need to continue the outer loop. 😓 

cc/ @kakkoyun

kakkoyun: We can always patch 🤗 Sorry, that I have missed it during reviews.

PR #1477, opened 2023-03-20
pkg/stack/unwind: Filter out jitdump mappings

javierhonduco:
Exclude jitdump from the results because we don't need these mappings. If in the future we add support for unwinding via the information present in these files, we can link this information to the corresponding code

These mappings are just using previous mapping entries and don't offer any advantage at this time.

Test Plan
=========
Existent unit tests.


v-thakkar (approved):
LGTM :+1:

Issue #250, opened 2022-02-22
One-off mode


brancz:
It would be really nice if parca-agent supported a one-off mode, similar to `perf` to profile CLI commands. I imagine the experience to be very similar to that of perf, but integrated with Parca APIs. Eg:

```
$ parca-agent --one-off -- ./my-cli
View profile at https://demo.parca.dev/<query parameters>
```

PR #1489, opened 2023-03-27
Option to profile a single process

marselester:
I was not sure why `bpf_get_stackid` returns random memory addresses of instructions of a profiled function. Here is a [blog post](https://marselester.com/diy-cpu-profiler-the-simplest-case-of-symbolization.html) about that. @kakkoyun among other pertinent things advised me to compare results with Parca Agent (the results are similar to findings I described in my blog).

Therefore I've added `--profiling-pid` flag to collect CPU samples of a running process as follows. Perhaps it could be a first step towards https://github.com/parca-dev/parca-agent/issues/250.

```sh
$ ./fib &
$ sudo ./dist/parca-agent --profiling-pid=$(pidof fib) --local-store-directory=./fib-samples/
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,9 @@ Flags:
                                    of memory that may be locked into RAM. It is
                                    used to ensure the agent can lock memory for
                                    eBPF maps. 0 means no limit.
+      --profiling-pid=-1           The process id to record events on. Leave
+                                   this empty to use system-wide collection from
+                                   all CPUs.
       --profiling-duration=10s     The agent profiling duration to use. Leave
                                    this empty to use the defaults.
       --profiling-cpu-sampling-frequency=19
--- a/cmd/parca-agent/main.go
+++ b/cmd/parca-agent/main.go
@@ -103,6 +103,7 @@ type flags struct {
 	MemlockRlimit uint64 `default:"${default_memlock_rlimit}" help:"The value for the maximum number of bytes of memory that may be locked into RAM. It is used to ensure the agent can lock memory for eBPF maps. 0 means no limit."`
 
 	// Profiler configuration:
+	ProfilingPID                  int           `kong:"help='The process id to record events on. Leave this empty to use system-wide collection from all CPUs.',default='-1'"`
 	ProfilingDuration             time.Duration `kong:"help='The agent profiling duration to use. Leave this empty to use the defaults.',default='10s'"`
 	ProfilingCPUSamplingFrequency uint64        `kong:"help='The frequency at which profiling data is collected, e.g., 19 samples per second.',default='${default_cpu_sampling_frequency}'"`
 
@@ -457,6 +458,7 @@ func run(logger log.Logger, reg *prometheus.Registry, flags flags) error {
 				flags.DebuginfoTempDir,
 			),
 			labelsManager,
+			flags.ProfilingPID,
 			flags.ProfilingDuration,
 			flags.ProfilingCPUSamplingFrequency,
 			flags.MemlockRlimit,
--- a/pkg/profiler/cpu/cpu.go
+++ b/pkg/profiler/cpu/cpu.go
@@ -79,6 +79,7 @@ type combinedStack [doubleStackDepth]uint64
 type CPU struct {
 	logger                     log.Logger
 	reg                        prometheus.Registerer
+	profilingPID               int
 	profilingDuration          time.Duration
 	profilingSamplingFrequency uint64
 
@@ -128,6 +129,7 @@ func NewCPUProfiler(
 	profileWriter profiler.ProfileWriter,
 	debuginfoProcessor profiler.DebugInfoManager,
 	labelsManager profiler.LabelsManager,
+	profilingPID int,
 	profilingDuration time.Duration,
 	profilingSamplingFrequency uint64,
 	memlockRlimit uint64,
@@ -152,6 +154,7 @@ func NewCPUProfiler(
 		psMapCache:   psMapCache,
 		objFileCache: objFileCache,
 
+		profilingPID:               profilingPID,
 		profilingDuration:          profilingDuration,
 		profilingSamplingFrequency: profilingSamplingFrequency,
 
@@ -404,7 +407,7 @@ func (p *CPU) Run(ctx context.Context) error {
 			Size:   uint32(unsafe.Sizeof(unix.PerfEventAttr{})),
 			Sample: p.profilingSamplingFrequency,
 			Bits:   unix.PerfBitDisabled | unix.PerfBitFreq,
-		}, -1 /* pid */, i /* cpu id */, -1 /* group */, 0 /* flags */)
+		}, p.profilingPID /* pid */, i /* cpu id */, -1 /* group */, 0 /* flags */)
 		if err != nil {
 			return fmt.Errorf("open perf event: %w", err)
 		}


javierhonduco (commented):
Thanks for the PR @marselester! I am happy to merge as-is, but there are some things to consider. This approach only supports a single process, and while it's a good start as you said, perhaps in the future profiling multiple processes would be a good addition and it won't work with this approach.

The other thing to bear in mind is that we currently have `--debug-process-names`. Perhaps we could improve the name of the flag and reuse it?

The way it works is by still setting the perf event for every process, but then we filter out the samples from processes we don't want in BPF

https://github.com/parca-dev/parca-agent/blob/e3fd4c9506b1fff981256bf5b689021dc0f7a703/bpf/cpu/cpu.bpf.c#L970

Not sure what the random addresses is referring to but I haven't read your blogpost yet, but will try to do it today!

marselester:
@javierhonduco my bad, I didn't notice `--debug-process-names` flag, thank you! It covers more use cases, that's great!
Regarding naming: if the flag filters processes by name and additionally performs some debugging, then its name is quite descriptive, no need to rename it. I will close this PR.

---

> Not sure what the random addresses is referring to

I wrote a post about simple symbolization case using [this BPF code](https://github.com/marselester/diy-parca-agent/blob/main/cmd/profiler/bpf/parca-agent.bpf.c#L78) and came to a conclusion that `bpf_get_stackid()` captures PC values all over function's instructions. Previously I thought that only return values are collected.

Here are the sampled addresses.

```
401685
401691
401692
401697
401698
4016a4
4016a7
4016ab
4016b0
4016b3
4016b4
4016b6
4016b7
4016c6
401b1a
```

They correspond to almost all the instructions of `fibNaive` function.

```asm
$ objdump --disassemble=fibNaive fib

0000000000401685 <fibNaive>:
  401685:    48 83 ff 02              cmp    $0x2,%rdi
  401689:    7f 06                    jg     401691 <fibNaive+0xc>
  40168b:    b8 01 00 00 00           mov    $0x1,%eax
  401690:    c3                       ret    
  401691:    55                       push   %rbp
  401692:    48 89 e5                 mov    %rsp,%rbp
  401695:    41 54                    push   %r12
  401697:    53                       push   %rbx
  401698:    48 89 fb                 mov    %rdi,%rbx
  40169b:    48 8d 7f fe              lea    -0x2(%rdi),%rdi
  40169f:    e8 e1 ff ff ff           call   401685 <fibNaive>
  4016a4:    49 89 c4                 mov    %rax,%r12
  4016a7:    48 8d 7b ff              lea    -0x1(%rbx),%rdi
  4016ab:    e8 d5 ff ff ff           call   401685 <fibNaive>
  4016b0:    4c 01 e0                 add    %r12,%rax
  4016b3:    5b                       pop    %rbx
  4016b4:    41 5c                    pop    %r12
  4016b6:    5d                       pop    %rbp
  4016b7:    c3                       ret    
```

Here is the program itself.

```
$ cat main.c 
#include <stdio.h>

long fibNaive(long n) {
    if (n <= 2) {
        return 1;
    }
    return fibNaive(n-2) + fibNaive(n-1);
}

int main() {
    long n = 50;
    long res = fibNaive(n);
    printf("Fibonacci number %li: %li\n", n, res);
    return 0;
}
$ gcc -static -Og -fno-omit-frame-pointer -fno-pie -no-pie -fcf-protection=none -o fib main.c
$ file fib
fib: ELF 64-bit LSB executable, x86-64, version 1 (GNU/Linux), statically linked, BuildID[sha1]=fbe39487a4ad2867d125b6734823db131ca006ff, for GNU/Linux 3.2.0, not stripped
$ ./fib & sudo go run ./cmd/profiler3 -pid=$(pidof fib)
$ zcat cpu.pprof | protoc --decode perftools.profiles.Profile ./pprof/proto/profile.proto
```

These are addresses obtained by the Parca Agent `./fib & sudo ./dist/parca-agent --profiling-pid=$(pidof fib) --local-store-directory=./fib-samples/`. They also belong to the `fibNaive` function.

```
401685
401691
401692
401697
401698
4016a4
4016a7
4016ab
4016b0
4016b3
4016b4
4016b6
4016b7
4016c6
401b1a
403377
```

marselester:
Good question :). It looks like the addresses show up in a single trace 🤔.

This is first stack trace `UserStackID:47`.

```
401685
4016a4
4016a4
4016a4
4016a4
4016a4
4016a4
4016a4
4016a4
4016a4
4016a4
4016a4
4016a4
4016a4
4016b0
4016b0
4016b0
4016b0
4016b0
4016b0
4016b0
4016b0
4016b0
4016b0
4016b0
4016b0
4016b0
4016b0
4016b0
4016b0
4016b0
4016b0
4016b0
4016b0
4016c6
401b1a
```

Here is a fragment of what profiler prints, i.e., contents of user stack BPF map (please ignore the error messages related to kernel stack).

```
$ sudo go run ./cmd/profiler3 -pid=$(pidof fib)
/proc/42909/maps
start=0x401000 limit=0x498000 offset=0x1000 /vagrant/diy-parca-agent/fib
start=0x7ffe189d5000 limit=0x7ffe189d7000 offset=0x0 [vdso]
start=0xffffffffff600000 limit=0xffffffffff601000 offset=0x0 [vsyscall]

Waiting for stack traces for 10s...
{PID:42909 UserStackID:47 KernelStackID:-14} seen 1 times
	47 [401685 4016a4 4016b0 4016b0 4016a4 4016b0 4016b0 4016a4 4016a4 4016b0 4016b0 4016a4 4016b0 4016b0 4016b0 4016a4 4016b0 4016a4 4016b0 4016b0 4016a4 4016b0 4016b0 4016a4 4016b0 4016a4 4016b0 4016b0 4016a4 4016b0 4016a4 4016b0 4016b0 4016a4 4016c6 401b1a]
2023/03/28 14:55:13 failed to read kernel-space stack traces: EOF
{PID:42909 UserStackID:354 KernelStackID:-14} seen 1 times
	354 [401698 4016b0 4016a4 4016b0 4016b0 4016b0 4016a4 4016b0 4016a4 4016b0 4016a4 4016b0 4016b0 4016a4 4016a4 4016b0 4016a4 4016b0 4016b0 4016b0 4016b0 4016a4 4016a4 4016b0 4016b0 4016b0 4016b0 4016b0 4016b0 4016a4 4016a4 4016a4 4016b0 4016a4 4016c6 401b1a]
2023/03/28 14:55:13 failed to read kernel-space stack traces: EOF
{PID:42909 UserStackID:629 KernelStackID:-14} seen 1 times
	629 [401697 4016a4 4016b0 4016b0 4016b0 4016b0 4016a4 4016a4 4016b0 4016a4 4016b0 4016b0 4016b0 4016b0 4016b0 4016a4 4016a4 4016b0 4016b0 4016b0 4016b0 4016a4 4016b0 4016b0 4016b0 4016a4 4016b0 4016a4 4016b0 4016b0 4016a4 4016a4 4016a4 4016b0 4016a4 4016c6 401b1a]
2023/03/28 14:55:13 failed to read kernel-space stack traces: EOF
{PID:42909 UserStackID:337 KernelStackID:-14} seen 1 times
	337 [4016b0 4016b0 4016b0 4016b0 4016b0 4016b0 4016b0 4016b0 4016a4 4016a4 4016b0 4016a4 4016b0 4016a4 4016b0 4016a4 4016b0 4016b0 4016b0 4016b0 4016b0 4016b0 4016b0 4016a4 4016b0 4016b0 4016b0 4016b0 4016b0 4016a4 4016b0 4016b0 4016b0 4016b0 4016b0 4016b0 4016b0 4016a4 4016a4 4016c6 401b1a]
2023/03/28 14:55:13 failed to read kernel-space stack traces: EOF
{PID:42909 UserStackID:232 KernelStackID:-14} seen 1 times
	232 [401697 4016a4 4016a4 4016b0 4016a4 4016b0 4016b0 4016a4 4016b0 4016a4 4016b0 4016a4 4016b0 4016a4 4016b0 4016b0 4016a4 4016a4 4016b0 4016a4 4016b0 4016a4 4016b0 4016b0 4016a4 4016b0 4016a4 4016b0 4016a4 4016b0 4016a4 4016c6 401b1a]
```

marselester:
I've annotated a few more stack traces.

```
401698 random "mov    %rdi,%rbx"
4016a4 instruction following fibNaive(n-2) call, i.e., "mov    %rax,%r12"
4016b0 instruction following fibNaive(n-1) call, i.e., "add    %r12,%rax"
4016c6 main
401b1a __libc_start_call_main

401697 random "push   %rbx"
4016a4 instruction following fibNaive(n-2) call, i.e., "mov    %rax,%r12"
4016b0 instruction following fibNaive(n-1) call, i.e., "add    %r12,%rax"
4016c6 main
401b1a __libc_start_call_main

4016a4 instruction following fibNaive(n-2) call, i.e., "mov    %rax,%r12"
4016b0 instruction following fibNaive(n-1) call, i.e., "add    %r12,%rax"
4016b7 "ret" instruction
4016c6 main
401b1a __libc_start_call_main
```

javierhonduco:
Just to confirm if I am understanding things correctly: you didn't expect to see addresses that aren't the start of functions, right? Let me know if I didn't get this right.

This is correct because frame-pointer based unwinders roughly work like this, in pseudocode:
```python
stack = []
stack.append(regs.ip) # Add current instruction pointer, to get the leaf frame at the exact location where the code was running when the sample was taken

while have_not_reached_bottom_of_stack(regs):
  return_address = *(regs.bp + 8) # Fetch return address
  stack.append(return_address)
  regs.bp = *(regs.bp) # Fetch the frame pointer from the parent frame
```

Let's check the kernel implementation for x86 to ensure that this matches our expectations:

* The first IP is added here: https://github.com/torvalds/linux/blob/3d7cb6b04c3f3115719235cc6866b10326de34cd/arch/x86/events/core.c#L2898
* The rest of the stack trace is added here: https://github.com/torvalds/linux/blob/3d7cb6b04c3f3115719235cc6866b10326de34cd/arch/x86/events/core.c#L2907-L2918


The reason why you are seeing pretty much all addresses without the leaf function in your stacks is that all this code is being executed by your CPU 😄. When we take a sample that's exactly the instruction we are executing. Not adding the current instruction pointer would result in inaccurate stack traces, and the leaf frame would not be totally correct.

We can argue whether this is a big problem or not, and the answer is... it depends! Most profiling probably can do with slightly inaccurate addresses, but microarchitectural performance work would not be possible without precise profile data.

For example, see [`perf annotate`](https://man7.org/linux/man-pages/man1/perf-annotate.1.html) which hydrates profiles with its assembly code.

marselester:
Right! That's an awesome explanation, thank you Javier 👍.

PR #1491, opened 2023-03-28
pkg/objectfile: do not cache base

zdyj3170101136:
even same mapping file.
The base address will changed when addr changed, so we should not cache it.
```
benchmark                  old ns/op     new ns/op     delta
BenchmarkELFObjAddr-10     2.47          8.65          +249.62%
```

### DIFF
--- a/pkg/objectfile/object_file.go
+++ b/pkg/objectfile/object_file.go
@@ -40,13 +40,12 @@ type ObjectFile struct {
 	BuildID string
 	ElfFile *elf.File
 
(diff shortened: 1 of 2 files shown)


javierhonduco:
Hi! Will take a look at this, might take us some days to get back to you but we will :)

kakkoyun:
Deprecated after #1507.

Thanks a lot for the work ❤️ 
We're still open to improving the mechanism around this. Let us know if you need something specific.

PR #1499, opened 2023-03-29
pkg/convert: Add JFR to pprof converter

brancz:


kakkoyun (approved):
Awesome 🏆 🥇 💯

I wonder. have you been able to find any document that specifies JFR?

pkg/convert/jfr.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	return b.profile, nil
+}
+
+func (b *builder) addJFRChunk(ctx context.Context, c parser.Chunk) error {
+	var event string
+	for _, e := range c.Events {
+		if as, ok := e.(*parser.ActiveSetting); ok {
+			// What is this for!?
```
v-thakkar, line 63: Is this comment still relevant or you meant to add some information there?

brancz:
I haven't but I was also not looking very much.

pkg/convert/jfr.go
```diff
@@ -0,0 +1,163 @@
...
+			}
+		}
+	}
+	if event != "cpu" {
+		return nil
+	}
+
+	for _, event := range extractExecutionSampleEvents(c.Events) {
+		if event.State.Name == "STATE_RUNNABLE" {
+			increaseSample(b.getOrCreateSample(event.StackTrace))
```
v-thakkar, line 75: This is fine for now. And probably you've already thought about this but I think we'll need to keep a cache of stack traces that have already been seen so that we don't end up with duplicate samples. I suppose better to handle that when we call `addJFRChunk`?

v-thakkar (approved):
LGTM :tada:

v-thakkar:
> I wonder. have you been able to find any document that specifies JFR?

AFAIK, it's not publicly documented. Most tool developers get information about it from the code in JDK.

pkg/convert/jfr.go
brancz: that cache already exists within the builder, if we've already seen a stack then the previously created sample is returned

pkg/convert/jfr.go
brancz: oh right, I need to update the comment, I actually realized half way through what this is for

kakkoyun (approved):
🚀 🚀 🚀

Issue #1510, opened 2023-03-30
Add command line package to interact with async-profiler


v-thakkar:
As part of #1115:
- [x] Revert https://github.com/parca-dev/parca-agent/pull/1464 as we have changed the idea of interacting with `async-profiler` so it's no longer required
- [x] Add `libasyncprofiler.so` and `jattach` in the container image (WIP @brancz)
- [x] Add Go pkg to interact with jattach and `libasyncprofiler.so` (replacement of `profiler.sh`) (WIP @v-thakkar - #1512) 
- [x] Add a logic to use fdtransfer
- [ ] Add an integration test

Issue #1508, opened 2023-03-30
Implement per-process Java profiler


brancz:
Part of  #1115 
Pseudo code

```
package javaprofiler

type Profiler struct {
    // ... fields
}

func New(
    pid int,
    w profiler.ProfileWriter,
    duration time.Duration,
) *Profiler {
    // ...
}

func (p *Profiler) Run(ctx context.Context) error {
    // essentially https://github.com/async-profiler/async-profiler/blob/32601bccd9e49adda9510a2ed79d142ac6ef0ff9/profiler.sh#L364-L380
    // And then convert JFR to pprof and write via profile writer
}

func (p *Profiler) profileLoop(ctx context.Context) (*profile.Profile, error) {
  // use async profiler low-level Go package, to collect JFR profile
  // convert JFR profile to pprof and return
}
```

cc @v-thakkar

v-thakkar:
For this, we can't use the same implementation as the current cpu profiler for other languages as we'll only be able to attach this profiler after metadata gets attached to the process and we know that it's a java process. This means that profiling will start for java processes after the first profiling round(s). So the implementation flow will look something like this:

- Find the java process and attach the metadata to it
- Start the java profiler for that process with certain parameters (required by async-profiler)
- Stop the java profiler in 2 cases: if parca-agent exits or if the process exits

Another way would be to check for each pid if it's a Java process, although this can get quite complicated and may waste some resources. So currently above solution looks best but let me know if there are some other ideas.

javierhonduco:
Sounds reasonable! Kemal and I were thinking on adding information about the runtime using the new facilities in the Process Information work

kakkoyun:
Deprecated.

We will use #1 to track the latest implementation.

PR #1512, opened 2023-03-30
Add Go package to interact with libasyncprofiler

v-thakkar:
As part of the ongoing work for #1510, this PR adds a go package to load the `libasyncprofiler.so` via `jattach`. It also loads `fdtransfer` to deal with the java processes with limited privileges.

**Test Plan:**
[This](https://github.com/parca-dev/parca-agent/pull/1590) toy program has been used to check the basic functionality provided by this package.

**Output of the above program:**
```
vaishali@pop-os:~/parca-agent$ ./asyncprofiler-test
paths are ok
fdtransfer server is running
Server ready at '59064'
context is ok
command is ok
start output is ok
stop output is ok
```

Note: The function `NewAsyncProfiler` only provides the basic functionality at the moment so that we can test the other functions. It will be reworked as part of #1508 to provide the fully functional API that can be used by the parca-agent.

### DIFF
--- a/pkg/asyncprofiler/asyncprofiler.go
+++ b/pkg/asyncprofiler/asyncprofiler.go
@@ -1,0 +1,152 @@
+// Copyright 2022-2023 The Parca Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package asyncprofiler
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"strconv"
+)
+
+// Action represents the various actions available for AsyncProfiler.
+type Action string
+
+var validActions = []string{
+	"start",
+	"resume",
+	"stop",
+}
+
+type ProfilerOptions map[string]string
+
+type AsyncProfiler struct {
+	jattachPath  string
+	libasyncPath string
+	pid          int
+	action       Action
+	options      ProfilerOptions
+}
+
+type ProfilerOption func(*AsyncProfiler)
+
+func WithEventType(eventType string) ProfilerOption {
+	return func(p *AsyncProfiler) {
+		p.options["event"] = eventType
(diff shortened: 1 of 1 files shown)


javierhonduco:
Sorry, I accidentally made it ready for review

v-thakkar:
As suggested by @javierhonduco, separated out the revert of SWIG generated package in #1518.

brancz (commented):
Looks good for a start! I think for this to be usable in the way that we want, we'll want the API sketched out in this issue: https://github.com/parca-dev/parca-agent/issues/1508

We should also add the sketched-out test of the description.

v-thakkar:
> Looks good for a start! I think for this to be usable in the way that we want, we'll want the API sketched out in this issue: #1508
> 
> We should also add the sketched-out test of the description.

Thanks for the review!

Yes, I was planning to do that after this PR is merged.  Especially `NewAsyncProfiler` function is not supposed to be as it is. I've just added it with minimal required functionality so that this package can be tested as we wanted to work on this whole feature in parts. In reality, it'll be a new package with a fully functional API using the functions of this package and the logic mentioned in #1508.

Are there any specific parts of this package's API that you would like me to elaborate more on in that issue? I can definitely add the this test there.

pkg/asyncprofiler/asyncprofiler.go
```diff
@@ -0,0 +1,122 @@
...
+
+var validActions = []string{
+	"start",
+	"resume",
+	"stop",
+	"dump",
+	"check",
+	"status",
+	"list",
+	"collect",
```
brancz, line 36: I'm not sure about all of these, but I know `collect` is not something that the async profiler itself does, it was just a meta-thing that the `profiler.sh` did. I think the only valid actions interacting with the java-agent are `start`, `stop`, and `resume`.

brancz (commented):
Got it, happy to go ahead as is then.

pkg/asyncprofiler/asyncprofiler.go
v-thakkar: You are correct about `collect`. I'll remove it and actually let me remove all the others we are not using just yet. Most of the others are valid though. Can be checked with something like `/home/vaishali/async-profiler/build/jattach 839511 load /home/vaishali/async-profiler/build/libasyncProfiler.so true <action>`

But I think it's better to add them when we add logic around using them so that it's not confusing for others. For example, checking `status` for debugging/some error handling, using `dump` to dump the output etc.

v-thakkar:
Moved to draft again as I'm adding a functionality to use the fdtransfer in the same PR.

v-thakkar:
The review comment has been addressed. And as suggested by @brancz , I've added the loading of fdtransfer in the same PR.

brancz (approved):

PR #1519, opened 2023-04-03
pkg/profiler/cpu: Fix condition for RequestUnwindInformation

javierhonduco:
This somehow slips through, my bad. This is not correct. The issue shows up if no other executables end up triggering a persist of the unwind table


kakkoyun (approved):
LGTM 🚀

PR #1529, opened 2023-04-05
pkg/objectfile: Tests for position independent executable

marselester:
I've mistakenly assumed that PIE is not supported in Parca Server and opened https://github.com/parca-dev/parca/pull/2910. @kakkoyun explained that the sampled addresses are normalized on Parca Agent side using the pprof package. Just in case I added some tests that check normalization of PIE and non-PIE programs.

Another minor thing: linter complained that `relocationSymbol` param is always blank, so I've added its resolution just like in the pprof but without an allocation, see https://github.com/google/pprof/blob/main/profile/encode.go#L258-L263.

```go
// If this a main linux kernel mapping with a relocation symbol suffix
// ("[kernel.kallsyms]_text"), extract said suffix.
// It is fairly hacky to handle at this level, but the alternatives appear even worse.
if strings.HasPrefix(m.File, "[kernel.kallsyms]") {
	m.KernelRelocationSymbol = strings.ReplaceAll(m.File, "[kernel.kallsyms]", "")
}
```

```
# new
BenchmarkKernelRelocationSymbol-12    	154684881	         7.796 ns/op	       0 B/op	       0 allocs/op
# strings.ReplaceAll
BenchmarkKernelRelocationSymbol-12    	23326599	        49.38 ns/op	       8 B/op	       1 allocs/op
```


pkg/objectfile/object_file.go
```diff
@@ -51,28 +50,53 @@ type ObjectFile struct {
...
 	f, err := os.Open(filePath)
 	if err != nil {
-		return nil, fmt.Errorf("error opening %s: %w", filePath, err)
+		return false, fmt.Errorf("error opening %s: %w", filePath, err)
 	}
 	defer f.Close()
 
 	// Read the first 4 bytes of the file.
 	var header [4]byte
-	if _, err = io.ReadFull(f, header[:]); err != nil {
```
marselester, line 62: `io.ReadFull` allocated, so I replaced it with `f.Read`.

kakkoyun (commented):
LGTM ❤️ 

Thanks for adding all the tests!

For completeness sake, could you please share the benchmark comparison as part of PR comment?

pkg/objectfile/object_file.go
marselester: Here is a difference between `f.Read` and `io.ReadFull`.
```
# new
BenchmarkIsELF-12    	  123748	      9700 ns/op	     128 B/op	       3 allocs/op
# old
BenchmarkIsELF-12    	  120458	      9864 ns/op	     132 B/op	       4 allocs/op
```

marselester:
Good point, I've added the benchmark results.

kakkoyun (approved):

PR #1531, opened 2023-04-06
build: Add async-profiler to tarball and container image artifacts

brancz:


scripts/download-async-profiler.sh
```diff
@@ -0,0 +1,31 @@
...
+version="${2:-2.9}"
+
+rm -rf "${target_dir}"
+mkdir -p "${target_dir}"
+
+wget -O - "https://github.com/async-profiler/async-profiler/releases/download/v${version}/async-profiler-${version}-linux-x64.tar.gz" | tar -xz -C "${target_dir}"
+
+mkdir -p "${target_dir}/x64/libc"
+mv "${target_dir}/async-profiler-${version}-linux-x64/build/jattach" "${target_dir}/x64/libc"
+mv "${target_dir}/async-profiler-${version}-linux-x64/build/fdtransfer" "${target_dir}/x64/libc"
```
v-thakkar, line 28: We don't need fdtransfer, no? It's an interface to access `perf_event_open` for profiling processes with low privileges. As parca-agent runs as a root user anyway, I think it's not required.

v-thakkar (commented):
I'll test this locally later today but looks good except for the comment inlined above. Thanks for working on this!

scripts/download-async-profiler.sh
brancz: You're absolutely right. Removed it.

v-thakkar (approved):

kakkoyun (approved):
LGTM 🥇

PR #1533, opened 2023-04-06
pkg/debuginfo: Fix extraction logic

brancz:
Previously the `stripDebuginfo` field was constantly overridden by whether or not the given binary had a text section or not, which lead to disregarding if stipping debuginfos was explicitly turned off.

Additionally, this was unsafe as it was overriding the global configuration, which could have been written concurrently.


v-thakkar (approved):
Good find. The code is also much clearer now with a separate function. LGTM!

PR #1534, opened 2023-04-06
Hidden flag to disable address normalization for debugging

marselester:
As per @brancz, the address normalization should happen on the Parca Server side.

> We definitely want to remove the normalization from Parca Agent if possible, as the pprof profiles it generates are currently broken with standard tooling (eg. regular pprof).

With a flag, we should be able to test the changes without interfering with release cycles on both sides.


brancz:
I like this idea a lot!

brancz (approved):

zdyj3170101136:
is this will fix https://github.com/parca-dev/parca-agent/issues/1187?

zdyj3170101136:
you must merge this pr https://github.com/parca-dev/parca-agent/pull/1196/files.
to let the extract debuginfo have enough information to compute base address.

brancz:
Not yet, this is just for experimenting with different approaches, but the goal is to normalize on the server-side, yes!

PR #1544, opened 2023-04-11
Return error only if none of the vdso file is found

v-thakkar:
We only care about either finding `vdso.so` or `vdso64.so`. So return an error only after we're done looking for both.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -36,7 +36,6 @@ require (
 	github.com/stretchr/testify v1.8.2
 	github.com/xyproto/ainur v1.3.2
 	go.uber.org/automaxprocs v1.5.2
-	go.uber.org/multierr v1.11.0
 	golang.org/x/exp v0.0.0-20230321023759-10a507213a29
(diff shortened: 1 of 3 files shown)


v-thakkar:
I think you're correct about keeping multierr as it'll give information about all the tried paths. I did this because user got confused with the error and thought that they would have to symlink. So yes, joining the errors or storing the paths that have been tried in a slice and returning it along with the error if func fails to open vdso files will be nice. I'll push the changes here.

Regarding why the error is happening, I haven't been able to reproduce it locally. Tried for both latest main and the release version it was reported for. So, maybe that's where we should start. One hunch I had was probably we're not reading the same kernel modules as the one running on the machine. But awaiting for user's confirmation on the same in the discord channel :D

Nonetheless, assigned it to you as discussed on Discord. Thanks for taking care of it.

v-thakkar:
Alright, in this particular case errors.Join doesn't seem to be improving errors any better. So closing this.

PR #1547, opened 2023-04-12
fix: prevent the panics when we try to retry with a different number of shards

kakkoyun:
cc @javierhonduco


javierhonduco (approved):
Amazing stuff!! LGTM!!

Issue #1543, opened 2023-04-11
Parca agent is not able to find vdso file


v-thakkar:
```
level=error name=parca-agent caller=main.go:472 msg="failed to initialize vdso cache" err="failed to open elf file:/usr/lib/modules/5.15.0-1033-aws/vdso/vdso.so, err:open /usr/lib/modules/5.15.0-1033-aws/vdso/vdso.so: no such file or directory; failed to open elf file:/usr/lib/modules/5.15.0-1033-aws/vdso/vdso64.so, err:open /usr/lib/modules/5.15.0-1033-aws/vdso/vdso64.so: no such file or directory"

```
vdso file seems to exists and parca-agent seem to be looking for and handling `vdso64.so` [here](https://github.com/parca-dev/parca-agent/blob/main/pkg/vdso/vdso.go#L46)
```
ls /usr/lib/modules/5.15.0-1033-aws/vdso
vdso32.so  vdso64.so  vdsox32.so
```
The bug has been reported for the v0.17.1.

v-thakkar:
@zdyj3170101136 I'm checking this and writing a test for vdso package but let me know if you have any idea of why this is failing. Thanks!

zdyj3170101136:
> /usr/lib/modules/5.15.0-1033-aws/vdso/vdso.so

from the information you provided, the vdso cache should correctly open the vdso file.

i have an assumption that the parca-agent process do not have enough priority to find these file.

may be some docker yaml config is wrong?

v-thakkar:
> > /usr/lib/modules/5.15.0-1033-aws/vdso/vdso.so
> 
> from the information you provided, the vdso cache should correctly open the vdso file.
> 
> i have an assumption that the parca-agent process do not have enough priority to find these file.
> 
> may be some docker yaml config is wrong?

User had confirmed that /lib/modules is correctly mapped inside the container. So, I don't think that's the issue. My another guess was agent checking the wrong kernel modules dir but we're waiting for user's reply on this.

Nonetheless, thanks for checking @zdyj3170101136 !

v-thakkar:
@javierhonduco Does https://github.com/parca-dev/parca-agent/pull/1547 fixes this? Happy to close this one in that case. :)

javierhonduco:
@v-thakkar No, the two issues are unrelated

rafaeljesus:
[...]

PR #1549, opened 2023-04-12
pkg/vdso: find vdso file from absolute path

zdyj3170101136:
this will fix find vdso file when running in docker.

and may be this will fix https://github.com/parca-dev/parca-agent/issues/1543


zdyj3170101136:
@v-thakkar cc

kakkoyun:
This is deprecated after recent changes. Please feel free to report if it doesn't work as you expected. Thanks a lot for your contribution.

PR #1583, opened 2023-04-24
unwinder/native: Remove polling mode

javierhonduco:
As it's not used or tested anymore, we rely on the on-demand unwind info generation.

Test Plan
=========

Ran the agent for a little while without issues


kakkoyun (approved):
LGTM 🥇

v-thakkar (approved):
LGTM!

PR #1587, opened 2023-04-26
ksym: Introduce disk-based symbolizer

javierhonduco:
Context
======

There are several use-cases, such as symbolization, that conceptually boil
down to a list of tuples, each formed of an identifier and a string, where
we want to efficiently find the string for a particular identifier.

A possible approach to solve this problem is to create a large list
in memory, sorted by identifier, so we can binary-search over the ids
and find the entry for which entry_i.Id <= Id < entry_i+1.Id.

While this works well and it's easy to understand and maintain, when
the list grows too large, this can be a large source of retained memory.
This issue is particularly bad when memory offloading (swap/zswap) is not
enabled, as the "cold" anonymous memory won't have a way to be moved to
secondary storage.

This implementation produces a simple binary format that's easy to write
and read, but most importantly, it should be efficient to query.

```
┌─────────┬────────────────────────────┬────────────────────────────────────────────┐
│         │                            │                                            │
│ Header  │  Strings with nul endings  │  Sorted ids + meta information on strings  │
│         │                            │                                            │
└─────────┴────────────────────────────┴────────────────────────────────────────────┘
```
The strings aren't deduplicated or optimized in any way, to reduce the memory
usage during the write phase.

The file is read with `mmap(2)`, to avoid performing any read system calls
while binary searching over the identifiers, and leveraging the caching layer
of the filesystem. As we now have a backing file, rather than being anonymous
memory, the OS can remove cached pages if there's a need for more memory.

Test Plan
=======

- ksym tests pass;
- ran the agent for 1h without issues, spot-checked several kernel
  symbols and they looked good;

In terms of efficiency it would be best to check on Demo / prod, but
this is the early data I've gotten after running the Agent for 10mins

### Memory

**before**
```
[javierhonduco@fedora parca-agent]$  cat /proc/`pidof parca-agent-debug`/status | grep -i rss
VmRSS:    102212 kB
RssAnon:           62148 kB
RssFile:           40064 kB
RssShmem:              0 kB
```

**after**
```
[javierhonduco@fedora parca-agent]$ cat /proc/`pidof parca-agent-debug`/status | grep -i rss
VmRSS:     83996 kB
RssAnon:           39324 kB
RssFile:           40192 kB
RssShmem:           4480 kB
```

### CPU
<details>

**before**
image

**after**

image

No significant difference in CPU usage

</details>

This is roughly a 20% decrease in RSS, which is expected for the number
of symbols that my box has. The more symbols the bigger the savings! :)

### DIFF
--- a/pkg/ksym/optimizer.go
+++ b/pkg/ksym/optimizer.go
@@ -1,0 +1,294 @@
+// Copyright 2022-2023 The Parca Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ksym
+
+import (
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"sort"
+	"unsafe"
+
+	"golang.org/x/exp/mmap"
+)
+
+// There are several use-cases, such as symbolization, that conceptually boil
+// down to a list of tuples, each formed of an identifier and a string, where
+// we want to efficiently find the string for a particular identifier.
+//
+// A possible approach to solve this problem is to create a large list
+// in memory, sorted by identifier, so we can binary-search over the ids
+// and find the entry for which entry_i.Id <= Id < entry_i+1.Id.
+//
+// While this works well and it's easy to understand and maintain, when
+// the list grows too large, this can be a large source of retained memory.
+// This issue is particularly bad when memory offloading (swap/zswap) is not
+// enabled, as the "cold" anonymous memory won't have a way to be moved to
+// secondary storage.
+//
+// This implementation produces a simple binary format that's easy to write
+// and read, but most importantly, it should be efficient to query.
+//
+// ┌─────────┬────────────────────────────┬────────────────────────────────────────────┐
+// │         │                            │                                            │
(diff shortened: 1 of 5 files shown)


pkg/ksym/optimizer.go
```diff
@@ -0,0 +1,292 @@
...
+	Magic           uint32
+	Version         uint32
+	AddressesOffset uint32
+	AddressesCount  uint32
+}
+
+type entry struct {
+	address uint64
+	offset  uint32
+	len     uint32
```
brancz, line 82: if we null-terminate the strings, why not just read until we hit the null? we'd save 25% of size of entries

brancz (approved):
Love it! Super simple and effective!

pkg/ksym/optimizer.go
javierhonduco: I've been thinking about this and I like having the size so we can pre-allocate the buffer to read the bytes that make the string with the right size.

But now that you mention this, I was thinking that perhaps we can use a `uint8` for the len, what do you think?

brancz: What I've done elsewhere is write the largest string ever seen into the header, and allocate that size. I'm pretty sure the worst thing we might allocate with kernel strings is 1k bytes (and probably much less).

javierhonduco: Got it! I agree that the maximum length kernel symbol won't be too large. I think I will keep the code like this because I would like to make the read path allocate (and copy) as little memory as possible. But I am happy to continue exploring optimizations if the kernel symbolizer is not as efficient as we expect in production! 😄 

Meanwhile, I have changed the `len` to use 2 bytes instead of 4 bytes, while I hope there won't be kernel symbols of more than 2**8 = 256 characters, I wanna be sure we don't hit that. Some compilers / post compiler optimizers change function names and add suffixes, increasing some function's names

brancz: sgtm

kakkoyun (approved):
Nice 👌

Sylfrena:
Love it! Thanks for the detailed explanation in the PR desc! 💯

PR #1611, opened 2023-05-08
Fetch labels at the first encounter with a process

kakkoyun:
### What?

copilot:summary

### How?

copilot:walkthrough


pkg/metadata/labels/manager.go
```diff
@@ -31,6 +31,8 @@ import (
 	"github.com/parca-dev/parca-agent/pkg/metadata"
 )
 
+// TODO(kakkoyun): Re-evaluate the advantages and disadvantages of caching per profiler.
```
javierhonduco, line 34: What are the use-cases you have in mind for separate caching per profiler?

pkg/profiler/profiler.go
```diff
@@ -118,16 +118,17 @@ type Profile struct {
 
 // TODO(kakkoyun): Unify PID types.
```
javierhonduco, line 119: Should this be removed?

cmd/parca-agent/main.go
```diff
@@ -607,7 +607,7 @@ func run(logger log.Logger, reg *prometheus.Registry, flags flags) error {
 				var lastError error
 				var link, profilingStatus string
 				for _, prflr := range profilers {
-					lbls := labelsManager.Labels(prflr.Name(), uint64(pid))
```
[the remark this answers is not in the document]
kakkoyun: We have another to-do item to unify all the PID types to int32. Right now, I want to sync these two components. We send a subsequent PR to fix it.

pkg/profiler/profiler.go
kakkoyun: Noop. As we discussed above, we need to convert everything to int32.

pkg/metadata/labels/labels.go
```diff
@@ -0,0 +1,34 @@
...
+// limitations under the License.
+
+package labels
+
+import (
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/model/labels"
+)
+
+const profilerName = "__name__"
```
javierhonduco, line 21: Python vibes only

javierhonduco (approved):

Sylfrena (approved):
nit: any reason why we don't use `profiler` and `label` instead of `prflr` and `lbl` ?

but lgtm otherwise, feel free to merge

kakkoyun:
> nit: any reason why we don't use `profiler` and `label` instead of `prflr` and `lbl` ?

It's just a preference, in sync with Go conventions. If the scope of the variable is narrow enough, I tend to use shorter names.
https://research.swtch.com/names

maxbrunet:
> > nit: any reason why we don't use `profiler` and `label` instead of `prflr` and `lbl` ?
> 
> It's just a preference, in sync with Go conventions. If the scope of the variable is narrow enough, I tend to use shorter names. [research.swtch.com/names](https://research.swtch.com/names)

There is actually a technical reason for `prflr`, it avoids shadowing the import which already uses the name: `"github.com/parca-dev/parca-agent/pkg/profiler"`

PR #1620, opened 2023-05-09
cpu.go: Show error if there's an error loading the BPF code

javierhonduco:
cpu.go: Show error if there's an error loading the BPF code

## Test Plan

CI

### DIFF
--- a/pkg/profiler/cpu/cpu.go
+++ b/pkg/profiler/cpu/cpu.go
@@ -244,7 +244,7 @@ func loadBpfProgram(logger log.Logger, reg prometheus.Registerer, debugEnabled, 
 			break
 		}
 	}
-	level.Error(logger).Log("msg", "Could not create unwind info shards")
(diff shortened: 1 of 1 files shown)


v-thakkar (approved):

pkg/profiler/cpu/cpu.go
```diff
@@ -244,7 +244,7 @@ func loadBpfProgram(logger log.Logger, reg prometheus.Registerer, debugEnabled,
 			break
```
javierhonduco, line 244: Actually, will change the log to here so we see the error every time in case it changes

javierhonduco:
@v-thakkar I was going to change things a little bit, will do in another PR

v-thakkar:
@javierhonduco Sure, sounds good. Maybe for the next time, can you put PR in the draft stage then so that we know that it's not ready to merge for any reason :)

PR #1622, opened 2023-05-10
Upgrade libbpfgo to `80f41e18e690`

javierhonduco:
Upgrade libbpfgo to [v0.4.8-libbpf-1.2.0.0.20230509162948-80f41e18e690 from v0.4.7-libbpf-1.2.0-b2e29a1](https://github.com/aquasecurity/libbpfgo/compare/80f41e18e690..b2e29a1).

libbpf was bumped from [6597330c45d185381900037f0130712cd326ae59 to fbd60dbff51c870f5e80a17c4f2fd639eb80af90](https://github.com/libbpf/libbpf/compare/b2e29a10265eff85afa61b11c9588f49111df79a..fbd60dbff51c870f5e80a17c4f2fd639eb80af90)

Test Plan
=======

Ran it for a little while + tests are all green

### DIFF
--- a/pkg/profiler/cpu/cpu.go
+++ b/pkg/profiler/cpu/cpu.go
@@ -504,7 +504,7 @@ func (p *CPU) Run(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("failed to init perf buffer: %w", err)
 	}
-	perfBuf.Start()
+	perfBuf.Poll(250)
 	go p.listenEvents(ctx, eventsChan, lostChannel, requestUnwindInfoChannel)
 
 	go func() {
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.20
 
 require (
 	github.com/alecthomas/kong v0.7.1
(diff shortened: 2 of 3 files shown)


javierhonduco:
Supersedes https://github.com/parca-dev/parca-agent/pull/1599

pkg/profiler/cpu/cpu.go
```diff
@@ -504,7 +504,7 @@ func (p *CPU) Run(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("failed to init perf buffer: %w", err)
 	}
-	perfBuf.Start()
+	perfBuf.Poll(250)
```
kakkoyun, line 507: nit/question: Is `250` the default value? So there's no behavioral change, right?

kakkoyun (approved):
LGTM 🥇

pkg/profiler/cpu/cpu.go
javierhonduco: Good question! The default value is currently 300. Perhaps we can lower it to 250 and see what the overhead is in production. We can fine-tune this value to find the right trade-off for us regarding CPU usage and the ability to catch short-running processes

Issue #1226, opened 2023-01-17
Do not normalize address for JIT symbol


zdyj3170101136:
in kernel function dso__load_perf_map,
it read address from file directly and use it to build symbols.

so if we normalize addr and use it to get jit symbol in parca-agent,
we would get wrong result.
![截屏2023-01-17 下午6 29 52](https://user-images.githubusercontent.com/52686065/212875503-af242d74-7429-43f2-a009-e70cff413127.png)

kakkoyun:
@zdyj3170101136 I will close the issue, however, I'm open to discussing it further if you need more help.

PR #1651, opened 2023-05-16
Do not normalize addresses from JIT'd code

javierhonduco:
Addresses written to "perf maps" or jitdump are absolute to the processes, so currently symbolization is broken.

This the first step to unblock the broken JIT symbolization. There is more work to do, will open an issue to track this.

Test Plan
=========

The current integration tests `make test/profiler` pass. Will add an integration test once we have the mixed-mode unwinder ready.

image

### DIFF
--- a/pkg/process/maps.go
+++ b/pkg/process/maps.go
@@ -341,9 +341,10 @@ func (m *Mapping) computeBase(addr uint64) error {
 
 // ConvertToPprof converts the Mapping to a pprof profile.Mapping.
 func (m *Mapping) ConvertToPprof() *profile.Mapping {
-	buildID := "<unknown build id>"
-	path := "<unknown path>"
-	// ^ could be JIT segments, validate this!
+	buildID := "unknown"
+	// TODO: Maybe add detection for JITs that use files.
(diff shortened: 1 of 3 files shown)


pkg/process/maps.go
```diff
@@ -341,9 +341,10 @@ func (m *Mapping) computeBase(addr uint64) error {
 
 // ConvertToPprof converts the Mapping to a pprof profile.Mapping.
 func (m *Mapping) ConvertToPprof() *profile.Mapping {
-	buildID := "<unknown build id>"
-	path := "<unknown path>"
-	// ^ could be JIT segments, validate this!
+	buildID := "unknown"
+	// TODO: Maybe add detection for JITs that use files.
+	path := "jit"
```
kakkoyun, line 346: How are we sure that this is always the case?

kakkoyun (commented):
I'm planning some other changes in the same parts. I'll wait for your changes.

javierhonduco:
This should fix https://github.com/parca-dev/parca-agent/issues/1226 cc @zdyj3170101136

PR #1652, opened 2023-05-16
internal/pprof: Remove unused vendored pprof packages

javierhonduco:
These keep on popping when I grep on the codebase and I thought we wouldn't use most of it. Let's avoid code we don't need or use.


javierhonduco:
> There must be a lot of references in the Makefile, and some other CI check scripts like license checks etc. Let's remove them as part of this PR as well.

Removed the vendoring bits, but left the licensing and not formatting atm.

kakkoyun (approved):

kakkoyun:
> > There must be a lot of references in the Makefile, and some other CI check scripts like license checks etc. Let's remove them as part of this PR as well.
> 
> Removed the vendoring bits, but left the licensing and not formatting atm.

😅 Just out of curiosity, any particular reasons?

javierhonduco:
Licensing, that we didn't author these files, but I am not a lawyer 😝 . About formatting, we can add it, but right now we have made not that many (maybe zero?) modifications to these files

kakkoyun:
> Licensing, that we didn't author these files, but I am not a lawyer 😝 . About formatting, we can add it, but right now we have made not that many (maybe zero?) modifications to these files

I think you've misunderstood what I have said 😊 We have a couple of scripts or configurations and actions to EXCLUDE checking the license headers and formatting the files in these packages in our CI. I wanted to tell you that we should remove those references to `internal/pprof`. It could create confusion for future readers of the codebase :)

PR #1669, opened 2023-05-18
pkg/perf: Reduce allocs in ReadMap when parsing a large file

marselester:
### Why?

I watched @brancz optimize `ReadMap` in https://github.com/parca-dev/parca-agent/pull/1650 and wanted to see if `stringConverter` I used in [systemd](https://github.com/parca-dev/parca-agent/blob/main/pkg/discovery/systemd/decoder.go#L205) would make a difference.

On a large file (108K lines) the change looks promising.

```
$ go test -count=10 -run=^$ -bench=^BenchmarkPerfMapParseBig$ -benchmem ./pkg/perf > newbig.txt
$ benchstat oldbig.txt newbig.txt
name                old time/op    new time/op    delta
PerfMapParseBig-12    28.8ms ± 2%    19.6ms ± 1%  -31.87%  (p=0.000 n=10+10)

name                old alloc/op   new alloc/op   delta
PerfMapParseBig-12    25.9MB ± 0%     8.1MB ± 0%  -68.70%  (p=0.000 n=9+9)

name                old allocs/op  new allocs/op  delta
PerfMapParseBig-12      108k ± 0%        0k ± 0%  -99.99%  (p=0.000 n=10+10)
```

But on a small file (28 lines) the performance gets worse due to `stat` syscall overhead and using less than 64 array.

```
$ go test -count=10 -run=^$ -bench=^BenchmarkPerfMapParse$ -benchmem ./pkg/perf > new.txt
$ benchstat old.txt new.txt
name             old time/op    new time/op    delta
PerfMapParse-12    16.6µs ± 1%    18.2µs ± 1%   +9.42%  (p=0.000 n=9+10)

name             old alloc/op   new alloc/op   delta
PerfMapParse-12    7.76kB ± 0%    8.61kB ± 0%  +10.93%  (p=0.000 n=10+10)

name             old allocs/op  new allocs/op  delta
PerfMapParse-12      36.0 ± 0%      38.0 ± 0%   +5.56%  (p=0.000 n=10+10)
```

### What?

copilot:summary

### How?

copilot:walkthrough


pkg/perf/conv.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// stringConverter converts bytes to strings with less allocs.
+// The idea is to accumulate bytes in a buffer with specified capacity
+// and create strings with unsafe.String using bytes from a buffer.
+// For example, 10 "fizz" strings written to a 40-byte buffer
+// will result in 1 alloc instead of 10.
+//
+// Once a buffer is filled, strings will be allocated as usually.
+type stringConverter struct {
```
marselester, line 22: This is just like https://github.com/parca-dev/parca-agent/blob/main/pkg/discovery/systemd/decoder.go#L205, but it doesn't allocate more buffers when the initial one fills up.

brancz:
Given the absolute values are very small for small files I think this makes sense!

kakkoyun (approved):

PR #1670, opened 2023-05-18
bpf: Add mixed-mode unwinding for JIT sections

Sylfrena:
This patch introduces support for mixed-mode unwinding for binaries with mixed JIT and AoT(ahead of time) compiled sections with the `dwarf-unwinding-mixed` flag.

As of now, DWARF-based stackwalking discards or skips JIT sections of binaries resulting in incomplete/incorrect stacks. Currently, we can only unwind JITed sections in binaries that retain frame pointers through FP-based stackwalking mode.

This patch adds a JIT unwinder for walking JITed sections without frame pointers. The agent bpf unwinder can now switch between JIT-unwinding and DWARF-unwinding modes while profiling stripped binaries with both JIT and Ahead-of-Time sections.

Test Plan
=========

The current integration tests `make test/profiler` pass(both with and without the flag enabled).

```
$ make test/integration GO=$(which go)
=== RUN   TestCPUProfilerWorks
--- PASS: TestCPUProfilerWorks
PASS
ok  	github.com/parca-dev/parca-agent/test/integration
```

Stacks are correctly symbolised in Parca for the `basic-cpp-jit-no-fp` toy jit binary.

image

### DIFF
--- a/bpf/cpu/cpu.bpf.c
+++ b/bpf/cpu/cpu.bpf.c
@@ -454,6 +454,9 @@ static __always_inline enum find_unwind_table_return find_unwind_table(chunk_inf
     if (offset != NULL) {
       *offset = load_address;
     }
+
+    // "type" here is set in userspace in our `proc_info` map to indicate JITed and special sections,
+    // It is not something we get from procfs.
     if (type == 1) {
       return FIND_UNWIND_JITTED;
     }
@@ -719,10 +722,11 @@ int walk_user_stacktrace_impl(struct bpf_perf_event_data *ctx) {
 
       err = bpf_probe_read_user(&next_fp, 8, (void *)unwind_state->bp);
       if (err < 0) {
-        //TODO(sylfrena):
-        //for some weird reason commenting out this and the next err log line results in a panic
-        LOG("[debug] i=%d, err = %d && rbp = %llx && ra=%llx", i, err, next_fp, ra);
-        return false;
+        // TODO(sylfrena):
+        // For some weird reason commenting out this and the next err log line results in a panic
+        // Using more than 3 arguments also results in a panic in some older kernels because of https://github.com/libbpf/libbpf/blob/f7eb43b90f4c8882edf6354f8585094f8f3aade0/src/bpf_helpers.h#L287-L289
+        LOG("[debug] err = %d",  err);
+        return 0;
       }
 
       //LOG("[debug]  i=%d, err = %d && rbp = %llx && ra=%llx", i, err, next_fp, ra);
@@ -731,14 +735,15 @@ int walk_user_stacktrace_impl(struct bpf_perf_event_data *ctx) {
       err = bpf_probe_read_user(&ra, 8, (void *)unwind_state->bp + 8);
       if (err < 0) {
         //TODO(sylfrena)
-        //for some weird reason commenting out this and the next err log line results in a panic
-        LOG("[debug] reading ra i=%d, err = %d && rbp = %llx && ra=%llx", i, err, next_fp, ra);
-        return false;
+        // For some weird reason commenting out this and the next err log line results in a panic
+        // Using more than 3 arguments also results in a panic in some older kernels because of https://github.com/libbpf/libbpf/blob/f7eb43b90f4c8882edf6354f8585094f8f3aade0/src/bpf_helpers.h#L287-L289
+        LOG("[debug] err = %d", err);
+        return 0;
       }
 
       if (next_fp == 0) {
         LOG("[info] found bottom frame while walking JITed section");
-        return true;
+        return 1;
       }
 
       // TODO(sylfrena): add comments to explain calculations
--- a/bpf/cpu/cpu.bpf.c
+++ b/bpf/cpu/cpu.bpf.c
@@ -84,6 +84,7 @@ enum stack_walking_method {
 struct unwinder_config_t {
   bool filter_processes;
   bool verbose_logging;
+  bool mixed_stack_walking;
 };
 
 struct unwinder_stats_t {
@@ -186,6 +187,7 @@ typedef struct {
   u64 bp;
   u32 tail_calls;
   stack_trace_t stack;
+  bool jitted; // set to true during JITed unwinding; false unless mixed-mode unwinding is enabled
 } unwind_state_t;
 
 // A row in the stack unwinding table for x86_64.
@@ -573,14 +575,15 @@ static __always_inline bool retrieve_task_registers(u64 *ip, u64 *sp, u64 *bp) {
 // right now using both.
 static __always_inline bool has_fp(u64 current_fp) {
   u64 next_fp;
+  u64 ra;
 
   for (int i = 0; i < MAX_STACK_DEPTH; i++) {
     int err = bpf_probe_read_user(&next_fp, 8, (void *)current_fp);
+    bpf_probe_read_user(&ra, 8, (void *)current_fp + 8);
     if (err < 0) {
       // LOG("[debug] fp read failed with %d", err);
       return false;
     }
-
     // Some cpp binaries, such as testdata/out/basic-cpp
     // seem to have rbp set to 1 in the bottom frame. This
     // does not comply with the x86_64 ABI.
@@ -664,6 +667,7 @@ SEC("perf_event")
 int walk_user_stacktrace_impl(struct bpf_perf_event_data *ctx) {
   u64 pid_tgid = bpf_get_current_pid_tgid();
   int user_pid = pid_tgid;
+  int err = 0;
 
   bool reached_bottom_of_stack = false;
   u64 zero = 0;
@@ -675,6 +679,7 @@ int walk_user_stacktrace_impl(struct bpf_perf_event_data *ctx) {
   }
 
   for (int i = 0; i < MAX_STACK_DEPTH_PER_PROGRAM; i++) {
+    LOG("[debug] Within unwinding machinery loop");
     LOG("## frame: %d", unwind_state->stack.len);
 
     LOG("\tcurrent pc: %llx", unwind_state->ip);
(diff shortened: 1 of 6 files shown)


bpf/cpu/cpu.bpf.c
```diff
@@ -675,19 +679,82 @@ int walk_user_stacktrace_impl(struct bpf_perf_event_data *ctx) {
...
+
+      if (next_fp == 0) {
+        LOG("[info] found bottom frame while walking JITed section");
+        return true;
+      }
+
+      // TODO(sylfrena): add comments to explain calculations
+      unwind_state->sp = unwind_state->bp + 16;
+      unwind_state->bp = next_fp;
+      unwind_state->ip = ra - 1;
```
javierhonduco, line 747: I think this deserves an explanation, even if it's a short one-liner! Now we totally understand what we are doing here, and most importantly, _why_, but in some months... I am not sure 😄

bpf/cpu/cpu.bpf.c
```diff
@@ -675,19 +679,82 @@ int walk_user_stacktrace_impl(struct bpf_perf_event_data *ctx) {
...
+      len = unwind_state->stack.len;
+
+      // add ra for frame
+      if (len >= 0 && len < MAX_STACK_DEPTH) {
+        unwind_state->stack.addresses[len] = ra;
+      }
+
+      unwind_state->stack.len++;
```
javierhonduco, line 755: nit: Can you move this inside the bound checks? Sorry I left it outside before you touched the code!

bpf/cpu/cpu.bpf.c
```diff
@@ -747,15 +815,28 @@ int walk_user_stacktrace_impl(struct bpf_perf_event_data *ctx) {
       break;
     }
 
+    //LOG("[debug] Switching to mixed-mode unwinding? %d", unwind_state->jitted);
```
javierhonduco, line 818: I think this could be a nice comment to have a log when we switch from the FP -> DWARF unwinder!

bpf/cpu/cpu.bpf.c
```diff
@@ -747,15 +815,28 @@ int walk_user_stacktrace_impl(struct bpf_perf_event_data *ctx) {
...
     // Add address to stack.
     u64 len = unwind_state->stack.len;
     // Appease the verifier.
     // For some reason bailing out here if the condition is not true does
     // not work?
-    if (len >= 0 && len < MAX_STACK_DEPTH) {
-      unwind_state->stack.addresses[len] = unwind_state->ip;
+
+    // This is for the case when we are NOT switching unwinding from JIT to DWARF section
+    // i.e. unwind_state->jitted holds false
```
javierhonduco, line 827: nit: I think this is clear enough in the next line 😄

bpf/cpu/cpu.bpf.c
```diff
@@ -747,15 +815,28 @@ int walk_user_stacktrace_impl(struct bpf_perf_event_data *ctx) {
...
+      if (len >= 0 && len < MAX_STACK_DEPTH) {
+        unwind_state->stack.addresses[len] = unwind_state->ip;
+
+        unwind_state->stack.len++;
+      }
     }
 
+    // Set unwind_state->jitted to false once we have checked for switch from JITed unwinding to DWARF unwinding
+    unwind_state->jitted = false;
+    //LOG("[debug] Switched to mixed-mode DWARF unwinding? %d", unwind_state->jitted);
```
javierhonduco, line 838: nit: stray comment?

bpf/cpu/cpu.bpf.c
```diff
@@ -870,6 +951,8 @@ int walk_user_stacktrace_impl(struct bpf_perf_event_data *ctx) {
     // > stack frame by setting the frame > pointer to zero.
     //
     // https://refspecs.linuxbase.org/elf/x86_64-abi-0.99.pdf
+    LOG("[debug] reachy reachy bottom %d", unwind_state->jitted);
```
javierhonduco, line 954: stray comment? 😄 It will always be false here anyways, no? We always expect to go from the JIT -> AoT transition

bpf/cpu/cpu.bpf.c
```diff
@@ -918,6 +1001,7 @@ static __always_inline bool set_initial_state(struct pt_regs *regs) {
   // we aren't reading garbage data.
   unwind_state->stack.len = 0;
   unwind_state->tail_calls = 0;
+  unwind_state->jitted = false;
```
javierhonduco, line 1004: Yesssss

bpf/cpu/cpu.bpf.c
```diff
@@ -1001,18 +1085,23 @@ int profile_cpu(struct bpf_perf_event_data *ctx) {
...
 
       LOG("[warn] IP 0x%llx not covered, could be a new/JIT mapping.", unwind_state->ip);
+
       if (unwind_table_result == FIND_UNWIND_MAPPING_NOT_FOUND) {
         request_refresh_process_info(ctx, user_pid);
         bump_unwind_error_pc_not_covered();
+        return 1;
       } else if (unwind_table_result == FIND_UNWIND_JITTED) {
-        bump_unwind_error_jit();
+         if (!unwinder_config.mixed_stack_walking) {
```
javierhonduco, line 1094: Thanks for being careful with the error handling! Let's add more granular metrics (unwinding errors per JIT attempts, "pure" DWARF, etc)  it in another PR 😄

javierhonduco (approved):
Awesome work! 💞 Thanks so much for the detailed description. Let a couple of comments, mostly style nits.

Just to make this PR super duper awesome ✨ could you run the automated code formatted for the C code with `make bpf/fmt` and also fetch the unwinding metrics with the toy JIT to verify that the unwinding errors are within what we expect? 🙏 

Pre-emptively approving because this is super close to the finishing line and I am off next week 😄

bpf/cpu/cpu.bpf.c
```diff
@@ -573,14 +575,15 @@ static __always_inline bool retrieve_task_registers(u64 *ip, u64 *sp, u64 *bp) {
 // right now using both.
 static __always_inline bool has_fp(u64 current_fp) {
   u64 next_fp;
+  u64 ra;
 
   for (int i = 0; i < MAX_STACK_DEPTH; i++) {
     int err = bpf_probe_read_user(&next_fp, 8, (void *)current_fp);
+    bpf_probe_read_user(&ra, 8, (void *)current_fp + 8);
```
brancz, line 582: not sure, but in all the verifier errors I read something about `invalid read from stack off -120+0 size 8` ... 

1) should we not be checking the returned error here?
2) are we sure the `current_fp + 8` is correct?

javierhonduco: > should we not be checking the returned error here?

Yes, definitely (this should not be resulting in the invalid reads, though)

> are we sure the current_fp + 8 is correct?

It should, in x86 the return address is always 8 bytes ahead from the current frame pointer

bpf/cpu/cpu.bpf.c
Sylfrena: more like an explanation for future sumera who is very likely to get confused haha 😅 
(also ditto for above)

unless you are super against it, I would love to keep them :D

bpf/cpu/cpu.bpf.c
```diff
@@ -675,19 +679,82 @@ int walk_user_stacktrace_impl(struct bpf_perf_event_data *ctx) {
...
+
     if (unwind_table_result == FIND_UNWIND_JITTED) {
-      LOG("JIT section, stopping");
-      return 1;
+      if (!unwinder_config.mixed_stack_walking) {
+        LOG("JIT section, stopping");
+        return 1;
+      }
+
+      LOG("[debug] Unwinding JITed stacks");
```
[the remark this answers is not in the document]
Sylfrena: 🥳

bpf/cpu/cpu.bpf.c
```diff
@@ -675,19 +679,82 @@ int walk_user_stacktrace_impl(struct bpf_perf_event_data *ctx) {
   }
 
   for (int i = 0; i < MAX_STACK_DEPTH_PER_PROGRAM; i++) {
+    LOG("[debug] Within unwinding machinery loop");
```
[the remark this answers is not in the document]
Sylfrena: yes, makes sense! will remove it, thanks!

bpf/cpu/cpu.bpf.c
Sylfrena: I am not sure I understand this one. There already is a similar comment a few lines below when we switch
` // LOG("[debug] Switched to mixed-mode DWARF unwinding? %d", unwind_state->unwinding_jit);`
Do you want me to uncomment it? Or just add a comment instead of enclosing it in a log?

bpf/cpu/cpu.bpf.c
javierhonduco: If this is not being fixed in this PR, can you open an issue?

bpf/cpu/cpu.bpf.c
```diff
@@ -675,19 +682,84 @@ int walk_user_stacktrace_impl(struct bpf_perf_event_data *ctx) {
...
     u64 offset = 0;
+
     chunk_info_t *chunk_info = NULL;
     enum find_unwind_table_return unwind_table_result = find_unwind_table(&chunk_info, user_pid, unwind_state->ip, &offset);
 
     if (unwind_table_result == FIND_UNWIND_JITTED) {
-      LOG("JIT section, stopping");
-      return 1;
+      if (!unwinder_config.mixed_stack_enabled) {
+        LOG("JIT section, stopping. Please enable mixed-mode unwinding with the --dwarf-unwinding-mixed=true to profile JITed stacks.");
```
javierhonduco, line 699: Glad you added this!

javierhonduco (approved):
Re-tested locally, works great! Let's ensure we don't forget about the follow-ups we discussed (explanation of the `ra - 1` trick, fixing/adding the metrics, and tacking the TODOs)

image

bpf/cpu/cpu.bpf.c
javierhonduco: Let's uncomment it!

kakkoyun (approved):

v-thakkar:
Late to the party but awesome work :tada:

PR #1688, opened 2023-05-24
pkg/cpu: Fix variable shadowing of error value

Sylfrena:


brancz (approved):
@javierhonduco when you're back from PTO please do double check that this was the intention, but reading through this, this is what we determined is what we probably wanted to do with this function

PR #1706, opened 2023-05-30
Make sure pprof profile IDs are correct

kakkoyun:
### Why?

Fixes the following issue we have been seeing sometimes:
```
level=warn name=parca-agent ts=2023-05-30T12:10:41.404303976Z caller=batch_remote_write_client.go:139 msg="batch write client failed to send profiles" count=12 err="rpc error: code = InvalidArgument desc = invalid profile: function id is not sequential"
```

`pprof` format excepts all the IDs are monotonically increasing. This PR makes sure a single component manages the ID creation. It introduces a new `ProfileToPprofConverter` type and ensures the agent profile is converted to a `pprof` profile. (It could be better implementation, I'm aware). 

I intend to collect all the conversion logic to a single place and lean against our homegrown profile type so that the migration will be easier when we introduce Parca Wire Format (or similar).

### What?
copilot:summary

### How?
copilot:walkthrough

### Test Plan
1. Smoke testing.

### TODO
- [ ] Add a unit test in subsequent PRs

copilot:poem

cc @Sylfrena


pkg/convert/pprof.go
```diff
@@ -0,0 +1,139 @@
...
+			Type: "cpu",
+			Unit: "nanoseconds",
+		},
+		Period: periodNS,
+	}
+	if len(prs) == 0 {
+		return prof, nil
+	}
+
+	// Build Profile from samples, locations and mappings.
```
brancz, line 60: this needs to happen exactly the opposite way, first mappings and functions need to be normalized, then locations then samples, that's the only way we can ensure IDs are correct at the end

brancz, line 60: this needs to happen exactly the opposite way, first mappings and functions need to be normalized, then locations then samples, that's the only way we can ensure IDs are correct at the end

pkg/convert/pprof.go
```diff
@@ -0,0 +1,139 @@
...
+	for _, pr := range prs {
+		for _, s := range pr.Samples {
+			prof.Sample = append(prof.Sample, s.Sample)
+		}
+	}
+
+	// Locations.
+	for _, pr := range prs {
+		for i, l := range pr.Locations {
```
brancz, line 69: is there anything that ensures that there are no duplicate locations?

pkg/convert/pprof.go
```diff
@@ -0,0 +1,139 @@
...
+			l.ID = uint64(i) + 1
+			l.Location.Mapping = c.mappingToPprof(l.Mapping)
+			prof.Location = append(prof.Location, l.Location)
+		}
+	}
+
+	// User mappings.
+	for _, pr := range prs {
+		for i, m := range c.mappingsToPprof(pr.UserMappings) {
```
brancz, line 78: same as locations

pkg/convert/pprof.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// Kernel mappings.
+	lastProfile := prs[len(prs)-1]
+	kernelMapping := c.mappingToPprof(lastProfile.KernelMapping)
+	kernelMapping.ID = uint64(len(prof.Mapping)) + 1
+	prof.Mapping = append(prof.Mapping, kernelMapping)
+
+	// Symbolized functions.
+	for _, pr := range prs {
+		for i, f := range pr.Functions {
```
brancz, line 92: same as mappings

pkg/convert/pprof.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func NewProfileToPprofConverter() *ProfileToPprofConverter {
+	return &ProfileToPprofConverter{
+		mappingCache: map[string]*profile.Mapping{},
+	}
+}
+
+// Convert converts several per process Profile to a pprof Profile.
+func (c *ProfileToPprofConverter) Convert(captureTime time.Time, periodNS int64, prs ...*profiler.Profile) (*profile.Profile, error) {
```
brancz, line 40: What's the idea with multiple profiles here? All usages I can spot in this PR only passes single profiles.

pkg/profiler/cpu/cpu.go
```diff
@@ -849,7 +851,7 @@ func (p *CPU) obtainProfiles(ctx context.Context) ([]*profiler.Profile, error) {
 				locationIndex, ok := locationIndices[id][addr]
 				if !ok {
 					locationIndex = len(locations[id])
```
brancz, line 853: All this tracking of location indexes is exactly about the deduplication and monotonic increasing IDs.

brancz:
I think we need to flip this whole thing around and instead abstract the building of the profile and replace that in the future for our own protocol instead of having an intermediate representation. We're inevitably loosing context or are doing unnecessary work with an intermediate representation, and as we can see here we are already mixing multiple approaches.

Sylfrena:
Thanks so much for the quick PR! 

I'll look at this in more detail tomorrow but it'd be great if you could explain the implementation approach you've used here in the PR description. 

At first glance, I agree with what @brancz said about the approach - I feel we will make extra work for us with an intermediate format. 

Given that we've run into this so many times, I think it makes sense to already add the smoke test in this PR already- if you're close to finishing that then we should include that too.

Issue #1696, opened 2023-05-27
libbpf: map 'percpu_stats': unexpected value size 960 provided for per-CPU map, expected 32 * 80 = 2560


kakkoyun:
```
level=debug name=parca-agent ts=2023-05-27T00:40:36.963327087Z caller=cpu.go:233 msg="libbpf: map 'percpu_stats': unexpected value size 960 provided for per-CPU map, expected 32 * 80 = 2560"
level=error name=parca-agent ts=2023-05-27T00:40:36.963385016Z caller=metrics.go:174 msg="readPerCpuCounter failed" error="get count values: failed to lookup value 0xc00166d330 in map percpu_stats: invalid argument"
```

cc @javierhonduco

javierhonduco:
Somehow only 12 CPUs (960/80) are detected. Could you share your `/proc/cpuinfo` and what the following Go code outputs in your machine `runtime.NumCPU()` as well as the output from `/sys/devices/system/cpu/{present,possible,online}`?

Perhaps unrelated, but in some AMD machines, the online /

kakkoyun:
> runtime.NumCPU()

Go number of CPUs: 12



> /proc/cpuinfo

<details><summary>Details</summary>
<p>

```
processor	: 0
vendor_id	: AuthenticAMD
cpu family	: 23
model		: 113
model name	: AMD Ryzen 5 3600 6-Core Processor
stepping	: 0
microcode	: 0x8701030
cpu MHz		: 3015.051
cache size	: 512 KB
physical id	: 0
siblings	: 12
core id		: 0
cpu cores	: 6
apicid		: 0
initial apicid	: 0
fpu		: yes
fpu_exception	: yes
cpuid level	: 16
wp		: yes
[...]

PR #1709, opened 2023-06-01
Use libbpf's cpu number helper

javierhonduco:
As this is causing issues in some AMD CPUs. For context: https://github.com/parca-dev/parca-agent/issues/1696#issuecomment-1568736923.

Note: This should never be used in perf_event_open(2) as it will fail in the CPUs that are overreported by some AMD systems.

Fixes https://github.com/parca-dev/parca-agent/issues/1696


kakkoyun (approved):
Awesome 🏆 🥇 💯

PR #1727, opened 2023-06-05
bpf: enable JIT unwinding by default

Sylfrena:
Set "--dwarf-unwinding-mixed" flag to true by default


Sylfrena:
cc @brancz

javierhonduco (approved):
LGTM -- can you test that --dwarf-unwinding-mixed=false works as expected? Not sure how `kong` deals with these flags

PR #1739, opened 2023-06-07
debuginfo/manager: Exclude the time to wait for goroutines turn

kakkoyun:


brancz (approved):
Oooooo is that why we saw upload context cancelled so frequently?

kakkoyun:
> Oooooo is that why we saw upload context cancelled so frequently?

I'm not sure 100% sure but this is semantically more correct.

Most of the context cancellation happens at `ShouldInititateUpload` calls, and this doesn't effect it. I still don't know the reason for that DeadlineExceeds.

brancz:
Ah right, that's what I observed as well.

PR #1743, opened 2023-06-08
pkg/process: Ensure debuginfo is uploaded

brancz:
### Why?
Previously we missed uploading debuginfo if it failed on the first attempt but gathering process info succeeded, which happens quite a lot in practice.

### What?

copilot:summary

### How?

copilot:walkthrough

### Test Plan

Tested locally.


javierhonduco (approved):
LGTM

kakkoyun (approved):

PR #1758, opened 2023-06-13
tests/integration: enable in CI

javierhonduco:
Enable integration tests in CI


kakkoyun (approved):

javierhonduco:
Forgot to mention why we are enabling them now! The integration tests used to be very flaky (50% of the runs will fail due to spurious issues in the metadata collection) but thanks to all the hard work by @kakkoyun and others this is way better now. I have run the integration tests in a loop and saw zero failures!

Let's pay attention to CI to see whether there's some flakiness here, though!

javierhonduco:
Even though https://github.com/parca-dev/parca-agent/issues/1141 is already closed, this is the very last bit that was needed 😄

PR #1759, opened 2023-06-13
*: Avoid constant recreation of goroutines

kakkoyun:
* Add WorkerPool to the hot path
* Prevent excessive allocation on the hot path.

### Why?
Previously, we were constantly re-creating goroutines for each event we received from the kernel.
This PR introduces a worker pool on the hot path to prevent constant routine creation.
It also avoids creating a new task if there is an in-flight one. 
As one last optimization, it uses a `sync.Pool` to avoid re-allocating the jobs (previously, we have seen many closure creations in the profiles).

Moreover, it uses a similar strategy to ensure we do not create an upload request if there's already something in-flight.
Even though previously we made sure there was only one upload request in-flight at a time, we were creating a lot of blocked goroutines.

### Test Plan
1. make ENABLE_RACE=yes test
2. Local on-host run
3. Local cluster run

### TODO
- [ ] Add profiles and screeshots

### DIFF
--- a/pkg/objectfile/object_file.go
+++ b/pkg/objectfile/object_file.go
@@ -123,7 +123,6 @@ func (o *ObjectFile) close() error {
 		return err
 	}
 	// Successfully closed the file.
-	o.elf = nil
 	o.closed = true
 	o.closedBy = callers()
 	o.p.metrics.closed.WithLabelValues(lvSuccess).Inc()
--- a/pkg/process/info.go
+++ b/pkg/process/info.go
@@ -128,6 +128,9 @@ type InfoManager struct {
 	mapManager       *MapManager
 	debuginfoManager DebuginfoManager
 	labelManager     LabelManager
+
+	uploadJobQueue chan *uploadJob
+	uploadJobPool  *sync.Pool
 }
 
 func NewInfoManager(
@@ -142,7 +145,7 @@ func NewInfoManager(
 	profilingDuration time.Duration,
 	cacheTTL time.Duration,
 ) *InfoManager {
-	return &InfoManager{
+	im := &InfoManager{
 		logger:  logger,
 		tracer:  tracer,
 		metrics: newMetrics(reg),
@@ -162,7 +165,15 @@ func NewInfoManager(
 		mapManager:       mm,
 		debuginfoManager: dim,
 		labelManager:     lm,
+
+		uploadJobQueue: make(chan *uploadJob, 128),
+		uploadJobPool: &sync.Pool{
+			New: func() interface{} {
+				return &uploadJob{}
+			},
+		},
 	}
+	return im
 }
 
 type Info struct {
@@ -200,7 +211,7 @@ func (im *InfoManager) fetch(ctx context.Context, pid int) (info Info, err error
 	// See the cache initialization for the eviction policy and the eviction TTL.
 	info, exists := im.cache.Peek(pid)
 	if exists {
-		im.ensureDebuginfoUploaded(ctx, pid, info.Mappings)
+		im.ensureDebuginfoUploaded(ctx, info.Mappings)
 		return info, nil
 	}
 
@@ -243,7 +254,7 @@ func (im *InfoManager) fetch(ctx context.Context, pid int) (info Info, err error
 	}
 
 	// Upload debug information of the discovered object files.
-	im.ensureDebuginfoUploaded(ctx, pid, mappings)
+	im.ensureDebuginfoUploaded(ctx, mappings)
 
 	// No matter what happens with the debug information, we should continue.
 	// And cache other process information.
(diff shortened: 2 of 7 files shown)


pkg/concurrency/pool.go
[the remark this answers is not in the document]
kakkoyun: I will add more comments. And explanations.

kakkoyun:
> > Previously, we were constantly re-creating goroutines for each event we received from the kernel.
> > This PR introduces a worker pool on the hot path to prevent constant routine creation.
> > It also ensures to avoid creation of a new task if there is an in-flight one.
> 
> Why do we need multiple Goroutines? If we need multiple, why do we need a dynamic worker pool rather than a static one?
> 
> I am thinking of something like this:
> 
> ```
>                                                                _n_ Goroutines as workers, created 
>                                                               early on and not destroyed until the
>                                                                          profiler exits           
>                                                                                                   
>                                                                                                   
>                                                                     ┌──────────────────────┐      
>                                                                     │   Static Goroutine   │      
>                                                             ┌──────▶│                      │      
>                                                             │       └──────────────────────┘      
> ┌─────────────────────────┐             Go channel          │       ┌──────────────────────┐      
> │                         │                                 │       │   Static Goroutine   │      
> │       Perf buffer       │                                 ├──────▶│                      │      
> │                         │─────────────────────────────────┤       └──────────────────────┘      
> │                         │                                 │       ┌──────────────────────┐      
> └─────────────────────────┘                                 │       │   Static Goroutine   │      
>                                                             ├──────▶│                      │      
>                                                             │       └──────────────────────┘      
>                                                             │       ┌──────────────────────┐      
>                                                             │       │   Static Goroutine   │      
>                                                             └──────▶│                      │      
>                                                                     └──────────────────────┘      
> ```

It's also static. Its dynamism comes from if a panic occurs. It ensures there are always `n` goroutines running. And they get destroyed when a profiler exists.

javierhonduco:
Why would a panic occur here? I am thinking that if that's not going to happen we can avoid the complexity coming from the refilling of the worker pool

kakkoyun:
> Why would a panic occur here? I am thinking that if that's not going to happen we can avoid the complexity coming from the refilling of the worker pool

I understand the reasoning. Under the fetch method, we make calls to various components of the system and as of now, we don't have a sound way to guarantee everything won't fail. This is a layer that ensures resiliency and makes sure everything is optimized on the hot path. 

If you strongly feel, this isn't worth the complexity. We can start with something simpler. I don't want to introduce something you won't feel comfortable maintaining.

kakkoyun:
> > Why would a panic occur here? I am thinking that if that's not going to happen we can avoid the complexity coming from the refilling of the worker pool
> 
> I understand the reasoning. Under the fetch method, we make calls to various components of the system and as of now, we don't have a sound way to guarantee everything won't fail. This is a layer that ensures resiliency and makes sure everything is optimized on the hot path.
> 
> If you strongly feel, this isn't worth the complexity. We can start with something simpler. I don't want to introduce something you won't feel comfortable maintaining.

An example error propagated:
```
panic: errgroup failed: failed to fetch process info for 537724: process not found
failed to open proc 537724: stat /proc/537724: no such file or directory
```

kakkoyun:
> In general, in the agent, we have some inherent complexity we can't avoid because it's at the core of what we do, but we have other areas where we can make some adjustments in the complexity level we want to take. I think we need to choose carefully which areas we want to use our complexity budget on

Definitely, I agree with this.

> Sounds like this is a transient error due to the races that can happen in procfs, we should maybe log it and move on. A panic sounds like too much for something that's not a hard assertion. What do you think?

panic is just for testing purposes. Of course, it will be handled differently.

javierhonduco:
> panic is just for testing purposes. Of course, it will be handled differently.

Ah, was going to ask because I couldn't find it in the code! In general, I think we should try to make our code as reliable as we can, but without considering too many "what ifs". We already try hard to never use `panic`s unless it's a very bad error that we can't/don't want to recover from.

pkg/objectfile/object_file.go
```diff
@@ -123,7 +123,6 @@ func (o *ObjectFile) close() error {
 		return err
 	}
 	// Successfully closed the file.
-	o.elf = nil
```
kakkoyun, line 126: This was pointed out by the race detector.

pkg/process/info.go
```diff
@@ -162,7 +165,15 @@ func NewInfoManager(
 		mapManager:       mm,
 		debuginfoManager: dim,
 		labelManager:     lm,
+
+		uploadJobQueue: make(chan *uploadJob, 128),
+		uploadJobPool: &sync.Pool{
```
brancz, line 170: Was this pool really necessary? Couldn't we have just used a channel that passes value structs?

brancz (commented):
I think you made the right call not to use an additional library!

I think we'll want to add some monitoring to understand if the workers can keep up with what's thrown at them, and then ultimately make the number of workers configurable.

pkg/process/info.go
brancz: happy to do that after we release, it's minor at this point

Issue #1741, opened 2023-06-07
debuginfoClient.ShouldInitiateUpload returns unexpected context.DeadlineExceeded


kakkoyun:
```
level=debug name=parca-agent ts=2023-06-07T23:08:22.545027679Z caller=info.go:294 component=process_info msg="upload mapping" err="failed to check whether build ID exists: rpc error: code = DeadlineExceeded desc = context deadline exceeded" buildid=8764bb88ac39cf61 filepath=/proc/8497/root/usr/lib/libc-2.32.so
```

PR #1765, opened 2023-06-14
cpu/process info: Reject inflight requests

kakkoyun:
Increase unary timeouts

ref #1741 

### Why?
After the changes, I haven't seen any `context.DeadlineExceeded` errors. And the number of goroutines are in-bounds.

Both of the changes in collaboration made the errors disappear. First of all, a second timeout for our unary requests was too low. Second, repeatedly creating the same demand for the same processes and creating a goroutine for each practically overwhelmed the server for the `ShouldInitiateUpload` RPC until we get a successful response and cache it.

### Test Plan
1. make ENABLE_RACE=yes test
2. Local on host run
3. Local cluster run


pkg/grpc/conn.go
```diff
@@ -90,7 +90,7 @@ func Conn(logger log.Logger, reg prometheus.Registerer, tp trace.TracerProvider,
 			grpc.MaxCallRecvMsgSize(parcadebuginfo.MaxMsgSize),
 		),
 		grpc.WithChainUnaryInterceptor(
-			timeout.UnaryClientInterceptor(time.Second),
```
brancz, line 93: 🤦‍♂️

PR #1777, opened 2023-06-19
Do not cache process info if we failed to read a mapping

kakkoyun:

And expire the process info cache after-write to make sure we refresh.


pkg/profiler/cpu/cpu.go
```diff
@@ -97,9 +97,7 @@ type CPU struct {
 	lastSuccessfulProfileStartedAt time.Time
 	lastProfileStartedAt           time.Time
 
-	debugProcessNames      []string
-	isNormalizationEnabled bool
```
kakkoyun, line 101: It wasn't used.

javierhonduco (approved):

brancz (approved):

PR #1778, opened 2023-06-19
Add flags to control perf event buffer event handling

kakkoyun:

- Add flags to control perf event buffer event handling
- Try to reduce the impact of idly waiting for goroutines.

Profile: https://pprof.me/c76ea10


brancz (approved):

kakkoyun:
> Nice one with reducing the case statements!

I tried a few things, but nothing improved more than what we already have. It is good enough to release now.
We can try improve performance on the hot path if it becomes bottle-neck.

brancz:
👍

PR #1795, opened 2023-06-25
pkg/convert: support convert collapsed async profiler to pprof

zdyj3170101136:
### Why?
support convert collapsed format of async profiler output to pprof

### What?
use a state machine to parse the file.


brancz:
In which situation would we use this instead of the JFR->pprof converter?

zdyj3170101136:
> In which situation would we use this instead of the JFR->pprof converter?

just to Provide additional options

brancz:
If we're not using this, I don't think we should have this code here.

brancz:
Agreed with @v-thakkar

Issue #1798, opened 2023-06-27
panic: file is already closed


sdirkwinkel:
**Describe the bug**
parca-agent crashes after a couple seconds/minutes

**To Reproduce**

**Logs**
```
panic: file is already closed
[...]

PR #1800, opened 2023-06-27
Increase default object file pool size and improve error

brancz:
It's possible on machines that create a lot of very short lived processes to evict an object file from the pool before it was possible to be used. In these cases the best solution is to increase the object pool size since not increasing it would lead to extremely high cache misses which would also be very bad.

### Why?

#1798 

### What?

Increase the default object file pool (128 seemed extraordinarily low) and add instructions to the panic output what to do in this case.

### Alternative

We could also return an error, however, it is likely that incorrect data would be produced, so I felt it is better to fail and be a bit more forcing to the user to update the value to something better that will work for their environment.

### Test Plan
n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ Flags:
                                    of memory that may be locked into RAM. It is
                                    used to ensure the agent can lock memory for
                                    eBPF maps. 0 means no limit.
-      --object-file-pool-size=128
+      --object-file-pool-size=512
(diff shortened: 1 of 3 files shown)


brancz:
Judging by the stacks reported by the user it was eviction that explicitly closed the file, while other parts of the program still had a pointer to the object file struct. This causes a race between the eviction and the object file struct being used.

PR #1803, opened 2023-06-28
Add analytics

brancz:
### Why?

We need to know what kind of work to prioritize, and that works best by having (anonymous) data of usage.

### What?

This patch adds anonymous usage data starting at 10 minutes of a parca-agent being up, that sends data to `analytics.parca.dev`. All Parca maintainers have access to this data.

The initial set of data includes:
* `architecture` to help us understand which architectures we should be working on
* `version` which versions are out there, which helps us understand when to deprecate/remote APIs from Parca
* `cpu_cores` which helps us prioritize work for either machines with high CPU count or low CPU count
* Kernel infos `kernel_version` and`kernel_osrelease` to help us understand which kernel versions its actively used on (maintaining old kernel support is tedious)
* OS infos `os_name`, `os_vendor", `os_version`, `os_release` to help us understand on which distros we should be testing Parca Agent
* `is_container` to understand whether or not parca agent is most popularly deployed as a container

### How?

Using Prometheus remote write, we send the anonymized data to a Prometheus server.

### Test Plan

Tested locally and will test on Polar Signals production once we have a container image.

### DIFF
--- a/cmd/parca-agent/main.go
+++ b/cmd/parca-agent/main.go
@@ -43,6 +43,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/prometheus/procfs"
 	"github.com/prometheus/prometheus/promql/parser"
+	"github.com/zcalusic/sysinfo"
 	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/automaxprocs/maxprocs"
 	"google.golang.org/grpc"
@@ -52,9 +53,11 @@ import (
 	_ "google.golang.org/grpc/encoding/proto"
 
 	"github.com/parca-dev/parca-agent/pkg/agent"
+	"github.com/parca-dev/parca-agent/pkg/analytics"
 	"github.com/parca-dev/parca-agent/pkg/buildinfo"
 	"github.com/parca-dev/parca-agent/pkg/byteorder"
 	"github.com/parca-dev/parca-agent/pkg/config"
+	"github.com/parca-dev/parca-agent/pkg/cpuinfo"
 	"github.com/parca-dev/parca-agent/pkg/debuginfo"
 	"github.com/parca-dev/parca-agent/pkg/discovery"
 	parcagrpc "github.com/parca-dev/parca-agent/pkg/grpc"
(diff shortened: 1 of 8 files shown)


pkg/analytics/analytics.go
```diff
@@ -0,0 +1,119 @@
...
+	version string,
+) *AnalyticsSender {
+	return &AnalyticsSender{
+		logger:    logger,
+		client:    client,
+		machineID: randSeq(20),
+		arch:      arch,
+		cpuCores:  float64(cpuCores),
+		version:   version,
+	}
```
javierhonduco, line 65: Let's add the kernel release and linux distro, too!

pkg/analytics/analytics.go
```diff
@@ -0,0 +1,119 @@
...
+	logger log.Logger,
+	client *Client,
+	arch string,
+	cpuCores int,
+	version string,
+) *AnalyticsSender {
+	return &AnalyticsSender{
+		logger:    logger,
+		client:    client,
+		machineID: randSeq(20),
```
javierhonduco, line 61: A new machine id is generated every single time the agent boots up, so we can't tell how many unique machines are running the Agent nor what machine is sending what data, which I assumed was the goal.

Perhaps we could submit the hash of the hostname?

cmd/parca-agent/main.go
```diff
@@ -479,6 +483,29 @@ func run(logger log.Logger, reg *prometheus.Registry, flags flags) error {
...
+			runtime.GOARCH,
+			cpuinfo.NumCPU(),
+			version,
+		)
+		ctx, cancel := context.WithCancel(ctx)
+		g.Add(func() error {
+			a.Run(ctx)
+			return nil
+		}, func(error) {
+			cancel()
```
javierhonduco, line 505: Could you add a log if there's an error here?

pkg/analytics/remote_write.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func (c *Client) Send(ctx context.Context, wreq *prometheus.WriteRequest) error {
+	c.pBuf.Reset()
+	err := c.pBuf.Marshal(wreq)
+	if err != nil {
+		return err
+	}
+
+	// snappy uses len() to see if it needs to allocate a new slice. Make the
```
javierhonduco, line 78: Had no idea, thanks for the comment!

javierhonduco (approved):
Great stuff, this will be instrumental in helping us debug and understand the Agent in the wild

pkg/analytics/analytics.go
brancz: good idea!

PR #1810, opened 2023-06-30
use error.As()

Sylfrena:
### <samp>🤖 Generated by Copilot at 6c9de4f</samp>

> _We're sailing on the Go-lang ship, with errors in our code_
> _We need to check them properly, or else we'll sink our load_
> _So grab your `errors.As` and heave away with me_
> _We'll refactor all the `elf.FormatError`s on the count of three_

-->

### DIFF
--- a/pkg/objectfile/pool.go
+++ b/pkg/objectfile/pool.go
@@ -234,7 +234,8 @@ func (p *Pool) NewFile(f *os.File) (_ *ObjectFile, err error) { //nolint:nonamed
 	// > Clients of ReadAt can execute parallel ReadAt calls on the same input source.
 	ef, err := elfNewFile(f)
 	if err != nil {
-		if errors.Is(err, &elf.FormatError{}) {
+		var elfErr *elf.FormatError
+		if errors.As(err, &elfErr) {
 			p.metrics.openErrors.WithLabelValues(lvNotELF).Inc()
 		} else {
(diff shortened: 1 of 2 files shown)


brancz (approved):

maxbrunet (commented):
That's a good sea shanty :boat::pirate_flag:

pkg/objectfile/pool.go
```diff
@@ -234,7 +234,8 @@ func (p *Pool) NewFile(f *os.File) (_ *ObjectFile, err error) { //nolint:nonamed
 	// > Clients of ReadAt can execute parallel ReadAt calls on the same input source.
 	ef, err := elfNewFile(f)
 	if err != nil {
-		if errors.Is(err, &elf.FormatError{}) {
+		var elfErr *elf.FormatError
+		if errors.As(err, &elfErr) {
```
maxbrunet, line 238: Curious, what's the benefit of `error.As()` if `elfErr` is never used?

maxbrunet (commented):
That's a good sea shanty :boat::pirate_flag:

pkg/objectfile/pool.go
brancz: Unfortunately errors.Is only works to check identical copies. Just doing errors.Is with an empty elf.FormatError is not equal to one that has the (unexported) fields set.

maxbrunet: Ah thank you, makes sense!

PR #1812, opened 2023-06-30
Fix JITDump symbolization edge-case

brancz:
### Why?

We noticed Julia was not being symbolized correctly.

### What?

Julia behaves a little unexpectedly with its jitdump files. Instead of setting them in the respective mapping file, the jit sections have no files and there is a single jitdump file referenced in another mapping section.

### How?

Iterate over all mappings and attempt to symbolize using all known jitdump files whenever encountering a mapping that has no file.

### Test Plan

Tested locally with Julia 1.9.1

This was co-authored by @Sylfrena, huge thanks to her for a very long debugging session.


vchuravy:
This kinda sounds like an upstream LLVM bug...

brancz:
We tried to find some sort of standard or specification, but it turns out there is none, so while perhaps inconsistent with for example nodejs, there is nothing that says the behavior is incorrect.

vchuravy:
Can you expand on the behavior you observed? Since this may break other consumers as well. I thought we only ever should emit one jitdump file.

PR #1835, opened 2023-07-06
Do not use global tracer provider

kakkoyun:

- Use passed down trace provider rather than implicitly depending on the global one.
- Make sure the analytics component has proper context propagation.


javierhonduco:
Why is this changed needed? How does it change the traces?

kakkoyun:
> Why is this changed needed? How does it change the traces?

This is to make code more predictable. Previously, it was grabbing the tracer provider implicitly from the global state.

Moreover, this makes sure we have context propagation for the analytic layer. These are minor things I've realized working things on the server side.

PR #1836, opened 2023-07-06
Use frame pointer unwinding for v8-based application

javierhonduco:
We can't unwind nodejs and other v8-base applications as we can't continue unwinding after the first AoT frame:

```
[...]
	current pc: 7fb0cf84904a
	current sp: 7fff93701498
	current bp: 7fff937014d8
[debug] Unwinding JITed stacks
[debug] Within unwinding machinery loop
	current pc: 157a329
	current sp: 7fff937014e8
	current bp: 7fff93701548
~about to check shards found=1
~checking shards now
[info] found chunk
le offset: 0
========== left 0 right 249997
	.done
	=> table_index: 206094
	=> adjusted pc: 157a329
	cfa type: 4, offset: 0 (row pc: 156f656)
[info] PC 157a329 not contained in the unwind info, found marker
[warn] mapping not added yet rbp 7fff93701548
pid 227936 tgid 227936
```

Turns out this AoT function doesn't have the corresponding DWARF unwind information:

```
$ dist/eh-frame --executable /home/javierhonduco/.nvm/versions/node/v16.13.0/bin/node | grep 157a3
$
```

After cross-validating the output with `bpftrace`'s and checking v8's source code it was clear why this is happening:

```
157a32a Builtins_InterpreterEntryTrampoline+202 (/home/javierhonduco/.nvm/versions/node/v16.13.0/bin/node)
157a32a Builtins_InterpreterEntryTrampoline+202 (/home/javierhonduco/.nvm/versions/node/v16.13.0/bin/node)
157a32a Builtins_InterpreterEntryTrampoline+202 (/home/javierhonduco/.nvm/versions/node/v16.13.0/bin/node)
1578538 Builtins_JSEntryTrampoline+88 (/home/javierhonduco/.nvm/versions/node/v16.13.0/bin/node)
15782c3 Builtins_JSEntry+131 (/home/javierhonduco/.nvm/versions/node/v16.13.0/bin/node)
[...]
7f65c5e29510 __libc_start_call_main+128 (/usr/lib64/libc.so.6)
```

The `Builtins_` functions above seem to be emmited via a custom code generator[0]. No special linker `.cfi` directives are used, so this is effectively like writting raw assembly with frame pointers set up, but without telling the rest of the tooling how to produce unwidn information it.

Test Plan
=========

- node: https://pprof.me/95f85d0
- basic-cpp-no-fp: https://pprof.me/a6099f2
- ruby: https://pprof.me/4ecf047

- [0]: https://github.com/v8/v8/blob/e554f1e97e30f6e4cf6eef2c5049c17ca708e103/src/builtins/x64/builtins-x64.cc

### DIFF
--- a/bpf/cpu/cpu.bpf.c
+++ b/bpf/cpu/cpu.bpf.c
@@ -565,32 +565,35 @@ static __always_inline bool retrieve_task_registers(u64 *ip, u64 *sp, u64 *bp) {
 static __always_inline bool has_fp(u64 current_fp) {
   u64 next_fp;
   u64 ra;
+  int i;
 
-  for (int i = 0; i < MAX_STACK_DEPTH; i++) {
+  for (i = 0; i < MAX_STACK_DEPTH; i++) {
     int err = bpf_probe_read_user(&next_fp, 8, (void *)current_fp);
     bpf_probe_read_user(&ra, 8, (void *)current_fp + 8);
     if (err < 0) {
-      // LOG("[debug] fp read failed with %d", err);
-      return false;
-    }
-    // Some cpp binaries, such as testdata/out/basic-cpp
-    // seem to have rbp set to 1 in the bottom frame. This
-    // does not comply with the x86_64 ABI.
-    //
-    // Additionally, we consider that stacks with just one
-    // frame aren't valid. This is just a heuristic, as most
-    // processes should at least have two frames.
-    //
-    // For both cases above, we prefer to unwind using the
-    // DWARF-derived unwind information.
-    if (next_fp == 0) {
-      // LOG("[debug] fp success");
-      return i > 0;
+      // LOG("[debug] fp read failed with %d i %d", err, i);
+      // We might have reached the bottom frame.
+      break;
     }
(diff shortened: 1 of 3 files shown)


brancz (approved):
Very nice find!

bpf/cpu/cpu.bpf.c
```diff
@@ -1097,6 +1104,14 @@ int profile_cpu(struct bpf_perf_event_data *ctx) {
...
 
+  // 2. We did not have unwind information, let's see if we can unwind with frame
+  // pointers.
+  if (has_fp(unwind_state->bp)) {
+    add_stack(ctx, pid_tgid, STACK_WALKING_METHOD_FP, NULL);
+    return 0;
+  }
+
+  // 3. Request unwind information.
   request_unwind_information(ctx, user_pid);
```
Sylfrena, line 1115: Thanks for all the numbered comments here!

javierhonduco: :D hopefully things will be clearer and this will also prevent accidental reorders!

PR #1864, opened 2023-07-18
pkg/cpu: Reduce .text section error spam

### LINKED ISSUE #1857: Could not find `.text` section spams error log
My logs are being spammed with the error message:

```
level=error name=parca-agent ts=2023-07-13T16:16:20.242517888Z caller=cpu.go:308 component=cpu_profiler msg="failed to add unwind table" pid=1890187 err="setUnwindTableForMapping for
 executable /root/.julia/artifacts/37805a06093bdc0d5938601719cf02fa0cf3aa68/lib/rr/librrpage.so starting at 0x6fffd000 failed: BuildID failed /proc/1890187/root/root/.julia/artifacts/37805a06093bdc0d5938601719cf02fa0cf3aa68/lib/rr/librrpage.so: could not find .text section"
```

It would be better if the error would only be reported once.

Sylfrena:
Fixes #1857

### Why?
User report that logs are being spammed with 

```
level=error name=parca-agent ts=2023-07-13T16:16:20.242517888Z caller=cpu.go:308 component=cpu_profiler msg="failed to add unwind table" pid=1890187 err="setUnwindTableForMapping for executable /root/.julia/artifacts/37805a06093bdc0d5938601719cf02fa0cf3aa68/lib/rr/librrpage.so starting at 0x6fffd000 failed: BuildID failed /proc/1890187/root/root/.julia/artifacts/37805a06093bdc0d5938601719cf02fa0cf3aa68/lib/rr/librrpage.so: could not find .text section"
```

in #1857

### What?
Add a counter to keep track of said error and only logs it once every 50 times we hit the error case to reduce spamming the logs in default logging mode.
 

### <samp>🤖 Generated by Copilot at 502844c</samp>

> _We search for the text in the maps of the dead_
> _But sometimes we find nothing but `ErrTextSectionNotFound`_
> _We track the errors and we log the doom_
> _We need the unwind tables to escape the CPU's tomb_

-->

### DIFF
--- a/pkg/buildid/elf.go
+++ b/pkg/buildid/elf.go
@@ -26,6 +26,8 @@ import (
 
 const goBuildIDSectionName = ".note.go.buildid"
 
+var ErrTextSectionNotFound = errors.New("could not find .text section")
+
 // FromELF returns the build ID for an ELF binary.
 func FromELF(ef *elf.File) (string, error) {
 	// First, try fast methods.
@@ -63,7 +65,7 @@ func buildid(ef *elf.File) (string, error) {
 	// If we didn't find a GNU build ID, try hashing the .text section.
 	text := ef.Section(".text")
 	if text == nil {
-		return "", errors.New("could not find .text section")
+		return "", ErrTextSectionNotFound
 	}
 	h := xxhash.New()
 	if _, err := io.Copy(h, text.Open()); err != nil {
--- a/pkg/profiler/cpu/cpu.go
+++ b/pkg/profiler/cpu/cpu.go
@@ -41,7 +41,9 @@ import (
 	"github.com/puzpuzpuz/xsync/v2"
 	"golang.org/x/sys/unix"
 
+	"github.com/parca-dev/parca-agent/pkg/buildid"
 	"github.com/parca-dev/parca-agent/pkg/byteorder"
+	"github.com/parca-dev/parca-agent/pkg/cache"
 	"github.com/parca-dev/parca-agent/pkg/cpuinfo"
 	"github.com/parca-dev/parca-agent/pkg/metadata/labels"
 	"github.com/parca-dev/parca-agent/pkg/pprof"
@@ -100,6 +102,7 @@ type CPU struct {
 
 	lastError                      error
 	processLastErrors              map[int]error
+	processErrorTracker            *cache.LRUCache[string, int]
 	lastSuccessfulProfileStartedAt time.Time
 	lastProfileStartedAt           time.Time
 
@@ -157,6 +160,9 @@ func NewCPUProfiler(
 		metrics:   newMetrics(reg),
 
 		memlockRlimit: memlockRlimit,
+
+		// increase cache length if needed to track more errors
+		processErrorTracker: cache.NewLRUCache[string, int](prometheus.WrapRegistererWith(prometheus.Labels{"cache": "no_text_section_error_tracker"}, reg), 512),
 
 		debugProcessNames: debugProcessNames,
 
@@ -304,6 +310,19 @@ func (p *CPU) addUnwindTableForProcess(pid int) {
 			level.Debug(p.logger).Log("msg", "failed to add unwind table due to a procfs race", "pid", pid, "err", err)
 		} else if errors.Is(err, errTooManyExecutableMappings) {
 			level.Warn(p.logger).Log("msg", "failed to add unwind table due to having too many executable mappings", "pid", pid, "err", err)
+		} else if errors.Is(err, buildid.ErrTextSectionNotFound) {
+			v, ok := p.processErrorTracker.Peek(err.Error())
+			if ok {
+				p.processErrorTracker.Add(err.Error(), v+1)
+			} else {
+				p.processErrorTracker.Add(err.Error(), 1)
+			}
+			v, _ = p.processErrorTracker.Get(err.Error())
+			if v%50 == 0 || v == 1 {
+				level.Error(p.logger).Log("msg", "failed to add unwind table due to unavailable .text section", "pid", pid, "err", err, "encounters", v)
+			} else {
+				level.Debug(p.logger).Log("msg", "failed to add unwind table due to unavailable .text section", "pid", pid, "err", err, "encounters", v)
+			}
 		} else {
 			level.Error(p.logger).Log("msg", "failed to add unwind table", "pid", pid, "err", err)
 		}


pkg/profiler/cpu/cpu.go
```diff
@@ -304,6 +308,13 @@ func (p *CPU) addUnwindTableForProcess(pid int) {
 			level.Debug(p.logger).Log("msg", "failed to add unwind table due to a procfs race", "pid", pid, "err", err)
 		} else if errors.Is(err, errTooManyExecutableMappings) {
 			level.Warn(p.logger).Log("msg", "failed to add unwind table due to having too many executable mappings", "pid", pid, "err", err)
+		} else if errors.Is(err, ErrTextSectionNotFound) {
+			if p.processErrorTracker[err]%50 == 0 {
+				level.Error(p.logger).Log("msg", "failed to add unwind table due to unavailable .text section", "pid", pid, "err", err)
+			} else {
+				level.Debug(p.logger).Log("msg", "failed to add unwind table due to unavailable .text section", "pid", pid, "err", err)
+			}
+			p.processErrorTracker[err] = p.processErrorTracker[err] + 1
```
brancz, line 317: shorten to `p.processErrorTracker[err]++`

pkg/profiler/cpu/cpu.go
```diff
@@ -304,6 +308,13 @@ func (p *CPU) addUnwindTableForProcess(pid int) {
 			level.Debug(p.logger).Log("msg", "failed to add unwind table due to a procfs race", "pid", pid, "err", err)
 		} else if errors.Is(err, errTooManyExecutableMappings) {
 			level.Warn(p.logger).Log("msg", "failed to add unwind table due to having too many executable mappings", "pid", pid, "err", err)
+		} else if errors.Is(err, ErrTextSectionNotFound) {
+			if p.processErrorTracker[err]%50 == 0 {
```
brancz, line 312: let's make this 50 at least a configurable field on the `CPU` struct, so we can potentially make it a flag if we wanted to one day

pkg/profiler/cpu/cpu.go
```diff
@@ -280,6 +283,7 @@ func loadBpfProgram(logger log.Logger, reg prometheus.Registerer, mixedUnwinding
 
 func (p *CPU) addUnwindTableForProcess(pid int) {
 	executable := fmt.Sprintf("/proc/%d/exe", pid)
+	//p.processErrorTracker = map[error]int{}
```
brancz, line 286: leftover?

pkg/profiler/cpu/maps.go
```diff
@@ -151,6 +151,7 @@ var (
 	errUnrecoverable             = errors.New("unrecoverable error")
 	errTooManyExecutableMappings = errors.New("too many executable mappings")
 	ErrNeedMoreProfilingRounds   = errors.New("not enough profiling rounds with this unwind info")
+	ErrTextSectionNotFound       = errors.New("could not find .text section")
```
brancz, line 154: We don't actually return this error anywhere so the `errors.Is` above won't work.

pkg/profiler/cpu/cpu.go
```diff
@@ -158,6 +159,8 @@ func NewCPUProfiler(
 
 		memlockRlimit: memlockRlimit,
 
+		processErrorTracker: map[error]int{},
```
brancz, line 162: the `error` type can behave a bit weird here, I think we should just use `string` and use `err.Error()`

brancz, line 162: I'm a bit worried about this being unbounded, let's use a capped cache implementation instead.

brancz (commented):
Nice yes this is exactly the type of direction I was thinking of.

pkg/profiler/cpu/cpu.go
```diff
@@ -304,6 +308,13 @@ func (p *CPU) addUnwindTableForProcess(pid int) {
 			level.Debug(p.logger).Log("msg", "failed to add unwind table due to a procfs race", "pid", pid, "err", err)
 		} else if errors.Is(err, errTooManyExecutableMappings) {
 			level.Warn(p.logger).Log("msg", "failed to add unwind table due to having too many executable mappings", "pid", pid, "err", err)
+		} else if errors.Is(err, ErrTextSectionNotFound) {
+			if p.processErrorTracker[err]%50 == 0 {
+				level.Error(p.logger).Log("msg", "failed to add unwind table due to unavailable .text section", "pid", pid, "err", err)
```
kakkoyun, line 313: nit: It would be nice to see the number of encounters so far in the logs.
```suggestion
				level.Error(p.logger).Log("msg", "failed to add unwind table due to unavailable .text section", "pid", pid, "err", err, "encounters", p.processErrorTracker[err])
```

pkg/profiler/cpu/maps.go
kakkoyun: And if this doesn't cross the package boundary, maybe we should consider not exporting it?

pkg/profiler/cpu/cpu.go
brancz: I like this

pkg/profiler/cpu/maps.go
Sylfrena: agreed! 
it was a typo on my end, but thanks so much for catching it!

pkg/profiler/cpu/maps.go
```diff
@@ -959,6 +960,9 @@ func (m *bpfMaps) setUnwindTableForMapping(buf *profiler.EfficientBuffer, pid in
 	}
 	buildID, err := buildid.FromELF(ef)
 	if err != nil {
+		if errors.Is(err, errTextSectionNotFound) {
+			return fmt.Errorf("BuildID failed %s: %w", fullExecutablePath, errTextSectionNotFound)
```
brancz, line 964: we should just use `err` in the last argument here

pkg/profiler/cpu/maps.go
```diff
@@ -959,6 +960,9 @@ func (m *bpfMaps) setUnwindTableForMapping(buf *profiler.EfficientBuffer, pid in
 	}
 	buildID, err := buildid.FromELF(ef)
 	if err != nil {
+		if errors.Is(err, errTextSectionNotFound) {
```
brancz, line 963: `buildid.FromELF` never actually return the `errTextSectionNotFound` so we will never go into this if. We need to move the error to the buildid package and return it here: https://github.com/parca-dev/parca-agent/blob/cefb78dc9c5582a57b2bbe62dd9fae3a0b20033b/pkg/buildid/elf.go#L66

Then we can check whether we're getting that error in this if with `if errors.Is(err, buildid.ErrTextSectionNotFound)`

pkg/profiler/cpu/cpu.go
```diff
@@ -158,6 +161,9 @@ func NewCPUProfiler(
 
 		memlockRlimit: memlockRlimit,
 
+		// increase cache length if needed to track more errors
+		processErrorTracker: cache.NewLRUCache[string, int](prometheus.WrapRegistererWith(prometheus.Labels{"cache": "error_tracker"}, reg), 1),
```
brancz, line 165: 1 entry is not going to be enough, since this can happen for X amount of processes, and even N amount of mappings within those processes

brancz: Also let's choose a better name `error_tracker` is a bit too generic, maybe `no_text_section_error_tracker`?

pkg/profiler/cpu/cpu.go
```diff
@@ -304,6 +310,19 @@ func (p *CPU) addUnwindTableForProcess(pid int) {
 			level.Debug(p.logger).Log("msg", "failed to add unwind table due to a procfs race", "pid", pid, "err", err)
 		} else if errors.Is(err, errTooManyExecutableMappings) {
 			level.Warn(p.logger).Log("msg", "failed to add unwind table due to having too many executable mappings", "pid", pid, "err", err)
+		} else if errors.Is(err, buildid.ErrTextSectionNotFound) {
+			v, ok := p.processErrorTracker.Peek(err.Error())
```
brancz, line 314: What's the full error string here again? We should make sure it contains the PID and/or mapping file path.

Sylfrena: I think we should be good here already. The [reported error string](https://github.com/parca-dev/parca-agent/issues/1857#issue-1803351339)(as seen below) already contains pid and mapping start address and mapping file path.
> `level=error name=parca-agent ts=2023-07-13T16:16:20.242517888Z caller=cpu.go:308 component=cpu_profiler msg="failed to add unwind table" pid=1890187 err="setUnwindTableForMapping for
 executable /root/.julia/artifacts/37805a06093bdc0d5938601719cf02fa0cf3aa68/lib/rr/librrpage.so starting at 0x6fffd000 failed: BuildID failed /proc/1890187/root/root/.julia/artifacts/37805a06093bdc0d5938601719cf02fa0cf3aa68/lib/rr/librrpage.so: could not find .text section"`

pkg/profiler/cpu/cpu.go
Sylfrena: Cool I was thinking maybe we could eventually reuse the cache to log different errors too but no strong opinions :D

pkg/buildid/elf.go
```diff
@@ -26,6 +26,10 @@ import (
 
 const goBuildIDSectionName = ".note.go.buildid"
 
+var (
+	ErrTextSectionNotFound = errors.New("could not find .text section")
+)
```
Sylfrena, line 31:
```suggestion
var ErrTextSectionNotFound = errors.New("could not find .text section") 
```

brancz (approved):

PR #1866, opened 2023-07-18
deploy: Remove liveness and readiness probes

brancz:
Parca-Agent no longer listens on all interfaces by default, therefore this doesn't work anymore.

### DIFF
--- a/deploy/lib/parca-agent/parca-agent.libsonnet
+++ b/deploy/lib/parca-agent/parca-agent.libsonnet
@@ -204,12 +204,6 @@ function(params) {
           containerPort: pa.config.port,
         },
       ],
-      livenessProbe: {
-        httpGet: {
-          path: '/healthy',
-          port: 'http',
-        },
-      },
       readinessProbe: {
         httpGet: {
           path: '/ready',


maxbrunet (changes requested):
Please add `httpGet.host: 127.0.0.1` instead

brancz:
Does not work, I tried.

brancz:
If you can get it to work on a Kubernetes cluster I'm perfectly fine with keeping them but I tried that configuration and it didn't work so I think removing is the best option. Besides, parca-agent doesn't serve traffic so I find the HTTP based probes kind of silly to begin with.

brancz:
I think this explains quite well why: https://stackoverflow.com/questions/62615675/kubernetes-redinessprobe-httpget-on-tomcat-image-reports-connection-refused

The `host` is the what the kubelet uses to perform the request, but the kubelet is in a different net namespace than the container, therefore the container's 127.0.0.1 is not the same interface as the kubelet's 127.0.0.1. The pod IP (which is the default) worked before because that was uniquely identifying.

brancz:
Let me know if you have any other concerns!

maxbrunet:
Also listening on localhost, does it mean we do not want observability endpoints (`/metrics` and `/debug/pprof`) accessible by default?

maxbrunet:
I think we would need make them accessible in the demo again at least, I believe we were using them there

maxbrunet (approved):
> kubelet is in a different net namespace than the container

I had totally forgotten what performs the checks

maxbrunet:
And I guess it is not possible to port-forward with this config either. For a systemd service, it'd be fine, but maybe it should continue to listen on all interfaces in the Kubernetes deployment by default? We could add a NetworkPolicy blocking by default if it's a concern?

brancz:
I think I'd be ok with explicitly setting `--listen-address=:7071` as well if you prefer that?

brancz:
Ok so false alarm, it was just Polar Signals' manifest that was not working, as we didn't specify `--listen-address=:7070` explicitly, but actually these upstream manifests do.

brancz:
I'd still like to remove liveness probes as I think they're generally more harm than good, but I'll close this PR here and open a new one.

PR #1867, opened 2023-07-18
pkg/stack/unwind: Ignore no symbols present for nodejs check

brancz:
If a binary has no symbols then this caused a bunch of errors potentially not generating unwind tables.


Sylfrena:
@javierhonduco

Sylfrena (approved):

PR #1874, opened 2023-07-20
pkg/profiler/cpu: Fix noop due to value struct

brancz:
This caused any process to only report on a single thread per profile cycle, which resulted in incorrect profiling data.

### Test

Main (only single thread stack traces per profile) :

CleanShot_2023-07-20_at_09 55 21

CleanShot_2023-07-20_at_09 55 44

Patch (multiple thread stack traces per profile):

CleanShot_2023-07-20_at_09 57 42
CleanShot_2023-07-20_at_09 57 29

### DIFF
--- a/pkg/profiler/cpu/cpu.go
+++ b/pkg/profiler/cpu/cpu.go
@@ -629,7 +629,11 @@ func (p *CPU) Run(ctx context.Context) error {
 				}
 				continue
 			}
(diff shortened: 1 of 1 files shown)


kakkoyun (approved):
LGTM 👍 

This is my bad. Thanks for catching it.

PR #1875, opened 2023-07-20
Maintainers: Remove Vaishali from the maintainers list

v-thakkar:
Today is my last day at Polar Signals. As I don't plan to be active in the `parca-agent` project after that, remove myself as a maintainer. :)


brancz:
Thank you for your service! You'll be missed!

PR #1876, opened 2023-07-20
deploy: Remove liveness probe

brancz:
Liveness probes are dangerous: https://srcco.de/posts/kubernetes-liveness-probes-are-dangerous.html


maxbrunet (approved):
LGTM

PR #1886, opened 2023-07-27
pkg/convert: reduce allocation

zdyj3170101136:
### Why?
in my company, the jfr to pprof allocate too much times of objects, and those cost most of cpu.
this is the allocation memory profile of my parca-agent:
image

### What?
this pr reduce 99% percent of memory allocation and reduce 80% of cpu cost in jfr to pprof.
image

### How?
1, cache parseArgs and getFileName and getPid and dimensions.
2, reuse slice as more as we can.
3, use bytes.buffer to build function key, and use unsafe string in map search to avoid memory allocation.
4, use unsafe string sample key in map search, only create locations when new sample created.
5, use uint64 as location key to avoid int to string convert.

### DIFF
--- a/pkg/convert/jfr.go
+++ b/pkg/convert/jfr.go
@@ -294,7 +294,7 @@ func (b *builder) parseArgs(s string) string {
 			}
 			s = s[i:]
 		}
-		result = "(" + strings.Join(b.args, ",") + ")"
+		result = "(" + strings.Join(b.args, ", ") + ")"
 		b.argCache[k] = result
 	}
 	return result
--- a/pkg/convert/jfr.go
+++ b/pkg/convert/jfr.go
@@ -104,11 +104,11 @@ type label struct {
 	value string
 }
 
-func (b *builder) getPid(JavaThreadID int64) string {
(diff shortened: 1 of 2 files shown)


pkg/convert/jfr.go
```diff
@@ -81,51 +99,87 @@ func increaseSample(s *profile.Sample) {
...
-		locKey, loc := b.getOrCreateLocation(fun, f.LineNumber)
-		locations = append(locations, loc)
-		locationKeys = append(locationKeys, locKey)
+		loc := b.getOrCreateLocation(fun, frame.LineNumber)
+		b.locationIDBuf = binary.AppendUvarint(b.locationIDBuf, loc.ID)
 	}
+	for _, l := range labels {
+		b.locationKeys = append(b.locationKeys, l.value)
+	}
+	JavaThreadID := b.getPid(thread.JavaThreadID)
```
brancz, line 134: lowercase `J`

brancz (commented):
Very nice!

brancz (approved):

brancz:
Nice work!!

PR #1897, opened 2023-08-01
pkg/convert: bump github.com/pyroscope-io/jfr-parser from v0.6.0 to  …

zdyj3170101136:
…github.com/zdyj3170101136/jfr-parser v0.0.0-20230801105541-d66951dc9605

### Why?
package github.com/pyroscope-io/jfr-parser malloc too much of memory
截屏2023-08-01 下午7 47 30

### What?
i reduce 99% percent of memory allocate and  67 % cpu cost in read jfr file:
https://github.com/grafana/jfr-parser/pull/22

### How?
1, do not store data as interface, let memory not leak.
2, stream read event, those cache common event type.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -185,3 +185,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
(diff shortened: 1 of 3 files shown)


brancz:
Very nice! Let's wait for the upstream PR to be merged, otherwise lgtm.

PR #1901, opened 2023-08-02
pkg/runtime: Prevent allocations from searching symbol tables

brancz:
### Why?
Before this patch searching symbol tables allocated a lot of memory in one allocation which caused heap to spike dramatically at the start.

### What? / How?

Same strategy as in https://github.com/xyproto/ainur/pull/5

Remove big bulk allocation. 

### Test Plan

Haven't tested other than compiling.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/parca-dev/parca-agent
 go 1.20
 
 require (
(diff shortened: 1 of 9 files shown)


javierhonduco (approved):
LGTM

kakkoyun (approved):

PR #1928, opened 2023-08-15
*: Don't perform address normalization locally

brancz:
Instead send the necessary information to perform normalization to the server and let the server do it. This also allows the server to potentially retrieve other sources for normalization calculations.

Client side to https://github.com/parca-dev/parca/pull/3625


kakkoyun (approved):
Awesome 🏆 🥇 💯

pkg/runtime/lua.go
```diff
@@ -0,0 +1,38 @@
...
+	"errors"
+	"fmt"
+)
+
+var luaJitIdentifyingSymbols = [][]byte{
+	[]byte("luaJIT"),
+}
+
+// HACK: This is a somewhat a brittle check.
+func IsLuaJit(f *elf.File) (bool, error) {
```
kakkoyun, line 27: It doesn't seem to be related to PR. Maybe a leftover?

pkg/process/maps.go
```diff
@@ -318,65 +315,81 @@ func (m *Mapping) findProgramHeader(ef *elf.File, addr uint64) (*elf.ProgHeader,
...
+		return nil, fmt.Errorf("specified address %x is outside the mapping range [%x, %x]", addr, m.StartAddr, m.EndAddr)
 	}
 
 	loadSegment, err := m.findProgramHeader(ef, addr)
 	if err != nil {
-		return 0, fmt.Errorf("failed to find program header for mapping %#v: %w", m, err)
+		return nil, fmt.Errorf("failed to find program header for mapping %#v: %w", m, err)
 	}
 
-	base, err := elfreader.Base(
```
kakkoyun, line 334: Since we move this method to upstream Parca, shall we clean it up? Is it used anywhere else?

pkg/runtime/lua.go
brancz: whops, yes, thanks

pkg/process/maps.go
brancz: no you're right!

PR #1933, opened 2023-08-16
Add initial Ruby support

javierhonduco:
This commit adds initial Ruby support by integrating rbperf's unwinder
and adapting it to this Agent.

There is more work to do, but in the spirit of keeping this commit a bit
more contained, right now:
- only Ruby 3.0.4 is supported;
- no focus on efficiency;
- missing clean-ups (such as clearing the interpreter stack maps);
- the Ruby stack is fetched even if the Ruby thread is sleeping as long
  as the native stack is on-CPU. This shows up confusingly as the Ruby stack in
many native stacks that aren't actually running the Ruby code;
- there are no integration tests, this will change soon;

All the shortcomings above, and then some, will be addressed in
forthcoming commits.

Architecture
============

The current BPF unwinding architecture as well as the changes that this
commit makes can be seen here: https://excalidraw.com/#json=E-W6lSLZ8rTOcri7HZdAl,4AjIUmb-olsQkdbwCqyxqg

image

Test Plan
=========

All current tests pass + running https://github.com/javierhonduco/rbperf/blob/main/tests/programs/simple_two_stacks.rb

image

    
### DIFF
--- a/bpf/cpu/rbperf.bpf.c
+++ b/bpf/cpu/rbperf.bpf.c
@@ -1,0 +1,436 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+//
+// Copyright (c) 2022 The rbperf authors
+
+#include "rbperf.h"
+
+#include "vmlinux.h"
+
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+
+#include "hash.h"
+#include "shared.h"
+
+/* struct {
+    // This map's type is a placeholder, it's dynamically set
+    // in rbperf.rs to either perf/ring buffer depending on
+    // the configuration.
+    __uint(type, BPF_MAP_TYPE_RINGBUF);
+} events SEC(".maps"); */
+
+struct {
+    __uint(type, BPF_MAP_TYPE_PROG_ARRAY);
+    __uint(max_entries, 3);
+    __type(key, u32);
+    __type(value, u32);
+} programs SEC(".maps");
+
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __uint(max_entries, 4096);
+    __type(key, u32);
+    __type(value, ProcessData);
+} pid_to_rb_thread SEC(".maps");
+
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __uint(max_entries, 64000);
+    __type(key, RubyFrame);
+    __type(value, u32);
+} frame_table SEC(".maps");
+
+struct {
+    __uint(type, BPF_MAP_TYPE_ARRAY);
+    __uint(max_entries, 12);
+    __type(key, u32);
+    __type(value, RubyVersionOffsets);
+} version_specific_offsets SEC(".maps");
+
+struct {
+    __uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+    __uint(max_entries, 1);
+    __type(key, u32);
+    __type(value, u64);
+} frame_index_storage SEC(".maps");
+
+struct {
+    __uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+    __uint(max_entries, 1);
+    __type(key, u32);
+    __type(value, SampleState);
+} global_state SEC(".maps");
+
+const volatile bool verbose = false;
+const volatile int num_cpus = 200; // Hard-limit of 200 CPUs.
+const volatile bool use_ringbuf = false;
+const volatile bool enable_pid_race_detector = false;
+const volatile enum rbperf_event_type event_type = RBPERF_EVENT_UNKNOWN;
+
+#define LOG(fmt, ...)                       \
+    ({                                      \
+        if (verbose) {                      \
+            bpf_printk(fmt, ##__VA_ARGS__); \
+        }                                   \
+    })
+
+static inline_method int read_syscall_id(void *ctx, int *syscall_id) {
+    return bpf_probe_read_kernel(syscall_id, SYSCALL_NR_SIZE, ctx + SYSCALL_NR_OFFSET);
+}
+
+static inline_method u32 find_or_insert_frame(RubyFrame *frame) {
+    u32 *found_id = bpf_map_lookup_elem(&frame_table, frame);
+    if (found_id != NULL) {
+        return *found_id;
+    }
+
+    u32 zero = 0;
+    u64 *frame_index = bpf_map_lookup_elem(&frame_index_storage, &zero);
+    // Appease the verifier, this will never fail.
+    if (frame_index == NULL) {
+        return 0;
+    }
+
+    // The previous __sync_fetch_and_add does not seem to work in 5.4 and 5.10
+    //  > libbpf: prog 'walk_ruby_stack': -- BEGIN PROG LOAD LOG --\nBPF_STX uses reserved fields
+    //
+    // Checking for the version does not work as these branches are not pruned
+    // in older kernels, so we shard the id generation per CPU.
+    u64 idx = *frame_index * num_cpus + bpf_get_smp_processor_id();
+    *frame_index += 1;
+
+    int err;
+    err = bpf_map_update_elem(&frame_table, frame, &idx, BPF_ANY);
+    if (err) {
+        LOG("[error] frame_table failed with %d", err);
+    }
+    return idx;
+}
+
+static inline_method void read_ruby_string(RubyVersionOffsets *version_offsets, u64 label, char *buffer,
+                                           int buffer_len) {
+    u64 flags;
+    u64 char_ptr;
+
+    rbperf_read(&flags, 8, (void *)(label + 0 /* .basic */ + 0 /* .flags */));
+
+    if (STRING_ON_HEAP(flags)) {
+        rbperf_read(&char_ptr, 8,
+                    (void *)(label + as_offset + 8 /* .long len */));
+        int err = rbperf_read_str(buffer, buffer_len, (void *)(char_ptr));
+        if (err < 0) {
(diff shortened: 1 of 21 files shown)


pkg/process/info.go
```diff
@@ -189,7 +217,168 @@ type Info struct {
...
-	Mappings Mappings
+	Interpreter *Interpreter
+	Mappings    Mappings
+}
+
+// fetchRubyInterpreterInfo receives a process pid and memory mappings and
+// figures out whether it might be a Ruby interpreter. In that case, it
+// returns an `Interpreter` structure with the data that is needed by rbperf
+// (https://github.com/javierhonduco/rbperf) to walk Ruby stacks.
+func fetchRubyInterpreterInfo(pid int, mappings Mappings) (*Interpreter, error) {
```
kakkoyun, line 228: ♻️ I think this deserves its own package, and we have a `runtime` package for this. Let's consider moving it.

pkg/profiler/cpu/cpu.go
```diff
@@ -301,7 +365,21 @@ func (p *CPU) addUnwindTableForProcess(pid int) {
 
 	level.Debug(p.logger).Log("msg", "adding unwind tables", "pid", pid)
 
-	err = p.bpfMaps.addUnwindTableForProcess(pid, nil, true)
+	// @nocommit: Improve.
+	procInfo, err := p.processInfoManager.Fetch(context.TODO(), pid)
```
kakkoyun, line 369: Probably it should get the context from the caller. Or consider passing down the process info from the profiler.

pkg/profiler/cpu/cpu.go
```diff
@@ -331,9 +409,19 @@ func (p *CPU) addUnwindTableForProcess(pid int) {
...
+	procInfo, err := p.processInfoManager.Fetch(ctx, pid)
+	if err != nil {
 		level.Debug(p.logger).Log("msg", "failed to prefetch process info", "pid", pid, "err", err)
 	}
+
+	// TODO: This should only be called once.
+	if procInfo.Interpreter != nil {
+		err := p.bpfMaps.addInterpreter(pid, *procInfo.Interpreter)
+		if err != nil {
+			// Must never fail.
```
kakkoyun, line 421: Still, let's not panic and stop the whole profiler. There might be other processes we would be interested in the system.

pkg/profile/profile.go
```diff
@@ -23,10 +23,11 @@ type ProcessRawData struct {
...
 
 type RawSample struct {
-	TID         PID
-	UserStack   []uint64
-	KernelStack []uint64
-	Value       uint64
+	TID              PID
+	UserStack        []uint64
+	KernelStack      []uint64
+	InterpreterStack []uint64
```
kakkoyun, line 29: We should comment that the values here are not addresses but instead symbol IDs that need to be looked up in a BPF map.

pkg/profiler/cpu/cpu.go
```diff
@@ -673,7 +767,7 @@ func (p *CPU) Run(ctx context.Context) error {
 				pi.Mappings.ExecutableSections(),
 				p.LastProfileStartedAt(),
 				samplingPeriod,
-			).Convert(ctx, perProcessRawData.RawSamples)
+			).Convert(ctx, perProcessRawData.RawSamples, p.bpfMaps.InterpreterFrames)
```
kakkoyun, line 770: 🔧 I believe `p.bpfMaps.InterpreterFrames` should be passed to the `NewConverter` factor method. It's used as a lookup dependency, so it better fits there as a dependency.

pkg/profiler/cpu/cpu.go
```diff
@@ -886,6 +982,14 @@ func (p *CPU) obtainRawData(ctx context.Context) (profile.RawData, error) {
 			}
 		}
 
+		if key.InterpreterStackID != 0 {
+			// TODO: Improve error handling.
+			interpErr := p.bpfMaps.readInterpreterStack(key.InterpreterStackID, interpreterStack)
```
kakkoyun, line 987: I think this method should return the actually read symbol table from BPF same, and `obtainRawData` should return it to the caller. It makes more sense to follow.

pkg/profiler/cpu/maps.go
```diff
@@ -185,15 +193,21 @@ func clearBpfMap(bpfMap *bpf.BPFMap) error {
...
-	processInfo      *bpf.BPFMap
+	stackCounts                *bpf.BPFMap
+	stackTraces                *bpf.BPFMap
+	dwarfStackTraces           *bpf.BPFMap
+	interpreterStackTraces     *bpf.BPFMap
+	pidToRubyThread            *bpf.BPFMap
+	rubyVersionSpecificOffsets *bpf.BPFMap
+	frameTable                 *bpf.BPFMap
+	processInfo                *bpf.BPFMap
+	InterpreterFrames          map[uint32]string
```
kakkoyun, line 210: I think this shouldn't be a field. It should be instead returned from the method (please check my other comment). What do you think?

bpf/cpu/shared.h
```diff
@@ -0,0 +1,85 @@
...
+    bpf_printk("[error] bpf_map_lookup_or_try_init with ret: %d", err);
+    return 0;
+  }
+
+  return bpf_map_lookup_elem(map, key);
+}
+
+// To be called once we are completely done walking stacks and we are ready to
+// aggregate them in the 'counts' map and end the execution of the BPF program(s).
+#define aggregate_stacks() \
```
kakkoyun, line 75: ❤️

bpf/cpu/rbperf.bpf.c
```diff
@@ -0,0 +1,438 @@
...
+        // profiling is the one we expect, we check the pid + start_time
+        // of the process.
+        //
+        // When we start profiling, the start_time will be zero, so we set
+        // it to the actual start time. Otherwise, we check that the start_time
+        // of the process matches what we expect. If it's not the case, bail out
+        // early, to avoid profiling the wrong process.
+        if (enable_pid_race_detector) {
+            u64 process_start_time;
+            bpf_core_read(&process_start_time, 8, &task->start_time);
```
kakkoyun, line 351: Do we provide this in the parca agent user-space code?

kakkoyun (commented):
Awesome 🏆 🥇 💯

I liked the additions to the rbperf way of sending data for adapting to the parca agent way.
BPF code looks pretty solid to me.

I have left some comments for user space and as you mentioned in your comments there is a lot of room for optimizations.

I think it's sure close to be merged.

pkg/process/info.go
```diff
@@ -189,7 +217,168 @@ type Info struct {
...
+	var rubyExecutable string
+	if librubyBaseAddress == nil {
+		rubyExecutable = path.Join("/proc/", fmt.Sprintf("%d", pid), "/exe")
+	} else {
+		rubyExecutable = path.Join("/proc/", fmt.Sprintf("%d", pid), "/root/", librubyPath)
+	}
+
+	// Read the Ruby version.
+	//
+	// PERF(javierhonduco): Using Go's ELF reader in the stdlib is very
```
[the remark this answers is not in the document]
javierhonduco: As discussed, let's fix this in a future PR

pkg/profiler/cpu/cpu.go
javierhonduco: Agreed, this was a stray panic and we should not stop the world if this fails

pkg/profiler/cpu/maps.go
javierhonduco: I did this thinking on some use cases but completely forgot what it is, so maybe it was something silly and unfounded? 😄. Will change to what you suggested

bpf/cpu/rbperf.bpf.c
javierhonduco: No because this feature is currently disabled https://github.com/parca-dev/parca-agent/blob/e5a1dc8e844d5cd19159471e745a8f63d1ef7e17/bpf/cpu/rbperf.bpf.c#L72. If we decide to enable this in the Agent (we should do this in the entry-point program, though) we should do it in another PR I think

pkg/process/info.go
javierhonduco: If it's ok with you, I would like to do this in a forthcoming PR

kakkoyun: As long as we don't forget about it 👍

javierhonduco: 💯

bpf/cpu/rbperf.bpf.c
```diff
@@ -0,0 +1,438 @@
...
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+
+#include "hash.h"
+#include "shared.h"
+
+extern int LINUX_KERNEL_VERSION __kconfig;
+
+/* struct {
```
kakkoyun, line 20: nit: Shall we remove it?

kakkoyun (approved):

kakkoyun:
I'll retrigger the CI soon.

kakkoyun:
Fix for container step https://github.com/parca-dev/parca-agent/pull/1964

pkg/profiler/cpu/cpu.go
```diff
@@ -958,32 +1068,43 @@ func preprocessRawData(rawData map[profileKey]map[combinedStack]uint64) profile.
...
 			userStack := make([]uint64, userStackDepth)
 			kernelStack := make([]uint64, kernelStackDepth)
+			interpreterStack := make([]uint64, interpreterStackDepth)
 
 			copy(userStack, stack[:userStackDepth])
 			copy(kernelStack, stack[stackDepth:stackDepth+kernelStackDepth])
+			copy(interpreterStack, stack[stackDepth*2:stackDepth*2+interpreterStackDepth])
 
 			p.RawSamples = append(p.RawSamples, profile.RawSample{
-				TID:         profile.PID(pKey.tid),
```
kakkoyun, line 983: Oppsie @javierhonduco 🙈

PR #1943, opened 2023-08-22
tests: Add integration tests for mixed-mode unwinding

javierhonduco:
aka JIT <-> DWARF unwinding. This also tests that the symbolization using
perfdump works.

Also changed the running time for the profiler during non-CI runs to
reduce feedback cycles.

### DIFF
--- a/test/integration/profiler_test.go
+++ b/test/integration/profiler_test.go
@@ -17,6 +17,7 @@ package integration
 import (
 	"context"
 	"fmt"
+	"os"
 	"os/exec"
 	"strconv"
 	"strings"
@@ -24,6 +25,7 @@ import (
 	"time"
 
 	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
 	pprofprofile "github.com/google/pprof/profile"
 	profilestorepb "github.com/parca-dev/parca/gen/proto/go/parca/profilestore/v1alpha1"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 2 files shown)


javierhonduco:
Mmm the JITed stacks are consistently failing in CI...

test/integration/profiler_test.go
```diff
@@ -319,12 +343,26 @@ func TestAnyStackContains(t *testing.T) {
...
 }
 
+// profileDuration sets the profile runtime to a shorter time period
+// when running outside of CI. The logic for this is that very loaded
+// systems, such as GH actions might take a long time to spawn processes.
+// By increasing the runtime we reduce the chance of flaky test executions,
+// but we shouldn't have to pay this price during local dev.
+func profileDuration() time.Duration {
+	_, ok := os.LookupEnv("CI")
+	panic(ok)
```
kakkoyun, line 353: This will always panic regarless of the `ok` value. I assume this is not the intendent behaviour.

test/integration/profiler_test.go
```diff
@@ -341,19 +379,26 @@ func TestCPUProfilerWorks(t *testing.T) {
 	defer noFramePointersCmd.Process.Kill()
 	dwarfUnwoundPid := noFramePointersCmd.Process.Pid
 
+	// Test unwinding JIT without frame pointers in the AoT code.
+	jitCmd := exec.Command("../../testdata/out/x86/basic-cpp-jit-no-fp")
+	err = jitCmd.Start()
+	require.NoError(t, err)
+	defer jitCmd.Process.Kill()
```
kakkoyun, line 386: I'd prefer `t.Cleanup` instead of `defer` in the tests. In this case it won't change anything but with nested tests behaviour would change.

kakkoyun:
I have added some commets. I haven't checked the failures but one of those could be the reason of constant failure.

test/integration/profiler_test.go
javierhonduco: Added this for testing, as it seemed to be running the wrong branch (see commit name)

test/integration/profiler_test.go
javierhonduco: Sure thing

javierhonduco:
In the spirit of pragmatism I have disabled the assertions of JIT'd stacks in CI. Will fix this later on. With these changes, we will test these code paths locally, and the iteration times will be reduced, too.

kakkoyun (approved):
LGTM 🥇

Issue #1797, opened 2023-06-26
unwinding/dwarf:  arm64 support


javierhonduco:
### Context

While we currently build arm64 containers for the Agent, we don't truly support BPF-based profiling anymore. This is due to a complete rework of the core BPF bits to unwind through code without frame pointers that focused on supporting x86_64.

We want to change this, as there's an ever-growing amount of developers using this architecture locally, see Apple Silicon, as well as in production ([Graviton](https://www.techpowerup.com/275971/aws-arm-based-graviton-processors-sees-the-biggest-growth-in-instance-share) et al).

#### Enabling work
- [x] https://github.com/parca-dev/parca-agent/commit/9f6546ae1887151fd9f4b8839d93982d7a8fb1fe Per-architecture vmlinux.h. As of this time, we only have [the one for x86_64](https://github.com/parca-dev/parca-agent/commit/97a2a3e76a9fcdc5a87806f20ba49c1896f5820f9), we need to have one per architecture, and choose the right one at build time. This is needed CO-RE and BTF support ;
- [x] https://github.com/parca-dev/parca-agent/commit/abdaf108a7c343b43d57909ca6e0356b2ade8cc9 Build the agent in arm64 (we only release it now, but we don't build it in normal CI runs AFAIK);
- [x] Run CI tests in arm64 to ensure further changes (build tags, unwinder changes, etc) don't break our code under this architecture;

#### Adding frame pointer unwinding
- [x] https://github.com/parca-dev/parca-agent/commit/9f6546ae1887151fd9f4b8839d93982d7a8fb1fe Frame pointer unwinding is significantly simpler than DWARF-based unwinding. Very small changes are required both in the BPF code and userspace to make it work, so we'll start here, focusing on Go apps;

#### DWARF-based unwinding (https://github.com/parca-dev/parca-agent/issues/1209)
- [x] Ensure our unwind info reader and parser can deal with arm64 binaries (e.g. the info being in the same ELF section, always being present according to its ABI, etc)
- [x] Adapt our unwind info parser/interpreter to work with multiple architectures, particularly, arm64;
- [x] Adapting the compact unwind row, as well as the unwinder code for arm64's saved return address (link register);

#### Other things to look out for:
- [x] Hardcoded DWARF expressions (plt1 and plt2);
- [x] Pointer authentication;
- [ ] Split stacks;
[...]

PR #1953, opened 2023-08-28
pkg/stack/unwind:  Add unwind tables support for Arm64

Sylfrena:
### What?
Add support for extracting unwind table information from the ".eh-frame"
section of Arm64 binaries. So far, only X86 binaries were supported.
    
 - Arm64 majorly differs from X86 in having a "Link Register(x30)" which saves the Return Address
 - The changes are only limited to userspace to generate "compact-unwind-tables"; we do
 not send the content of the Link Register to the BPF side- this will be done in follow-ups.
 - Enable DWARF-unwinding for Arm64 in "main.go" to enable building unwind-tables.
 - `CompactUnwindTableRow` has a new field `lrOffset` which holds the offset for the Link Register.
    This is initiialised to 0 for X86 architectures.

### Test Plan

```
$ go test  ./pkg/stack/unwind -count=1
ok      github.com/parca-dev/parca-agent/pkg/stack/unwind       0.007s
```

- tested for `libc.so.6`, `basic-cpp-no-fp`, `systemd` on both Arm64 and X86
    using `make-test-dwarf-unwind-tables`

- tables generated using "dist/eh-frame" verified against `readelf -wF` output

``` 
 1. Run `make build` in `parca-agent/testdata`
    `cd testdata && make build`
 2. Run `make build && make-test-dwarf-unwind-tables` from parca-agent root on the main branch.
    `cd .. && make build && make-test-dwarf-unwind-tables`
 3. Copy the output in `testdata/tables` and `testdata/compact_tables` in a separate untracked folder
    `mkdir ~/testdata-arm64-main && cp -r testdata/tables testdata/compact_tables ~/testdata-arm64-main`
 4. Switch to current `dwarf-unwinding-arm64` branch
 5. Run `make build` in parca-agent root. (Make sure you don't do this in the `testdata` submodule)
 6. Run `make-test-dwarf-unwind-tables`.
 7. Compare the output in `testdata/tables` and `testdata/compact_tables` to the results saved earlier in
    `~/testdata-arm64-main`
 8. Verify all tables against `readelf -wF "required-binary"`
 9. Repeat above for x86 binaries.
```

Check #1797 for Arm64 support roadmap

### DIFF
--- a/internal/dwarf/frame/table.go
+++ b/internal/dwarf/frame/table.go
@@ -24,18 +24,24 @@ const (
 	RuleCFA // Value is rule.Reg + rule.Offset
 )
 
-// From 3.4.1 Initial Stack and Register State
+// NOTE:
+// Each register in arm64 or x86 has a DWARF Register Number mapped
+// to its architecture specific Register Name defined in its respective spec.
+// The register numbers are not arbitrary constants but obtained from the spec
+// linked below for each architecture.
+//
+// From 3.4.1 Initial Stack and Register State for x86_64
 // https://refspecs.linuxbase.org/elf/x86_64-abi-0.99.pdf
+// From 4.1 DWARF Register Names for Aarch64/Arm64
+// https://github.com/ARM-software/abi-aa/blob/2023q1-release/aadwarf64/aadwarf64.rst#dwarf-register-names
 const (
-	X86_64FramePointer = 6 // $rbp
-	X86_64StackPointer = 7 // $rsp
+	X86_64FramePointer = 6  // $rbp
+	X86_64StackPointer = 7  // $rsp
+	Arm64FramePointer  = 29 // $fp // assumption: frame pointers are not stripped
+	Arm64StackPointer  = 31 // $sp or $r31
+	Arm64LinkRegister  = 30 // $x30 or $lr
+	// Arm64ProgramCounter = 32 //$x32; may be needed later while unwinding.
 )
-
-type UnwindRegisters struct {
-	StackPointer DWRule
-	FramePointer DWRule
-	SavedReturn  DWRule
-}
 
 // DWRule wrapper of rule defined for register values.
 type DWRule struct {
@@ -45,10 +51,16 @@ type DWRule struct {
 	Expression []byte
 }
 
+type UnwindRegisters struct {
+	StackPointer DWRule
+	FramePointer DWRule
+	SavedReturn  DWRule // save LinkRegister in here TODO(sylfrena)
+}
+
 // InstructionContext represents each object code instruction
 // that we have unwind information for.
 type InstructionContext struct {
-	loc           uint64
+	loc           uint64 // holds the PC
 	CFA           DWRule
 	Regs          UnwindRegisters
 	initialRegs   UnwindRegisters
@@ -393,26 +405,28 @@ func lookupFunc(instruction byte, ctx *Context) instruction {
 	return fn
 }
 
+// TODO(sylfrena): Reuse types.
 func setRule(reg uint64, frame *InstructionContext, rule DWRule) {
 	switch reg {
-	case X86_64StackPointer:
+	case Arm64StackPointer, X86_64StackPointer:
 		frame.Regs.StackPointer = rule
-	case X86_64FramePointer:
+	case Arm64FramePointer, X86_64FramePointer:
 		frame.Regs.FramePointer = rule
-	case frame.RetAddrReg:
+	case Arm64LinkRegister, frame.RetAddrReg: // TODO(sylfrena): should I just let it remain or reuse?
 		frame.Regs.SavedReturn = rule
 	}
 }
 
 func restoreRule(reg uint64, frame *InstructionContext) {
 	switch reg {
-	case X86_64StackPointer:
+	// TODO(sylfrena): Reuse types
+	case Arm64StackPointer, X86_64StackPointer:
 		if frame.initialRegs.StackPointer.Rule == RuleUnknown {
 			frame.Regs.StackPointer = DWRule{Rule: RuleUndefined}
 		} else {
 			frame.Regs.StackPointer = DWRule{Offset: frame.initialRegs.StackPointer.Offset, Rule: RuleOffset}
 		}
-	case X86_64FramePointer:
+	case Arm64FramePointer, X86_64FramePointer:
 		if frame.initialRegs.FramePointer.Rule == RuleUnknown {
 			frame.Regs.FramePointer = DWRule{Rule: RuleUndefined}
 		} else {
--- a/pkg/stack/unwind/compact_unwind_table.go
+++ b/pkg/stack/unwind/compact_unwind_table.go
@@ -15,6 +15,7 @@
 package unwind
 
 import (
+	"debug/elf"
 	"fmt"
 	"sort"
 
@@ -23,8 +24,11 @@ import (
 
 type BpfCfaType uint16
 
+// Constants are just to denote the rule type of calculation we do
+// i.e whether we should compute based on rbp or rsp.
 const (
-	//nolint: deadcode,varcheck
+	//nolint: deadcode,varcheck,unused
+	// iota assigns a value to constants automatically.
 	cfaTypeUndefined BpfCfaType = iota
 	cfaTypeRbp
 	cfaTypeRsp
@@ -43,21 +47,22 @@ const (
 )
 
 // CompactUnwindTableRows encodes unwind information using 2x 64 bit words.
+// `lrOffset` is the link register for arm64; it is initialized to 0 for x86.
 type CompactUnwindTableRow struct {
-	pc                uint64
-	_reservedDoNotUse uint16
-	cfaType           uint8
-	rbpType           uint8
-	cfaOffset         int16
-	rbpOffset         int16
+	pc        uint64
+	lrOffset  int16
+	cfaType   uint8
+	rbpType   uint8
+	cfaOffset int16
+	rbpOffset int16
 }
 
 func (cutr *CompactUnwindTableRow) Pc() uint64 {
 	return cutr.pc
 }
 
-func (cutr *CompactUnwindTableRow) ReservedDoNotUse() uint16 {
-	return cutr._reservedDoNotUse
+func (cutr *CompactUnwindTableRow) LrOffset() int16 {
+	return cutr.lrOffset
 }
 
 func (cutr *CompactUnwindTableRow) CfaType() uint8 {
@@ -88,13 +93,13 @@ func (t CompactUnwindTable) Swap(i, j int)      { t[i], t[j] = t[j], t[i] }
 
 // BuildCompactUnwindTable produces a compact unwind table for the given
 // frame description entries.
-func BuildCompactUnwindTable(fdes frame.FrameDescriptionEntries) (CompactUnwindTable, error) {
+func BuildCompactUnwindTable(fdes frame.FrameDescriptionEntries, arch elf.Machine) (CompactUnwindTable, error) {
 	table := make(CompactUnwindTable, 0, 4*len(fdes)) // heuristic: we expect each function to have ~4 unwind entries.
 	for _, fde := range fdes {
 		frameContext := frame.ExecuteDwarfProgram(fde, nil)
 		for insCtx := frameContext.Next(); frameContext.HasNext(); insCtx = frameContext.Next() {
 			row := unwindTableRow(insCtx)
-			compactRow, err := rowToCompactRow(row)
+			compactRow, err := rowToCompactRow(row, arch)
 			if err != nil {
 				return CompactUnwindTable{}, err
 			}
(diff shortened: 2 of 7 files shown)


Sylfrena:
This is still a WIP - some work still needs to be done for unit tests and testing a few other binaries. Apart from that, comments and a lot of `println` debugging need to be cleaned up as well- but they are helpful for testing at,- so please ignore them while reviewing atm.

pkg/stack/unwind/compact_unwind_table.go
```diff
@@ -109,20 +121,27 @@ func BuildCompactUnwindTable(fdes frame.FrameDescriptionEntries) (CompactUnwindT
...
+	var lrOffset int16
 
 	// CFA.
 	//nolint:exhaustive
 	switch row.CFA.Rule {
 	case frame.RuleCFA:
-		if row.CFA.Reg == frame.X86_64FramePointer {
+		// the values for registers are just numerical constants and can overlap(TODO: use a better word)
+		// so checking for machine architecture of the binary is necessary
+		if row.CFA.Reg == frame.X86_64FramePointer && arch == elf.EM_X86_64 {
```
kakkoyun, line 137: For me, moving the arch check to a level above would make it easier. And group the branches per arch rather than inlining here.

kakkoyun (commented):
LGTM ❤️ 💚 💙 💛 💜 
Just one readability suggestion.

I think if you could clean it up, we can merge it.

pkg/stack/unwind/compact_unwind_table.go
```diff
@@ -150,29 +185,49 @@ func rowToCompactRow(row *UnwindTableRow) (CompactUnwindTableRow, error) {
...
 	return CompactUnwindTableRow{
-		pc:                row.Loc,
-		_reservedDoNotUse: 0,
-		cfaType:           cfaType,
-		rbpType:           rbpType,
-		cfaOffset:         cfaOffset,
-		rbpOffset:         rbpOffset,
+		pc:        row.Loc,
+		lrOffset:  lrOffset, //TODO(sylfrena): dummy value; add offset here instead
+		cfaType:   cfaType,
```
Sylfrena, line 215: remove comment

pkg/stack/unwind/unwind_table.go
```diff
@@ -56,9 +57,56 @@ func x64RegisterToString(reg uint64) string {
...
+	arm64Regs := []string{
+		"x0", "x1", "x2", "x3", "x4", "x5", "x6", "x7", "x8", "x9", "x10", "x11",
+		"x12", "x13", "x14", "x15", "x16", "x17", "x18", "x18", "x19", "x20",
+		"x21", "x22", "x23", "x24", "x25", "x26", "x27", "x28", "x29", "x30",
+		"sp", "pc", "cpsr",
+	}
+
+	return arm64Regs[reg]
+}*/
+
```
Sylfrena, line 79: clean up but keep the comments

pkg/stack/unwind/unwind_table.go
```diff
@@ -181,7 +269,7 @@ func BuildUnwindTable(fdes frame.FrameDescriptionEntries) UnwindTable {
 
 	for _, fde := range fdes {
 		frameContext := frame.ExecuteDwarfProgram(fde, nil)
-		for insCtx := frameContext.Next(); frameContext.HasNext(); insCtx = frameContext.Next() {
+		for insCtx := frameContext.Next(); frameContext.HasNext(); insCtx = frameContext.Next() { //why framecontext.hasnext two times?
```
Sylfrena, line 272: remove comment

pkg/stack/unwind/compact_unwind_table.go
Sylfrena: Totally agree! But how do you feel about putting in a TODO or opening an issue for this now and fixing in a follow-up PR(after initial arm64 DWARF unwinding is merged)? I actually would like to reuse the constant types in which case we maybe can do away with the arch checks entirely.

bpf/cpu/cpu.bpf.c
```diff
@@ -55,6 +55,9 @@ _Static_assert(1 << MAX_BINARY_SEARCH_DEPTH >= MAX_UNWIND_TABLE_SIZE, "unwind ta
 // Special values.
 #define CFA_TYPE_END_OF_FDE_MARKER 4
 
+#define CFA_TYPE_FP_ARM 5
+#define CFA_TYPE_SP_ARM 6
```
javierhonduco, line 59: As discussed offline, let's try to unify these. If in the future we realise that for arm64 we need some different fields / ids, we can always change it back.

The benefit of keeping it unified is that is a bit more clear and elegant perhaps, but also that it will reduce the amount of code we'll need in the BPF programs. While it won't likely be a significant increase, every instruction we do not run matters

internal/dwarf/frame/table.go
```diff
@@ -24,19 +24,26 @@ const (
 	RuleCFA // Value is rule.Reg + rule.Offset
 )
 
-// From 3.4.1 Initial Stack and Register State
+// NOTE:
+// Each register in arm64 or x86 has a DWARF Register Number mapped
```
javierhonduco, line 28: Thanks for adding these, I should have added it for x86 a while back, my bad

internal/dwarf/frame/table.go
```diff
@@ -24,19 +24,26 @@ const (
...
+// https://github.com/ARM-software/abi-aa/blob/2023q1-release/aadwarf64/aadwarf64.rst#dwarf-register-names
+// TODO(sylfrena): Check the notes for arm64 against the register numbers^.
 const (
-	X86_64FramePointer = 6 // $rbp
-	X86_64StackPointer = 7 // $rsp
+	X86_64FramePointer = 6  // $rbp
+	X86_64StackPointer = 7  // $rsp
+	Arm64FramePointer  = 29 // $fp // assumption: frame pointers are not stripped
+	Arm64StackPointer  = 31 // $sp or $r31
+	Arm64LinkRegister  = 30 // $x30 or $lr
```
javierhonduco, line 43: Same as above, let's try to unify these

pkg/kernel/common.go
```diff
@@ -37,12 +36,16 @@ func int8SliceToString(arr []int8) string {
...
+		if err := syscall.Uname(&uname); err != nil {
+			return "", err
+		}
 
-	return int8SliceToString(uname.Release[:]), nil
+		return int8SliceToString(uname.Release[:]), nil
+	*/
+	// TODO(sylfrena): unhardcode this, no clue why this shows an error
+	return "5.19.0-46-generic", nil
 }
```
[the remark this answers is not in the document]
javierhonduco: If this is failing on your box this is something we should address because it could mean that this code is buggy

pkg/stack/unwind/compact_unwind_table.go
```diff
@@ -15,20 +15,27 @@
...
 	//nolint: deadcode,varcheck
+	// iota assigns a value to constants automatically
 	cfaTypeUndefined BpfCfaType = iota
 	cfaTypeRbp
 	cfaTypeRsp
 	cfaTypeExpression
 	cfaTypeEndFdeMarker
+	cfaTypeFp // cfa type is likely not defined for fp for arm64
+	cfaTypeSp // for arm64
+	cfaTypeLr // for arm64
```
javierhonduco, line 38: Same as above w.r.t. unifying these

pkg/stack/unwind/compact_unwind_table.go
javierhonduco: I possible let's do what Kemal suggested as it will make the code easier to understand, but we can always do later on if that's something that you would prefer

pkg/stack/unwind/compact_unwind_table.go
```diff
@@ -150,29 +185,49 @@ func rowToCompactRow(row *UnwindTableRow) (CompactUnwindTableRow, error) {
...
 
 	// Return address.
-	if row.RA.Rule == frame.RuleUndefined {
-		rbpType = uint8(rbpTypeUndefinedReturnAddress)
+	switch row.RA.Rule {
+	case frame.RuleOffset:
+		if arch == elf.EM_X86_64 {
+			// fmt.Println("entryyy")
+			// maybe just set RA rule to undefined for x86
+			lrOffset = 0
```
javierhonduco, line 193: Instead of having a special case here, why don't we store the actual value for every architecture? While we won't use it in the x86 unwinder, that way we avoid having a special case here, and perhaps the tests for the compact unwind table will be cleaner too. What do you think?

pkg/stack/unwind/compact_unwind_table_test.go
```diff
@@ -190,7 +191,179 @@ func TestCompactUnwindTable(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			have, err := CompactUnwindTableRepresentation(UnwindTable{test.input})
+			have, err := CompactUnwindTableRepresentation(UnwindTable{test.input}, elf.EM_X86_64.String())
```
javierhonduco, line 194: As we discussed, these two test tables are 99% the same, let's unify them too

pkg/stack/unwind/unwind_table.go
```diff
@@ -29,6 +29,7 @@ import (
 var (
 	ErrNoFDEsFound            = errors.New("no FDEs found")
 	ErrEhFrameSectionNotFound = errors.New("failed to find .eh_frame section")
+	ErrNoRegisterFound        = errors.New("failed to find register for architecture")
```
javierhonduco, line 32: This error is not handled anywhere, it's used more like a "default error" state, what about having that as a string directly rather than an error?

javierhonduco: Also now that I think about it it should be more like "architecture not supported", no?

pkg/stack/unwind/unwind_table.go
```diff
@@ -89,30 +137,35 @@ func (ptb *UnwindTableBuilder) PrintTable(writer io.Writer, path string, compact
...
 				}
 
+				// TODO(sylfrena): add for arm64
 				fmt.Fprintf(writer, "\t")
 				fmt.Fprintf(writer, "pc: %x ", compactRow.Pc())
 				fmt.Fprintf(writer, "cfa_type: %-2d ", compactRow.CfaType())
 				fmt.Fprintf(writer, "rbp_type: %-2d ", compactRow.RbpType())
 				fmt.Fprintf(writer, "cfa_offset: %-4d ", compactRow.CfaOffset())
 				fmt.Fprintf(writer, "rbp_offset: %-4d", compactRow.RbpOffset())
+				if arch == elf.EM_AARCH64 {
```
javierhonduco, line 152: Very nice! We def don't want to print this if we don't need it

pkg/stack/unwind/unwind_table.go
```diff
@@ -89,30 +137,35 @@ func (ptb *UnwindTableBuilder) PrintTable(writer io.Writer, path string, compact
...
 				case frame.RuleCFA:
-					CFAReg := x64RegisterToString(unwindRow.CFA.Reg)
-					fmt.Fprintf(writer, "\tLoc: %x CFA: $%s=%-4d", unwindRow.Loc, CFAReg, unwindRow.CFA.Offset)
+					CFAReg := registerToString(unwindRow.CFA.Reg, arch)
+					fmt.Fprintf(writer, "\tLoc: %x CFA: $%s=%-4d", unwindRow.Loc, CFAReg, unwindRow.CFA.Offset) // TODO(Sylfrena): correct
 				case frame.RuleExpression:
 					expressionID := ExpressionIdentifier(unwindRow.CFA.Expression)
 					if expressionID == ExpressionUnknown {
-						fmt.Fprintf(writer, "\tLoc: %x CFA: exp     ", unwindRow.Loc)
+						fmt.Fprintf(writer, "\tLoc: %x CFAunk: exp     ", unwindRow.Loc)
```
javierhonduco, line 166: 😝

javierhonduco (commented):
Looking good so far! A bit more work is needed so we can get this landed, but it's getting dangerously close 😄 .

One thing to think about as a follow-up (let's make a note, perhaps an issue or add a line in the tracking issue) about DWARF expressions, not sure if arm64 emits the same 2 PLT expressions we support and definitely the registers will be off, so we need to add special cases for them. Perhaps in the meantime, we can say that we don't support expressions at all and then change this later on?

pkg/stack/unwind/unwind_table.go
Sylfrena: Good catch! I do feel we should capture an error here and just log it for easier debugging later but open to other suggestions :D

pkg/kernel/common.go
```diff
@@ -66,7 +66,7 @@ func unameRelease() (string, error) {
 	}
 
 	var buf [65]byte
-	for i, b := range uname.Release {
+	for i, b := range uname.Release { // {
```
kakkoyun, line 69:
```suggestion
	for i, b := range uname.Release {
```

kakkoyun (approved):

PR #1966, opened 2023-08-30
Update testdata

Sylfrena:
Add arm64 binaries


javierhonduco:
This won't work as we are generating and writing the snapshot tests that haven't been created yet. We can either remove the writing from testdata or dump the snapshot tests and rebase this PR. Either one is fine :)

Issue #1979, opened 2023-09-05
Quick start failed with ebpf check


joeyleeeeeee97:
**Describe the bug**
$ ./parca-agent --node=test --remote-store-address=localhost:7070 --remote-store-insecure
ooooooooo.                                                  .o.                                            .
`888   `Y88.                                               .888.                                         .o8
 888   .d88'  .oooo.   oooo d8b  .ooooo.   .oooo.         .8"888.      .oooooooo  .ooooo.  ooo. .oo.   .o888oo
 888ooo88P'  `P  )88b  `888""8P d88' `"Y8 `P  )88b       .8' `888.    888' `88b  d88' `88b `888P"Y88b    888
 888          .oP"888   888     888        .oP"888      .88ooo8888.   888   888  888ooo888  888   888    888
 888         d8(  888   888     888   .o8 d8(  888     .8'     `888.  `88bod8P'  888    .o  888   888    888 .
o888o        `Y888""8o d888b    `Y8bod8P' `Y888""8o   o88o     o8888o `8oooooo.  `Y8bod8P' o888o o888o   "888"
                                                                      d"     YD
                                                                      "Y88888P'

level=info name=parca-agent ts=2023-09-05T09:13:55.974204229Z caller=main.go:330 msg="maxprocs: Leaving GOMAXPROCS=40: CPU quota undefined"
level=info name=parca-agent ts=2023-09-05T09:13:55.989982917Z caller=main.go:397 msg="eBPF is supported and enabled by the host kernel"
name=parca-agent ts=2023-09-05T09:13:56.011151701Z caller=main.go:551 msg=starting... node=test store=localhost:7070
level=warn name=parca-agent ts=2023-09-05T09:13:56.013255359Z caller=discovery_manager.go:196 msg="unable to start provider" provider=systemd/0 error="context canceled"
level=error name=parca-agent ts=2023-09-05T09:13:56.013784532Z caller=main.go:340 err="bpf check: perf event program type not supported: errno 0\nstack trace map type not supported: errno 0\nhash map type not supported: errno 0"

**To Reproduce**

**Expected behavior**

**Screenshots** or **Profiles** (using https://pprof.me)

**Logs**

**Software (please complete the following information):**
 - Parca Agent Version: [`parca-agent --version`]
 - Parca Server Version (if applicable): [`parca --version`]

**Workload (please complete the following information):**
- Runtime (if applicable):
- Compiler (if applicable):

**Environment (please complete the following information):**
[...]

PR #1992, opened 2023-09-07
Ensure we are running under root

javierhonduco:
This was previously discussed in
https://github.com/parca-dev/parca-agent/pull/719.

While ideally we could check for the individual capabilities this is fraught with issues: different kernel revisions might require different capabilities, over time we are adding new features that might require or drop the needed capabilities, and in general, capabilities don't seem to be a feature that many users are intimately familiar with.

Checking for the superuser is a reasonable thing to do as it will ensure we can run, and it's how every single deployment that we know of executes this software.

We can always re-evaluate this in the future, but in the meantime let's ensure that running the Agent does not result in surprises (https://github.com/parca-dev/parca-agent/issues/1979, but we've seen this happening multiple times).

As there were concerns regarding this check being to broad, I've added a escape hatch, which is disabled by default
(`--allow-running-as-non-root`).

Test Plan
========

**no root w/ escape hatch**
```
$ dist/parca-agent --allow-running-as-non-root
level=error name=parca-agent ts=2023-09-07T08:39:27.529651017Z caller=main.go:345 err="bpf check: perf event program type not supported: errno 0\nstack trace map type not supported: errno 0\nhash map type not supported: errno 0"
```

**no root**
```
$ dist/parca-agent
level=error name=parca-agent ts=2023-09-07T08:39:34.003842932Z caller=main.go:345 err="superuser (root) is required to run Parca Agent to load and manipulate BPF programs"
```

**root**

```
$ sudo dist/parca-agent
```

It works fine

### DIFF
--- a/cmd/parca-agent/main.go
+++ b/cmd/parca-agent/main.go
@@ -211,9 +211,10 @@ type FlagsDWARFUnwinding struct {
 	Mixed   bool `default:"true"                                    help:"Unwind using .eh_frame information and frame pointers"`
 }
 
-// FlagsHidden contains hidden flags. Hidden debug flags (only for debugging).
+// FlagsHidden contains hidden flags used for debugging or running with untested configurations.
 type FlagsHidden struct {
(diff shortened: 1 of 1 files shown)


kakkoyun (approved):
LGTM

Sylfrena (approved):

PR #2006, opened 2023-09-12
pkg/profiler: Fix unwinder metrics

javierhonduco:
Currently the unwinder metrics are not being updated and remain zero throughout profiling. The error in question:

```
readPerCpuCounter failed" error="get count values: failed to lookup value 0xc002720a40 in map percpu_stats: invalid argument
```

Was due to a mismatch of the structure in C and Go. In particular the `ErrorJitUnwindingMachinery` field only existed in Go, causing the read to fail as there were 8 extra bytes that were not expected.

Additionally, some other field was in the wrong position, which would have caused wrong data had the read succeed.

In the future we will fix this with a more automated binding generation. We have evaluated some options here but we weren't satisfied with the outcomes so far, but this is something we really want to tackle once we have the bandwidth.

Test Plan
=========

Tested locally, it works. I've added some integration tests which is not fully done yet, because I can't seem to figure out why the counters aren't bumped during tests. Will fix this later on.

```
[javierhonduco@fedora parca-agent]$ curl --silent http://localhost:7071/metrics | grep native | grep dwarf
parca_agent_native_unwinder_samples_total{unwinder="dwarf"} 8
parca_agent_native_unwinder_success_total{unwinder="dwarf"} 8
parca_agent_native_unwinder_success_total{unwinder="dwarf_reach_bottom"} 8
parca_agent_native_unwinder_success_total{unwinder="dwarf_to_jit"} 0
parca_agent_native_unwinder_success_total{unwinder="jit_to_dwarf"} 0
```

### DIFF
--- a/test/integration/profiler_test.go
+++ b/test/integration/profiler_test.go
@@ -17,6 +17,8 @@ package integration
 import (
 	"context"
 	"fmt"
+	"io"
+	"net/http"
 	"os"
 	"os/exec"
 	"strconv"
@@ -29,6 +31,7 @@ import (
 	pprofprofile "github.com/google/pprof/profile"
 	profilestorepb "github.com/parca-dev/parca/gen/proto/go/parca/profilestore/v1alpha1"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/procfs"
 	"github.com/prometheus/prometheus/model/relabel"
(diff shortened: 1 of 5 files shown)


test/integration/profiler_test.go
```diff
@@ -504,4 +525,40 @@ func TestCPUProfilerWorks(t *testing.T) {
...
+			resp.Body.Close()
+		})
+
+		body, err := io.ReadAll(resp.Body)
+		require.Nil(t, err)
+
+		metrics := parsePrometheusMetricsEndpoint(string(body))
+		require.Greater(t, len(metrics), 0)
+
+		// TODO: fix this assertion, all the BPF map metrics are zero but I am not sure why.
```
javierhonduco, line 559: Will create an issue to track this. As this works while running the Agent, I think this might be caused by a wrong setup/teardown of the code during testing. Spent some time debugging this but I haven't made much progress unfortunately

javierhonduco:
Thanks! Right now is not testing much, unfortunately, but once the commented out code is fixed it should ensure we don't have big regressions like this one

kakkoyun (approved):

PR #2026, opened 2023-09-19
Add panic reporting

javierhonduco:
(depends on https://github.com/parca-dev/parca/pull/3837)

Right now we don't really have a good understanding of the Agent's reliability beyond our demo / prod environment. Ideally we would like to know if any of our customers is hitting any issues such as unhandled panics, which can be due to explicit assertions, segmentation faults (we use CGO), etc.

Unfortunately Go doesn't provide the facilities that other programming languages and runtimes have to do top-level "exception" detection, as it lacks `atexit()` and `recover()` only works for the currently executing goroutine, which means that every single Goroutine (in our packages or third party dependencies) would have to be manually annotated. This is not practical or scalable.

In this commit an alternative approach is introduced. The idea is to use a wrapper binary that spawns the Agent. When the Agent exits it can read the last X Bytes of stderr and the status code and in case it can determine that it didn't exit cleanly, it can notify the backend.

Rather than adding another binary that will have to be not only executed in our OCI images, but also when running the Agent's executable directly we can have the Agent spawning itself but with a flag to indicate that the child process should not run the telemetry code, as it would result in an infinite loop of `fork()`.

A downside of this approach is that the Agent's `.text` and heap footprint is always going to be larger than a simple wrapper binary. To minimise the impact of this there are some things that help us out:
- Explicitely running the garbage collector before spawning the actual Parca Agent, to try to reduce memory as much as possible;
- The Linux's CoW will help with `.text` and other bits and bobs that won't have to be twice in memory;
- Most production setups should have [swap enabled](https://chrisdown.name/2018/01/02/in-defence-of-swap.html) which will move the cold memory pages from the telemetry program off memory;

Either way there's nothing like production so let's monitor the different mappings from our production environment and ensure that our expectations match reality.

Test Plan
=========

(note that we are running with non-main Parca until the changes get merged upstream)

**Forced panic**

```
$ sudo dist/parca-agent --node=test --remote-store-insecure --remote-store-address=localhost:7070 --force-panic
```

panics and sends RPC to the server

```
level=info name=parca ts=2023-09-19T10:29:07.207343254Z caller=telemetry.go:41 msg="agent panic'ed with" stderr="level=info name=parca-agent ts=2023-09-19T10:29:02.135401745Z caller=main.go:492 msg=\"maxprocs: Leaving GOMAXPROCS=12: CPU quota undefined\"\nlevel=info name=parca-agent ts=2023-09-19T10:29:02.185324025Z caller=main.go:562 msg=\"eBPF is supported and enabled by the host kernel\"\nname=parca-agent ts=2023-09-19T10:29:02.186802492Z caller=main.go:689 msg=starting... node=test store=localhost:7070\npanic: forced panic for testing purposes\n\ngoroutine 16 [running]:\nmain.main.func1.1(...)\n\tgithub.com/parca-dev/parca-agent/cmd/parca-agent/main.go:399\nmain.main.func1.2(...)\n\tgithub.com/parca-dev/parca-agent/cmd/parca-agent/main.go:403\nmain.main.func1.3(...)\n\tgithub.com/parca-dev/parca-agent/cmd/parca-agent/main.go:407\nmain.main.func1()\n\tgithub.com/parca-dev/parca-agent/cmd/parca-agent/main.go:410 +0x33\ncreated by main.main in goroutine 1\n\tgithub.com/parca-dev/parca-agent/cmd/parca-agent/main.go:395 +0x5f7\n" metadata="map[agent_version: cpu_cores:12 git_commit:b2602182d2f8393b09600b995e7f4ffb2d23fdc7 go_arch:amd64 kernel_release:6.4.6-200.fc38.x86_64]"
```

**Forced panic w/o telemetry**

```
$ sudo dist/parca-agent --node=test --remote-store-insecure --remote-store-address=localhost:7070 --force-panic --telemetry-disable-panic-telemetry
```

panics and doesn't send telemetry


cmd/parca-agent/main.go
```diff
@@ -267,6 +326,87 @@ func main() {
...
+		cmd := exec.Command(argsCopy[0], argsCopy[1:]...) //nolint:gosec
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = io.MultiWriter(os.Stderr, buf)
+
+		// Run garbage collector to minimize the amount of memory that the parent
+		// telemetry process uses.
+		runtime.GC()
+		err := cmd.Run()
+		if err != nil {
+			fmt.Fprintln(os.Stderr, "======================= unexpected error =======================")
```
javierhonduco, line 346: Want to keep things as bare-bones as possible in the error handler as anything could have happened, including out-of-memory conditions. We do a bunch of allocations below in the RPC itself anyways, but still 😄

cmd/parca-agent/main.go
```diff
@@ -214,6 +219,10 @@ type FlagsDWARFUnwinding struct {
 	Mixed   bool `default:"true"                                    help:"Unwind using .eh_frame information and frame pointers"`
 }
 
+type FlagsTelemetry struct {
+	DisablePanicTelemetry bool `default:"false"`
```
kakkoyun, line 223: Maybe just `DisablePanic` or `DisableTrackingPanics` since we already have telemetry prefix?

cmd/parca-agent/main.go
kakkoyun: I think we should use our structured logging facilities. I would prefer to be able query these on our logging infrastructure for example.

kakkoyun (commented):
Looking good 👍 

I have dropped a couple of comments.

And sorry for brevity I reviewed it while having a walk 🐕‍🦺

javierhonduco:
Thanks for the review @kakkoyun! Will push the updated changes + the version bump as soon as the server PR lands

javierhonduco:
Since the last commit:

- feedback changes;
- updated parca to latest;
- make stderr size configurable;

kakkoyun (approved):

go.mod
```diff
@@ -204,3 +205,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.3.0 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace golang.org/x/exp => golang.org/x/exp v0.0.0-20230713183714-613f0c0eb8a1
```
javierhonduco, line 209: Just merged and forgot to add a comment about this, leaving it here for future reference. Can add it in the code if you prefer, too:

```
// Pin x/exp due to a breaking change (https://github.com/golang/exp/commit/302865e7556b4ae5de27248ce625d443ef4ad3ed#diff-ec22fd7f0aec27f1c6dc503e7dd3960c8f5ab781684218aa59b404ada0d48231R29)
// ```
// ../../go/pkg/mod/github.com/prometheus/prometheus@v0.47.0/model/labels/labels.go:365:23: type func(a Label, b Label) bool of func(a, b Label) bool {…} does not match inferred type func(a Label, b Label) int for func(a E, b E) int
// ```

PR #2028, opened 2023-09-19
Reduce size of object pool

javierhonduco:
512 elf.Files (+ other metadata) can really balloon memory usage. Assuming
each entry is ~10 MB would result in a heap usage of at least 5 GB. Some
of our users reported 10 GB in heap usage in the pool which is
completely unreasonable.

In the future we want to update this to have an adaptive mechanism so we
can use the cache as much as possible while keeping memory usage at bay.

Right now having a pool to small will result in more files being opened,
(elf.Open) which results in lots of allocations and high CPU usage.
Hence we have a tradeoff of CPU + allocations vs heap usage.

Test Plan
=========

Ran overnight without issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Flags:
                                    of memory that may be locked into RAM. It is
                                    used to ensure the agent can lock memory for
                                    eBPF maps. 0 means no limit.
-      --object-file-pool-size=512
+      --object-file-pool-size=100
                                    The maximum number of object files to keep in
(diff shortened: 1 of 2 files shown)


javierhonduco:
Only land after https://github.com/parca-dev/parca-agent/pull/2031 as it fixes a race condition that we believe it was affecting the stability of the object pool

kakkoyun:
@javierhonduco FYI, There is a formatting issue that causes CI to fail.

Issue #1962, opened 2023-08-29
Race condition in our LRU cache implementation


javierhonduco:
There is a data race that can be reproduced by running `make GO=$(which go) ENABLE_RACE=yes test/integration` in a loop. The report shows that pool.go might be using `keyCache` (and `objCache` once the former race has been fixed) fwithout protecting our data structures from concurrent access

```
WARNING: DATA RACE
Read at 0x00c000740480 by goroutine 112:
  container/list.(*List).MoveToFront()
      /usr/local/go/src/container/list/list.go:181 +0x4f
  github.com/parca-dev/parca-agent/pkg/cache/lru.(*LRU[go.shape.string,go.shape.struct { github.com/parca-dev/parca-agent/pkg/cache.value go.shape.struct { github.com/parca-dev/parca-agent/pkg/objectfile.path string; github.com/parca-dev/parca-agent/pkg/objectfile.buildID string; github.com/parca-dev/parca-agent/pkg/objectfile.modtime time.Time }; github.com/parca-dev/parca-agent/pkg/cache.deadline time.Time }]).Get()
      /home/javierhonduco/code/parca-agent/pkg/cache/lru/lru.go:81 +0x11e
  github.com/parca-dev/parca-agent/pkg/cache.(*LRUCacheWithTTL[go.shape.string,go.shape.struct { github.com/parca-dev/parca-agent/pkg/objectfile.path string; github.com/parca-dev/parca-agent/pkg/objectfile.buildID string; github.com/parca-dev/parca-agent/pkg/objectfile.modtime time.Time }]).Get()
      /home/javierhonduco/code/parca-agent/pkg/cache/cache_with_ttl.go:86 +0x137
  github.com/parca-dev/parca-agent/pkg/cache.(*LRUCacheWithTTL[string,github.com/parca-dev/parca-agent/pkg/objectfile.cacheKey]).Get()
      /home/javierhonduco/code/parca-agent/pkg/cache/cache_with_ttl.go:84 +0x84
  github.com/parca-dev/parca-agent/pkg/objectfile.(*Pool).Open()
      /home/javierhonduco/code/parca-agent/pkg/objectfile/pool.go:174 +0x8a
  github.com/parca-dev/parca-agent/pkg/metadata.Compiler.func1()
      /home/javierhonduco/code/parca-agent/pkg/metadata/compiler.go:62 +0x2c4
  github.com/parca-dev/parca-agent/pkg/metadata.(*StatelessProvider).Labels()
[...]

PR #2031, opened 2023-09-19
pkg/cache: Fix data races

javierhonduco:
Get() modify state, for LRU bookkeeping purposes, hence we need to use a writer lock.

Stop spawning goroutines for every callback as this introduces another data race.

Test Plan
=========

The data races reported by tsan (make ENABLE_RACE=yes) are gone!

Both Kemal and myself have ran this code with a very small object pool
over night without any issues

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -40,8 +40,8 @@ func (c *LRUCache[K, V]) Add(key K, value V) {
 }
 
 func (c *LRUCache[K, V]) Get(key K) (V, bool) {
-	c.mtx.RLock()
-	defer c.mtx.RUnlock()
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
 	return c.lru.Get(key)
 }
 
--- a/pkg/cache/cache_with_eviction.go
+++ b/pkg/cache/cache_with_eviction.go
@@ -14,12 +14,10 @@
 package cache
 
 import (
-	"context"
 	"errors"
 	"sync"
 
 	"github.com/prometheus/client_golang/prometheus"
-	"golang.org/x/sync/semaphore"
 
 	"github.com/parca-dev/parca-agent/pkg/cache/lru"
 )
@@ -36,15 +34,10 @@ func NewLRUWithEviction[K comparable, V any](reg prometheus.Registerer, maxEntri
 	if onEvictedCallback == nil {
 		return nil, errors.New("onEvictedCallback must not be nil")
 	}
-	limiter := semaphore.NewWeighted(5)
 	c := &LRUWithEviction[K, V]{
 		mtx: &sync.RWMutex{},
 		onEvictedCallback: func(k K, v V) {
-			if err := limiter.Acquire(context.Background(), 1); err != nil {
-				return
-			}
 			onEvictedCallback(k, v)
-			limiter.Release(1)
 		},
 	}
 	c.lru = lru.New[K, V](
@@ -57,7 +50,7 @@ func NewLRUWithEviction[K comparable, V any](reg prometheus.Registerer, maxEntri
 
 // onEvicted is called when an entry is evicted from the underlying LRU.
 func (c *LRUWithEviction[K, V]) onEvicted(k K, v V) {
-	go c.onEvictedCallback(k, v)
+	c.onEvictedCallback(k, v)
 }
 
 // Add adds a value to the cache.
@@ -69,8 +62,8 @@ func (c *LRUWithEviction[K, V]) Add(key K, value V) {
 
 // Get looks up a key's value from the cache.
 func (c *LRUWithEviction[K, V]) Get(key K) (V, bool) {
-	c.mtx.RLock()
-	defer c.mtx.RUnlock()
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
 	return c.lru.Get(key)
 }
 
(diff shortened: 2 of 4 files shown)


javierhonduco:
Should fix https://github.com/parca-dev/parca-agent/issues/1962

brancz (approved):

pkg/cache/cache_with_eviction.go
```diff
@@ -57,7 +57,10 @@ func NewLRUWithEviction[K comparable, V any](reg prometheus.Registerer, maxEntri
 
 // onEvicted is called when an entry is evicted from the underlying LRU.
 func (c *LRUWithEviction[K, V]) onEvicted(k K, v V) {
-	go c.onEvictedCallback(k, v)
```
kakkoyun, line 60: Let's remove the goroutine separately. It needs more testing and thought. There's already a semaphore to create a backpressure around this. I don't believe this is the root cause. It's there to ensure we don't pay the cost of eviction while adding items.

pkg/cache/cache_with_eviction.go
```diff
@@ -57,7 +57,10 @@ func NewLRUWithEviction[K comparable, V any](reg prometheus.Registerer, maxEntri
...
+	c.mtx.Lock()
```
kakkoyun, line 60: `onEvicted` only gets called by one of the exported functions, and they are already locked. I believe this could try to lock the already locked lock.

pkg/cache/cache.go
```diff
@@ -40,8 +40,8 @@ func (c *LRUCache[K, V]) Add(key K, value V) {
 }
 
 func (c *LRUCache[K, V]) Get(key K) (V, bool) {
-	c.mtx.RLock()
-	defer c.mtx.RUnlock()
+	c.mtx.Lock()
```
kakkoyun, line 43: 💯 👍

pkg/cache/cache_with_eviction.go
javierhonduco: Good call! I completely missed that there's a semaphore, great stuff

pkg/cache/lru/lru.go
```diff
@@ -113,10 +113,10 @@ func (c *LRU[K, V]) removeOldest() {
 func (c *LRU[K, V]) removeElement(e *list.Element) {
 	c.evictList.Remove(e)
 	kv := e.Value.(entry[K, V])
-	delete(c.items, kv.key)
 	if c.onEvicted != nil {
 		c.onEvicted(kv.key, kv.value)
 	}
+	delete(c.items, kv.key)
```
javierhonduco, line 119: Sooooo if we delete the element *before* calling the callback it could happen that we attempt to close the file after it's been garbage collected. If it happens concurrently (there is a finalizer on files by default) then it might race and attempt to close a closed file.

By deleting it afterwards we would prevent this... the problem is... onEvicted might run in a goroutine and if that's the case changing this order won't do it, we will have to either not run the `onEvicted` to run serially or do something fancier @kakkoyun

pkg/cache/cache_with_eviction.go
javierhonduco: See comment below

pkg/cache/lru/lru.go
javierhonduco: This is the issue that I think happens due to the above https://github.com/parca-dev/parca-agent/issues/1798

kakkoyun: Yes, this is a good catch. Let me think about this for a while.

kakkoyun: > Sooooo if we delete the element _before_ calling the callback it could happen that we attempt to close the file after it's been garbage collected.

Why do you think this is? Since we give a reference of the file to call the callback, GC wouldn't touch this file. Also if are we talking about this panic https://github.com/parca-dev/parca-agent/issues/1798 ? Because if it's the case it's a panic that we created https://github.com/parca-dev/parca-agent/blob/fee025041d725eb361487ec704dbd3eea80d265c/pkg/objectfile/object_file.go#L67-L72 So we try to read a file after we evict it from the cache.

pkg/cache/cache_with_eviction.go
javierhonduco: Good catch, thanks!

pkg/cache/lru/lru.go
javierhonduco: I think you are right:

```
kv := e.Value.(entry[K, V])
```

We get a reference to the value and then `onEvicted` and `delete` don't have to be serialized. I am assuming that the delete won't clear the backing memory for the entry, which I am not sure if it's true, though

pkg/cache/lru/lru.go
```diff
@@ -113,10 +113,10 @@ func (c *LRU[K, V]) removeOldest() {
...
 	c.metrics.evictions.Inc()
 }
```
javierhonduco, line 121: Related to the comment above, the `Purge()` method could be incorrect, though, as no reference kept and the `onEvicted` and `delete` calls will run concurrently with the code in `main`

kakkoyun (approved):

pkg/cache/lru/lru.go
javierhonduco: @kakkoyun realised that this is not correct, we should have a reference to key and value

PR #2037, opened 2023-09-20
Do not log ExecutableInfo races dealing with files

javierhonduco:
As they are verbose and not actionable


pkg/pprof/pprof.go
```diff
@@ -356,7 +356,6 @@ func (c *Converter) addExecutableInfo(
 		if !(os.IsNotExist(err) || errors.Is(err, fs.ErrNotExist)) {
 			// We don't want to log the error if the file doesn't exist,
 			// because it's probably a very short-lived process that weren't fast enough to obtain the FDs
-			level.Debug(c.logger).Log("msg", "failed to get executable info", "address", fmt.Sprintf("%x", addr), "err", err)
 		} else {
 			level.Warn(c.logger).Log("msg", "failed to get executable info", "address", fmt.Sprintf("%x", addr), "err", err)
```
javierhonduco, line 360: We also need to add one more check (perhaps in another PR), for `no such file or directory`, in case the file has been deleted

pkg/pprof/pprof.go
```diff
@@ -356,7 +356,6 @@ func (c *Converter) addExecutableInfo(
 		if !(os.IsNotExist(err) || errors.Is(err, fs.ErrNotExist)) {
```
kakkoyun, line 356: Maybe we should revert the condition and move the else clause here?

javierhonduco: Definitely :D

pkg/pprof/pprof.go
javierhonduco: Just realised that the current check is not very good... Just changed it

Sylfrena:
closing in favour of #2039

PR #2038, opened 2023-09-20
Add LFU cache

kakkoyun:
Adds a caching implementation for our object pool based on [http://dhruvbird.com/lfu.pdf](http://dhruvbird.com/lfu.pdf)

### Test Plan
- Run locally over extended period of time and capture profiles while running:
```
for i in $(seq 0 1000000); do echo "int lol_$i = 0;" >> src/basic-cpp.cpp &&  gcc src/basic-cpp.cpp -fno-PIE -no-pie -o out/x86/basic-cpp-no-fp -fomit-frame-pointer && readelf -n out/x86/basic-cpp-no-fp |& grep "Build ID" && timeout 10s out/x86/basic-cpp-no-fp; done
```

  - CPU: https://pprof.me/b65f6df
  - Mem In-use (bytes) https://pprof.me/8fd6ca9
  - Mem In-use (bytes) https://pprof.me/81610d2
  - Allocs Objects: https://pprof.me/67caa2f


javierhonduco:
Will take a look as soon as I can, probably next week. 

We should try seeing what the hit-cache ratio is for LRU / LFU for different cache sizes and plot them to understand how the two behave for our use case. Let's also check for this data on demo.

As we've discussed, perhaps we could make the eviction strategy configurable with a feature flag for some of the core caches (object pool)

kakkoyun:
> As we've discussed, perhaps we could make the eviction strategy configurable with a feature flag for some of the core caches (object pool)

Let's do this! We already have metrics in place. We can test different strategies on our demo environment. I'll add the flag now.

kakkoyun:
@javierhonduco Here you go https://github.com/parca-dev/parca-agent/pull/2038/commits/13e92370ffa77540eba8bf26620c45e330152003

PR #2040, opened 2023-09-21
bpf: rate limit events

javierhonduco:
We've seen some reports of boxes with overloaded CPU usage (unrelated to the Agent) that significantly less frames that we expected. The profiles are already incomplete before they are on the wire and after some investigation and carefully looking at the metrics, we could not spot anything abnormal (no normalization errors, no dropped samples, etc).

The current working theory is that the BPF events channel (we use a perf buffer atm) might be writing faster than we are reading, so we are losing events. While instrumenting the code to find whether this was the case, it seemed that libbpfgo's lost events channel never receives anything. This is something we are debugging in parallel as we really want to fix this and have metrics on lost events.

To test the hyphotesis of a overloaded CPU producing lots of events (lots of forks) I tested it on my machine and manually instrumented both BPF and userspace to count events sent vs. received and the differences were abysmal.

In this commit a small rate limiting for events is added with the aim of reducing the chances of losing events.

Future improvements
===================

Sending events in a perf buffer and then having an additional map to rate limit events is a bit silly, we could just use a map as an "events" channel and then read from userspace.

Benefits:
- No need for two maps, lower memory usage;
- Easier code;
- Stop relying on `.Poll()` which allocates a lot of memory;

Issues:
- BPF maps just use a hash, if there are any collisions we can't store both keys, this might result in some events being lost. We should calculate the probability of this happening;
- We will have to do very frequent BPF map reads. We might want to use batched ops.

For these reasons we will consider other alternatives in the future, but in the meantime let's see if this fixes what some users are seeing.

Test Plan
=========

https://pprof.me/46d50b5

+ tested the flags locally


brancz (approved):
Very nice approach!

bpf/cpu/cpu.bpf.c
```diff
@@ -321,22 +324,51 @@ static void bump_samples() {
...
+  if (!unwinder_config.event_rate_limits_enabled) {
+    return false;
+  }
+
+  int max_events_per_event_id = 2;
+
+  u32 one = 1;
+  u32* val = bpf_map_lookup_or_try_init(&events_count, &event_id, &one);
+  if (val) {
+    __sync_fetch_and_add(val, 1);
```
kakkoyun, line 337: Maybe I might be misremembering the details but this was failing on 5.4, no?

cmd/parca-agent/main.go
```diff
@@ -145,8 +145,9 @@ type flags struct {
 	Telemetry FlagsTelemetry `embed:"" prefix:"telemetry-"`
 	Hidden    FlagsHidden    `embed:"" hidden:""           prefix:""`
 
-	// TODO: Move to FlagsBPF once we have more flags.
-	VerboseBpfLogging bool `help:"Enable verbose BPF logging."`
+	BPF FlagsBPF `embed:"" prefix:"bpf-"`
```
kakkoyun, line 148: ❤️

javierhonduco:
Will wait for Kemal's review 🙇

bpf/cpu/cpu.bpf.c
javierhonduco: I ran the vmtests here and it works fine. For some reason that I am not sure yet the code generated here vs in the interpreter unwinders tickles the verifier differently. We need to check the assembly to understand what's going on.

cmd/parca-agent/main.go
```diff
@@ -238,6 +239,12 @@ type FlagsHidden struct {
 	ForcePanic bool `default:"false" help:"Panics the agent in a goroutine to test that telemetry works." hidden:""`
 }
 
+type FlagsBPF struct {
+	VerboseLogging         bool `help:"Enable verbose BPF logging."`
+	EventsBufferSize       int  `default:"8192"                     help:"Size in pages of the events buffer."`
```
javierhonduco, line 244: Oops, forgot to actually use this!

Sylfrena (approved):

bpf/cpu/cpu.bpf.c
javierhonduco: (We also use atomically bump the stack_counts without issues)

javierhonduco:
Ooops what

kakkoyun (approved):

pkg/profiler/cpu/maps.go
```diff
@@ -1118,10 +1126,28 @@ func (m *bpfMaps) cleanStacks() error {
 	return result
 }
 
+func (m *bpfMaps) cleanEventsCount() error {
+	if err := clearBPFMap(m.eventsCount); err != nil {
+		return err
```
kakkoyun, line 1131: Since https://github.com/parca-dev/parca-agent/pull/2041 is merged, could we add a metric for the clean as well?

javierhonduco: Great shout!

PR #2046, opened 2023-09-22
[do not merge] bpf: Reduce iteration size by 1

javierhonduco:
Our e2e tests in CI are failing with `argument list too long`. They seem to be running on `6.2.0-1011-azure`. So this failure is not expected but let's see if we can address this.

Reducing the size has a small performance penalty so it would be preferred to use a well-known kernel that works.

### DIFF
--- a/bpf/cpu/cpu.bpf.c
+++ b/bpf/cpu/cpu.bpf.c
@@ -21,19 +21,10 @@
 #define RUBY_UNWINDER_PROGRAM_ID 1
 #define PYTHON_UNWINDER_PROGRAM_ID 2
 
-#if __TARGET_ARCH_arm64
 // Number of frames to walk per tail call iteration.
 #define MAX_STACK_DEPTH_PER_PROGRAM 8
 // Number of BPF tail calls that will be attempted.
 #define MAX_TAIL_CALLS 16
-#endif
-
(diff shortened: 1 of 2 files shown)


javierhonduco:
One second, we are setting minikube with:

```
      - name: Set up Minikube
        uses: medyagh/setup-minikube@e0505489c98496aff6ed17db0ba68bf7ecadf730 # v0.0.14
        with:
          minikube-version: ${{ env.MINIKUBE_VERSION }}
          driver: none
          kubernetes-version: stable # Determined by Minikube version
```

So it will use the bare-metal OS, which makes sense because we saw `6.2.0-1011-azure` in the uname, but... the integration tests (not to be confused with e2e) always run in bare metal and passed without issues... 🤔

javierhonduco:
Forgot to mention that this won't be merged unless we really have to unblock us, we should get to the bottom of it.

javierhonduco:
Still failing with 

```
level=error name=parca-agent ts=2023-09-22T12:06:21.852478469Z caller=main.go:510 err="load bpf program: failed to load BPF object: argument list too long

```

but https://github.com/parca-dev/parca-agent/issues/2047

javierhonduco:
Superseded by https://github.com/parca-dev/parca-agent/pull/2059

PR #2049, opened 2023-09-22
e2e: Reduce waits

javierhonduco:
Rather than wait 5 minutes no matter how long it actually takes, wait directly in the tests.

Also, reduce some waiting times and timeouts that were too large, add a build rule to run tests locally and remove printed kernel version as this doesn't run in k8s and it might be not very clear.

15 minutes have been chosen because running in CI typically takes 10 minutes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -365,14 +365,16 @@ observable-dev/down:
 
 E2E_KUBECONTEXT := parca-e2e
 
+.PHONY: actions-e2e-local
+actions-e2e-local:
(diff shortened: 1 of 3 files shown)


javierhonduco:
```
panic: test timed out after 10m0s
```

javierhonduco:
We can maybe pass

```
-timeout=20m
```

javierhonduco:
Closing, but we should revisit when the e2e tests get improved

PR #2050, opened 2023-09-22
pkg/process: Only keep executable mappings

brancz:
We only ever care about executable mappings, never about any other kinds of mappings. Executable mappings are a small fraction of all mappings typically so this should save a significant amount of memory.

This change was made based on a customer reported heap profile: https://pprof.me/edde8e7/


Sylfrena (approved):

kakkoyun (approved):

javierhonduco:
Thanks for doing this -- I am pretty sure we had this filtering before but it must have been accidentally deleted somewhere 😢

Issue #2047, opened 2023-09-22
tests/e2e: Tests pass despite the agent not running


javierhonduco:
Parca Agent exited with an error but the e2e passed, this might be because of scraping enabled? If that's the case we should disable it to ensure we are testing the Agent <-> Server interation

https://github.com/parca-dev/parca-agent/actions/runs/6273544310/job/17037459567?pr=2040

PR #2059, opened 2023-09-25
unwinder/native: Accomodate larger program due to rate limits

javierhonduco:
The newly added rate limits increase the code size of the native unwinder. By default they are disabled, so the abstract paths that the verifier has to analyse are reduced, compared to when they are enabled.

As in Go, structs with fields that aren't explicitely named have a default value, in the case of booleans, false, this feature was disabled in unit and integration tests, working both locally and CI.

In the Agent binary we have this flag enabled by default. This worked in local testing because the kernel I run is new enough to allow for larger programs (6.4.6-200.fc38.x86_64) but both the e2e tests in CI and in our own Demo deployment, which run older kernels, the program fails to load with `Argument list too long`.

Enabled rate limiting by default in tests shows the issue before this commit. To fix it, the the iterations per program were reduced. We might want to revisit this in the future.

Test Plan
=========

With rate-limits enabled both integration and unit tests pass now.

```
$ make vmtest
=============
Test results:
=============
- ✅ 5.4
- ✅ 5.10
- ✅ 5.19
- ✅ 6.1
```

### DIFF
--- a/bpf/cpu/cpu.bpf.c
+++ b/bpf/cpu/cpu.bpf.c
@@ -21,19 +21,10 @@
 #define RUBY_UNWINDER_PROGRAM_ID 1
 #define PYTHON_UNWINDER_PROGRAM_ID 2
 
-#if __TARGET_ARCH_arm64
 // Number of frames to walk per tail call iteration.
-#define MAX_STACK_DEPTH_PER_PROGRAM 8
+#define MAX_STACK_DEPTH_PER_PROGRAM 7
 // Number of BPF tail calls that will be attempted.
-#define MAX_TAIL_CALLS 16
-#endif
-
-#if __TARGET_ARCH_x86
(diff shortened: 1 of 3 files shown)


Sylfrena (approved):

javierhonduco:
Because of https://github.com/parca-dev/parca-agent/issues/2047 and https://github.com/parca-dev/parca-agent/issues/2047, we need to take a look at the Agent logs in the e2e tests before merging (https://github.com/parca-dev/parca-agent/actions/runs/6295970858/job/17090368809?pr=2059). 

Let's roll this out to Demo before cutting a new release. Also heads up @Sylfrena as this might break in arm64. Very excited for the future kernel tests for arm64 to prevent these regressions!

kakkoyun (commented):
LGTM

Issue #1769, opened 2023-06-16
Failure to parca-agent


cuiweiqiang:
I want to start a binary exec in Linux, but I get some errors like this:

```bash
level=debug name=parca-agent ts=2023-06-16T05:02:05.93688625Z caller=cpu.go:233 msg="libbpf: Error loading .BTF into kernel: -22. BTF is optional, ignoring."
level=debug name=parca-agent ts=2023-06-16T05:02:05.937436165Z caller=cpu.go:233 msg="libbpf: map 'heap': created successfully, fd=8"
level=debug name=parca-agent ts=2023-06-16T05:02:06.04057302Z caller=cpu.go:233 msg="libbpf: map 'unwind_tables': created successfully, fd=9"
level=debug name=parca-agent ts=2023-06-16T05:02:06.052532107Z caller=cpu.go:233 msg="libbpf: map 'process_info': created successfully, fd=10"
level=debug name=parca-agent ts=2023-06-16T05:02:06.052569752Z caller=cpu.go:233 msg="libbpf: map 'programs': created successfully, fd=11"
level=debug name=parca-agent ts=2023-06-16T05:02:06.052640035Z caller=cpu.go:233 msg="libbpf: map 'debug_pids': created successfully, fd=12"
level=debug name=parca-agent ts=2023-06-16T05:02:06.069764652Z caller=cpu.go:233 msg="libbpf: map 'stack_traces': created successfully, fd=13"
level=debug name=parca-agent ts=2023-06-16T05:02:06.087966004Z caller=cpu.go:233 msg="libbpf: map 'dwarf_stack_traces': created successfully, fd=14"
level=debug name=parca-agent ts=2023-06-16T05:02:06.088422834Z caller=cpu.go:233 msg="libbpf: map 'stack_counts': created successfully, fd=15"
level=debug name=parca-agent ts=2023-06-16T05:02:06.090076293Z caller=cpu.go:233 msg="libbpf: map 'unwind_info_chunks': created successfully, fd=16"
level=debug name=parca-agent ts=2023-06-16T05:02:06.09011322Z caller=cpu.go:233 msg="libbpf: map 'percpu_stats': created successfully, fd=17"
level=debug name=parca-agent ts=2023-06-16T05:02:06.090146836Z caller=cpu.go:233 msg="libbpf: map 'events': created successfully, fd=18"
[...]

PR #2063, opened 2023-09-25
docs: update kernel minimum requirement

dreamerlzl:
Related: #1769

According to [this](https://github.com/iovisor/bcc/blob/master/docs/kernel-versions.md), 5.4 only introduces `BPF_MAP_TYPE_DEVMAP_HASH` for eBPF. Any explanation for why requiring 5.4 instead of the original 4.18+(e.g. Almalinux 8.7 would be excluded)?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,7 @@ Parca-agent follows [CNCF Code of Conduct](https://github.com/cncf/foundation/bl
 <!-- TODO: Add code of conduct info-->
 # Prerequisites
 
-- Linux Kernel version 4.18+
+- Linux Kernel version 5.4+
 - A source of targets to discover from: Kubernetes or systemd.
 
 Install the following dependencies (Instructions are linked for each dependency).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,7 @@ Parca-agent follows [CNCF Code of Conduct](https://github.com/cncf/foundation/bl
 <!-- TODO: Add code of conduct info-->
 # Prerequisites
 
-- Linux Kernel version 5.4+
+- Linux Kernel version 5.3+ with BTF
 - A source of targets to discover from: Kubernetes or systemd.
 
 Install the following dependencies (Instructions are linked for each dependency).
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ The collected data can be viewed locally via HTTP endpoints and then be configur
 
 ## Requirements
 
-* Linux Kernel version 5.4+
+* Linux Kernel version 5.3+ with BTF
 
 ## Quickstart
 
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ The collected data can be viewed locally via HTTP endpoints and then be configur
 
 ## Requirements
 
-* Linux Kernel version 4.18+
+* Linux Kernel version 5.4+
 
 ## Quickstart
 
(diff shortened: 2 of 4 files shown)


javierhonduco:
The introduction of `BPF_MAP_TYPE_DEVMAP_HASH` is unrelated to our requirements. We mainly have two constraints here:
- We require BTF support in the kernel. This was added in `4.18`, but wasn't enabled by most distros until later. We could download the needed BTF files on demand, and in fact we started this feature in https://github.com/parca-dev/parca-agent/pull/1131, but we have higher priority items and it's not likely going to be worked on in the near future;
- We need reasonably modern kernels that are ok with some BPF features we use, such as bounded loops, introduced in `5.3`;

Theoretically, the Agent should run in 5.3 or greater, but we only guarantee that the agent will work in the kernels we test on. Any kernel version in between the supported ones should work, minus bugs in the kernel.

We've done some experiments to allow the Agent to work in `4.x` (we require tail-calls, released in 4.2, and there were some fixes around 4.10 that we need, too) but this is not a priority for us and it can introduce a performance regression for all kernels (1M instructions were only added in `5.2`) , including modern ones that have features that allow our programs to run more efficiently.

All these versions have been pulled from https://github.com/iovisor/bcc/blob/master/docs/kernel-versions.md

CONTRIBUTING.md
```diff
@@ -14,7 +14,7 @@ Parca-agent follows [CNCF Code of Conduct](https://github.com/cncf/foundation/bl
 <!-- TODO: Add code of conduct info-->
 # Prerequisites
 
-- Linux Kernel version 4.18+
+- Linux Kernel version 5.4+
```
javierhonduco, line 17:
```suggestion
- Linux Kernel version 5.3+ with BTF
```

README.md
```diff
@@ -11,7 +11,7 @@ The collected data can be viewed locally via HTTP endpoints and then be configur
 
 ## Requirements
 
-* Linux Kernel version 4.18+
+* Linux Kernel version 5.4+
```
javierhonduco, line 14: Same as above

docs/design.md
```diff
@@ -4,7 +4,7 @@ Parca Agent implements a sampling profiler, to sample stack traces [19 times per
...
+Parca Agent uses BPF CO-RE (Compile Once – Run Everywhere) using [libbpf](https://github.com/libbpf/libbpf), pre-compiles all BPF programs, and statically embeds them in the target binary, from where it is loaded via libbpf when used. This means that Parca Agent does not need to compile the BPF program at startup or runtime like when using [bcc-tools](https://github.com/iovisor/bcc/tree/master/tools), meaning no Clang & LLVM, nor kernel headers need to be installed on the host. The only requirement is a [BTF](https://www.kernel.org/doc/html/latest/bpf/btf.html)-capable Kernel (Linux Kernel 5.4+).
```
javierhonduco, line 7: Same

docs/security.md
```diff
@@ -13,7 +13,7 @@ Parca Agent binaries and container image build processes have been specifically
...
+Parca Agent uses BPF CO-RE (Compile Once – Run Everywhere) using [libbpf](https://github.com/libbpf/libbpf), and pre-compiles all BPF programs, and statically embeds them in the target binary, from where it is loaded via libbpf when used. This means that Parca Agent does not need to compile the BPF program at startup or runtime like when using [bcc-tools](https://github.com/iovisor/bcc/tree/master/tools), meaning no Clang & LLVM, nor kernel headers need to be installed on the host. The only requirement is a [BTF](https://www.kernel.org/doc/html/latest/bpf/btf.html) capable Kernel (Linux Kernel 5.4+).
```
javierhonduco, line 16: Same as above

dreamerlzl:
Thank you for the explanation. Updated.

javierhonduco (approved):
LGTM, thanks for the change!

PR #2066, opened 2023-09-26
bpf: Fix task identifier usage

javierhonduco:
The name of PID (process ID) and TID (thread ID) in the kernel vs. in userspace is different, which is a bit confusing [0].

We tried to be careful here, but in some cases we used the "per thread ID" (using this name for clarity's sake) when we should have used the "per process ID". While I don't believe this caused any correctness issues, it will add overhead for multi threaded applications. For example, in `process_info` we keep a mapping of "per process ID" to the process information, which among other things, it contains the executable mappings of the process, so we can find the current object we are executing. Having this "per thread ID" is redundant, as every sibling thread has the same memory mappings as they share the same address space.

With these changes we will reduce CPU usage as we have to do less work and we can keep more info in the BPF maps, as we will use them more efficiently.

Running https://github.com/parca-dev/testdata/pull/21: **Before**

```
$ ppstree -p `pidof basic-cpp-multithreaded-no-fp`
basic-cpp-multi(146858)─┬─{basic-cpp-multi}(146859)
                        └─{basic-cpp-multi}(146860)
```

```
$ sudo bpftool map dump name process_info | grep "key"
        "key": 146858,
        "key": 146859,
        "key": 146860,
```

**After**

```
$ sudo bpftool map dump name process_info | grep "key"
        "key": 146858,
```

Test Plan
=========

Before: https://pprof.me/2de82f2
After: https://pprof.me/4b7b713

+ integration tests + visual inspection of various profiles.

[0]: https://man7.org/linux/man-pages/man2/getpid.2.html (notes section)

### DIFF
--- a/bpf/cpu/cpu.bpf.c
+++ b/bpf/cpu/cpu.bpf.c
@@ -209,7 +209,7 @@ typedef struct {
 
 /*================================ MAPS =====================================*/
 
-BPF_HASH(debug_pids, int, u8, 1); // Table size will be updated in userspace.
+BPF_HASH(debug_threads_ids, int, u8, 1); // Table size will be updated in userspace.
 BPF_HASH(process_info, int, process_info_t, MAX_PROCESSES);
 
 BPF_STACK_TRACE(stack_traces, MAX_STACK_TRACES_ENTRIES);
@@ -405,16 +405,16 @@ static u64 find_offset_for_pc(stack_unwind_table_t *table, u64 pc, u64 left, u64
 
 // Finds whether a process should be unwound using the unwind
 // tables.
-static __always_inline bool has_unwind_information(pid_t pid) {
-  process_info_t *proc_info = bpf_map_lookup_elem(&process_info, &pid);
+static __always_inline bool has_unwind_information(pid_t per_process_id) {
+  process_info_t *proc_info = bpf_map_lookup_elem(&process_info, &per_process_id);
   if (proc_info) {
     return true;
   }
   return false;
 }
 
-static __always_inline bool is_debug_enabled_for_pid(int pid) {
-  void *val = bpf_map_lookup_elem(&debug_pids, &pid);
+static __always_inline bool is_debug_enabled_for_thread(int per_thread_id) {
+  void *val = bpf_map_lookup_elem(&debug_threads_ids, &per_thread_id);
   if (val) {
     return true;
   }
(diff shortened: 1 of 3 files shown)


bpf/cpu/cpu.bpf.c
```diff
@@ -622,22 +622,12 @@ static __always_inline bool has_fp(u64 current_fp) {
...
-  // and the thread group ID in the upper 32 bits
-  // (what user space often thinks of as the PID).
-
-  int user_pid = pid_tgid >> 32;
-  int user_tgid = pid_tgid;
-  stack_key->pid = user_pid;
-  stack_key->tgid = user_tgid;
+
+  int per_process_id = pid_tgid >> 32;
+  int per_thread_id = pid_tgid;
```
javierhonduco, line 627: I find tgid and pid confusing because they are context dependent, these names mean what it says on the tin: any thread in a process should return the same `per_process_id` while each thread should have a unique `per_thread_id`

brancz (approved):

bpf/cpu/cpu.bpf.c
kakkoyun: Let's have this as an inline code comment. WDYT?

kakkoyun (commented):
I like the code identifier changes. It's more clear now. However, from a UX perspective, I'd rather prefer to specify the process ID and gather all the profiles from all the threads and subprocesses.

This could be handled with additional user-space program changes. WDYT?

javierhonduco:
> I'd rather prefer to specify the process ID and gather all the profiles from all the threads and subprocesses.

Not sure if I understand this. This logic should remain unchanged, we are still collecting all the profiles from every thread (see test plan) but we won't have duplicated per-thread information, when it's exactly the same data

javierhonduco:
Kernel tests for 5.4 are OOMing, taking a look, but there hasn't been any changes to the memory we use I think

kakkoyun:
> > I'd rather prefer to specify the process ID and gather all the profiles from all the threads and subprocesses.
> 
> Not sure if I understand this. This logic should remain unchanged, we are still collecting all the profiles from every thread (see test plan) but we won't have duplicated per-thread information, when it's exactly the same data

Maybe I'm missing something but currently, we put process IDs in the renamed map `debug_pids` no? And with these changes, we expect thread IDs to be specified with that flag.

bpf/cpu/cpu.bpf.c
kakkoyun: We encountered a few misunderstandings around this piece of the code. I'm in for anything that helps us to prevent future mistakes. Unfortunately, I don't have any valuable suggestions besides inline comments.

javierhonduco:
> Maybe I'm missing something but currently, we put process IDs in the renamed map debug_pids no? And with these changes, we expect thread IDs to be specified with that flag.

This is super confusing, so let me go step by step to ensure we are on the same page 😄 

The `bpf_get_current_pid_tgid()` helper returns `current_task->tgid << 32 | current_task->pid`, note that tgid and pid appear inverted in the name ⚠️ (https://man7.org/linux/man-pages/man7/bpf-helpers.7.html), so retrieving these values can be done with:
```c
int tgid = bpf_get_current_pid_tgid() >> 32 
```

```c
int pid = bpf_get_current_pid_tgid();
```

Now, the `tgid` in the kernel is the process ID for the first thread in a process. The `pid` is the thread ID, which will be unique for every thread in the process (and in the box).

This is the second confusing bit ⚠️  because in userspace we refer to everything as "PIDs" (tgid in kernel parlance) but in reality we typically mean thread ID (what the kernel calls pid).

Cool, so let's talk userspace now, how do we want to filter the process with `--debug-process-names`? Thread or process? Right now we filter by threads:

https://github.com/parca-dev/parca-agent/blob/fe5823a4f2b59244563a61ecae46bc801c2904e7/pkg/profiler/cpu/cpu.go#L827-L843

Note that procfs is accessed by thread ID `/proc/<thread_id>/` *but*... yes you've guessed it (⚠️) there is one more confusing thing here, which is that listing `/proc/` only shows processes 😭 and totally skips threads. That's why we need to check threads for every process! 


So, if we want to filter by process, we can remove the code to fish out the thread IDs, and we can change the filter in BPF to use process IDs rather than Threads. 

Personally, I think that filtering by thread name is quite convenient

bpf/cpu/cpu.bpf.c
javierhonduco: I don't mind adding the comments but I just think it will be difficult reading the code and then we need to ensure that they will be kept in sync in 2-3 different places

kakkoyun:
This will keep happening because this is still confusing. 
 I think we should unify the naming on the user-space naming.

Here we put PIDs in the recently renamed map: `BPF_HASH(debug_threads_ids, int, u8, 1)`

https://github.com/parca-dev/parca-agent/blob/fe5823a4f2b59244563a61ecae46bc801c2904e7/pkg/profiler/cpu/cpu.go#L847-L877

If this is sematically correct, the naming does not match.

Hence my confusion.

javierhonduco:
Let's try to improve the namings and explanations in follow-ups as much as we can 😰

Issue #1675, opened 2023-05-19
parca-agent triggers kernel bug because it calls bpf_probe_read_user() in the perf_event IRQ


luisgerhorst:
**Describe the bug**

Unfortunately, on some systems parca-agent seems to trigger a rare upstream kernel BUG because it calls `bpf_probe_read_user()` inside the perf_event IRQ. This is because `bpf_probe_read_user()` will call [`copy_from_user_nofault > access_ok > ... > find_vmap_area`](https://elixir.bootlin.com/linux/latest/source/mm/vmalloc.c#L1861) with some kernel configs (i.e., `CONFIG_HARDENED_USERCOPY`) which will attempt to acquire `vmap_area_lock`. If the interrupt occurred while the lock is held (e.g., during [`alloc_vmap_area()`](https://elixir.bootlin.com/linux/latest/source/mm/vmalloc.c#L1634) in the `clone()` syscall) `find_vmap_area()` will never return. This causes the lock held by `clone()` to never be released and any other CPU attempting to acquire it is locked up in an infinite loop. Ultimately, this happens on all CPUs and the whole machine is locked up.

**To Reproduce**

Start a machine using the affected upstream kernel code (tested with v6.1 but I believe the bug is also present in most other kernels). To reproduce it, you can for example use an AWS EC2 `c6a.large` (64 vCPUs) instance with the AMI `al2023-ami-2023.0.20230503.0-kernel-6.1-x86_64`. Having more CPUs allows the bug to be triggered more quickly.

```
$ curl -sL https://github.com/parca-dev/parca-agent/releases/download/v0.19.0/parca-agent_0.19.0_`uname -s`_`uname -m`.tar.gz | tar xvfz -
$ sudo ./parca-agent --node=test --remote-store-address=localhost:7070 --remote-store-insecure
```

To trigger the bug quickly, execute some code that will also use `vmap_area_lock`. For example, the `clone()` syscall:

```
$ while true ; do
ls -al > /dev/null # do not use true which is a shell builtin
done
```

Within 10 minutes, the CPU soft lockup messages should appear on the serial console.

**Expected behavior**

The machine is not locked up. BPF should not be able to lock up the machine but because of the kernel bug this happens anyway.

**Logs**

Here's an annotated log from the serial console. Other traces are also printed (from the other CPUs attempting to acquire the lock), however, this is the root cause I believe:

```
[253905.544838] Sending NMI from CPU 27 to CPUs 55:
[...]

PR #2070, opened 2023-09-27
Add check for known kernel issues

javierhonduco:
Releases >=5.19 && <6.1 have a pretty bad kernel bug that can result in whole sytem lock ups that can only be fixed with a reboot (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1033398).

The fix got backported to -stable
(https://www.spinics.net/lists/stable/msg662452.html, and https://www.spinics.net/lists/stable/msg662218.html for 6.1 and 6.3 respectively).

Let's not run the Agent in these kernels, but provide a flag to bypass this check. Note that running a buggy kernel can result in your machine going down.

Related issue: https://github.com/parca-dev/parca-agent/issues/1675

Test Plan
=========

Tested locally + added unit tests

### DIFF
--- a/cmd/parca-agent/main.go
+++ b/cmd/parca-agent/main.go
@@ -237,6 +237,8 @@ type FlagsHidden struct {
 	EnableRubyUnwinding   bool `default:"false" help:"Enable Ruby unwinding."   hidden:""`
 
 	ForcePanic bool `default:"false" help:"Panics the agent in a goroutine to test that telemetry works." hidden:""`
+
+	IgnoreUnsafeKernelVersion bool `default:"false" help:"Forces runs in kernels with known issues. This might freeze your system or cause other issues." hidden:""`
 }
 
 type FlagsBPF struct {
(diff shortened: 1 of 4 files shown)


cmd/parca-agent/main.go
```diff
@@ -562,6 +564,11 @@ func run(logger log.Logger, reg *prometheus.Registry, flags flags) error {
 		return errors.New("the BPF events buffer is too small, should be at least 32 pages")
 	}
 
+	release, err := kernel.GetRelease()
+	if err == nil && kernel.HasKnownBugs(release) && !flags.Hidden.IgnoreUnsafeKernelVersion {
+		return errors.New("this kernel version is known to cause issues, such as freezing your system. This can be bypassed with --ignore-unsafe-kernel-version but bad things can happen")
```
javierhonduco, line 569: Add https://github.com/parca-dev/parca-agent/discussions/2071

kakkoyun (approved):

brancz (approved):

PR #2080, opened 2023-09-29
Do not produce bad synthetic rows

javierhonduco:
During DWARF unwinding, we should first find the function where a given PC is included, and then find the actual instruction boundary. In our case, we don't represent functions and instruction boundaries separately, but we need a way to figure out if a PC is between two functions, but not included in their PC ranges. The reason for this is that we should stop unwinding at that point, as not finding unwind info for a PC indicates this.

In order to solve this in our compact unwind table representation, we were appending a synthetic row after adding every unwind row mapping every instruction from a function. This works when the next unwind row from the following function's first instruction differs from the current function's last instruction, but we were not handling this case.

In practice, if we are binary searching over an unwind table with duplicate PCs, if we land on the synthetic row first, we will stop unwinding, as we consider that there is no unwind information, even if that's wrong (the right row will be the next one).

In the unwinder we would find this special marker
https://github.com/parca-dev/parca-agent/blob/bdef22c2b7aa544cb6c7dcd59eb9368e986d96e1/bpf/cpu/cpu.bpf.c#L866 continue here
https://github.com/parca-dev/parca-agent/blob/bdef22c2b7aa544cb6c7dcd59eb9368e986d96e1/bpf/cpu/cpu.bpf.c#L1026 take this branch
https://github.com/parca-dev/parca-agent/blob/bdef22c2b7aa544cb6c7dcd59eb9368e986d96e1/bpf/cpu/cpu.bpf.c#L1044 and fail here
https://github.com/parca-dev/parca-agent/blob/bdef22c2b7aa544cb6c7dcd59eb9368e986d96e1/bpf/cpu/cpu.bpf.c#L1062 if we aren't dealing with a JIT compiler.

We can verify that's actually happening in production in one of our customer's deployment (see `pc_not_covered`):

```
parca_agent_native_unwinder_samples_total{unwinder="dwarf"} 1.03295398e+08
parca_agent_native_unwinder_error_total{reason="pc_not_covered"} 384695
parca_agent_native_unwinder_error_total{reason="catchall"} 189958
parca_agent_native_unwinder_error_total{reason="truncated"} 2716
parca_agent_native_unwinder_error_total{reason="unsupported_cfa_register"} 18312
parca_agent_native_unwinder_error_total{reason="unsupported_expression"} 4775

parca_agent_native_unwinder_error_total{reason="pc_not_covered_jit"} 0
parca_agent_native_unwinder_error_total{reason="should_never_happen"} 0
parca_agent_native_unwinder_error_total{reason="frame_pointer_action"} 0
parca_agent_native_unwinder_error_total{reason="jit_mixed_mode_disabled"} 0
parca_agent_native_unwinder_error_total{reason="jit_unupdated_mapping"} 0
parca_agent_native_unwinder_error_total{reason="jit_unwnding_machinery"} 0
```

Test Plan
=========

- Added snapshot tests yesterday, this will help us review the changes;
- Wrote some unit tests to ensure that PCs can't be repeated;
- Ran this for a while with no issues;
- Once we release this we should follow up with this customer to ensure that this error rate dropped;

### DIFF
--- a/pkg/stack/unwind/compact_unwind_table.go
+++ b/pkg/stack/unwind/compact_unwind_table.go
@@ -112,7 +112,21 @@ func (t CompactUnwindTable) Swap(i, j int)      { t[i], t[j] = t[j], t[i] }
 func BuildCompactUnwindTable(fdes frame.FrameDescriptionEntries, arch elf.Machine) (CompactUnwindTable, error) {
 	table := make(CompactUnwindTable, 0, 4*len(fdes)) // heuristic: we expect each function to have ~4 unwind entries.
 	context := frame.NewContext()
+	lastFunctionPc := uint64(0)
 	for _, fde := range fdes {
+		// Add a synthetic row at the end of the function but only
+		// if there's a gap between functions. Adding it at the end
+		// of every function can result in duplicated unwind rows for
+		// the same PC. This would not be correct a it can result in
+		// stopping the unwinding earlier than we should and that stack
+		// will be dropped.
+		if lastFunctionPc != 0 && fde.Begin() != lastFunctionPc {
+			table = append(table, CompactUnwindTableRow{
+				pc:      lastFunctionPc,
+				cfaType: uint8(cfaTypeEndFdeMarker),
+			})
+		}
+
 		frameContext := frame.ExecuteDwarfProgram(fde, context)
 		for insCtx := frameContext.Next(); frameContext.HasNext(); insCtx = frameContext.Next() {
 			row := unwindTableRow(insCtx)
@@ -122,12 +136,16 @@ func BuildCompactUnwindTable(fdes frame.FrameDescriptionEntries, arch elf.Machin
 			}
 			table = append(table, compactRow)
 		}
-		// Add a synthetic row for the end of the function.
-		table = append(table, CompactUnwindTableRow{
-			pc:      fde.End(),
-			cfaType: uint8(cfaTypeEndFdeMarker),
-		})
+
+		lastFunctionPc = fde.End()
 	}
+	// Add a synthetic row at the end of the unwind table. It is fine
+	// if this unwind table's last PC is equal to the next unwind table's first
+	// PC as we won't cross this boundary while binary searching.
+	table = append(table, CompactUnwindTableRow{
+		pc:      lastFunctionPc,
+		cfaType: uint8(cfaTypeEndFdeMarker),
+	})
 	return table, nil
 }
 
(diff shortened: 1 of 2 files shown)


javierhonduco:
This PR depends on the testdata changes @ https://github.com/parca-dev/testdata/pull/23

pkg/stack/unwind/compact_unwind_table.go
```diff
@@ -150,15 +168,6 @@ func rowToCompactRow(row *UnwindTableRow, arch elf.Machine) (CompactUnwindTableR
 		}
 
 		cfaOffset = int16(row.CFA.Offset)
-		/*if row.CFA.Reg == frame.X86_64FramePointer && arch == elf.EM_X86_64 {
```
javierhonduco, line 153: All these were left over commented code from the arm64 implementation

pkg/stack/unwind/compact_unwind_table.go
```diff
@@ -150,15 +168,6 @@ func rowToCompactRow(row *UnwindTableRow, arch elf.Machine) (CompactUnwindTableR
...
-			cfaType = uint8(cfaTypeRbp)
-		} else if row.CFA.Reg == frame.X86_64StackPointer && arch == elf.EM_X86_64 {
-			cfaType = uint8(cfaTypeRsp)
-		} else if row.CFA.Reg == frame.Arm64FramePointer && arch == elf.EM_AARCH64 {
-			cfaType = uint8(cfaTypeFp)
-		} else if row.CFA.Reg == frame.Arm64StackPointer && arch == elf.EM_AARCH64 {
-			cfaType = uint8(cfaTypeSp) // TODO(sylfrena): Reuse cfaTypeRsp
```
javierhonduco, line 160: These TODOs have already been tackled

pkg/stack/unwind/compact_unwind_table.go
```diff
@@ -170,28 +179,12 @@ func rowToCompactRow(row *UnwindTableRow, arch elf.Machine) (CompactUnwindTableR
...
 		rbpOffset = int16(row.RBP.Offset)
-		// curious that the following condition doesn't satisfy. it should.
-		// On further investigation, it doesn't because only Offset Rule is applied, and register value is x0, not x29
-		// Ideally this whole thing should work with just rbpType = uint8(reusedrbp/fpRuleOffset)
-		// TODO(sylfrena): Delete this part later
-		// if row.RBP.Reg == frame.Arm64FramePointer && arch == elf.EM_AARCH64 {
-		// fmt.Println("Rule fp Offset(arm64)")
-		//	rbpType = uint8(fpRuleOffset)
-		// }
-	// TODO(sylfrena): Do these conditions also cover Arm64 DWARF?
```
javierhonduco, line 186: They do! It's the same for both architectures. The differences aren't in the DWARF itself but in the register mapping, the lr, etc

javierhonduco:
Forgot to mention that as a side effect, we will save a bit of storage, which is always welcomed. For the whole of `testdata/`:

```
$ git show | grep -- "-pc" | wc -l
5620
```

pkg/stack/unwind/compact_unwind_table_test.go
```diff
@@ -410,6 +411,50 @@ func TestCompactUnwindTableArm64(t *testing.T) {
...
+func ensureSorted(t *testing.T, ut CompactUnwindTable) {
+	t.Helper()
+	lastPc := uint64(0)
+	for _, row := range ut {
+		require.Greater(t, row.Pc(), lastPc)
+		lastPc = row.Pc()
+	}
+}
+
+func ensureNoDuplicatedPc(t *testing.T, ut CompactUnwindTable) {
```
kakkoyun, line 423: nit
```suggestion
func ensureNoDuplicatedPC(t *testing.T, ut CompactUnwindTable) {
```

kakkoyun: This makes sure we don't have duplicates in a row, right?
So the input should be sorted if we want to make sure there are no duplicates globally, no?

kakkoyun (approved):

pkg/stack/unwind/compact_unwind_table_test.go
javierhonduco: We check these two invariants already, both that they have to be sorted by PC and that there should not be any duplicates. See TestIsSorted

kakkoyun (approved):

PR #2085, opened 2023-09-30
readme: Update language support roadmap

kakkoyun:
We already have support for Ruby and Python.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,9 +171,18 @@ Flags:
 ## Roadmap
 
 * Additional **zero-instrumentation** language support for just-in-time (JIT) compilers and dynamic languages (non-exhaustive list):
-  * JVM
+  * [JVM](https://www.parca.dev/docs/java-support)
   * Lua
   * Node.js
+
+
+  > [!IMPORTANT]
+  > All JIT compiled languages with frame pointer, and [Perf map](https://github.com/torvalds/linux/blob/master/tools/perf/Documentation/jit-interface.txt) or [jitdump](https://github.com/torvalds/linux/blob/master/tools/perf/Documentation/jitdump-specification.txt)
+  
+  
+  > [!NOTE]
+  > Check our documentation for further details: https://www.parca.dev/docs/parca-agent-language-support
+  
 * Additional types of profiles:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -170,20 +170,19 @@ Flags:
 
 ## Roadmap
 
-* Additional language support for just-in-time (JIT) compilers, and dynamic languages (non-exhaustive list):
-  * Ruby
-  * Node.js
-  * Python
+* Additional language support for just-in-time (JIT) compilers and dynamic languages (non-exhaustive list):
   * JVM
```
javierhonduco, line 174: I believe we support the JIT bits with frame pointers

README.md
```diff
@@ -170,20 +170,19 @@ Flags:
...
+  * Lua
+  * Node.js
```
javierhonduco, line 176: We technically support it too, no?

brancz: I would say we err on the side of "if it's possible at all we say we support it". So in that sense we do support JVM and nodejs. Not completely 100% zero-instrumentation yet, but with a tiny bit of effort people can already get it to work.

README.md
```diff
@@ -170,20 +170,28 @@ Flags:
...
+* Additional **zero-instrumentation** language support for just-in-time (JIT) compilers and dynamic languages (non-exhaustive list):
+  * [JVM](https://www.parca.dev/docs/java-support)
+  * Lua
   * Node.js
-  * Python
-  * JVM
+
+
+  > [!IMPORTANT]
+  > All JIT compiled languages with frame pointer, and [Perf map](https://github.com/torvalds/linux/blob/master/tools/perf/Documentation/jit-interface.txt) or [jitdump](https://github.com/torvalds/linux/blob/master/tools/perf/Documentation/jitdump-specification.txt)
```
javierhonduco, line 180: This sentence feels incomplete, but might be me!

javierhonduco: Also, I feel this is too "technical" if this makes sense, we should point users to more user-friendly documentation. Most people enabling the flags for JIT symbolization probably don't know about the format

PR #2100, opened 2023-10-04
Move debuginfo extraction to elfwriter package

brancz:
The debuginfo package has dependencies to Linux only things which makes it hard to test extraction on osx. This is primarily to make development easier.

### DIFF
--- a/pkg/elfwriter/extract.go
+++ b/pkg/elfwriter/extract.go
@@ -12,7 +12,7 @@
 // limitations under the License.
 //
 
-package debuginfo
+package elfwriter
 
 import (
 	"context"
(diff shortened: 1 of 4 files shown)


javierhonduco (approved):

pkg/elfwriter/extract.go
```diff
@@ -26,15 +26,8 @@ import (
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"go.opentelemetry.io/otel/trace"
-
-	"github.com/parca-dev/parca-agent/pkg/elfwriter"
 )
 
-type SeekReaderAt interface {
```
brancz, line 33: removed the duplicate here

brancz:
Ran linting locally and it passed.

PR #2163, opened 2023-10-19
Add gcr.io as additional container registry

metalmatze:
### Why?

We want to push container images to gcr.io too.

### What?

This adds an extra step to our CI to also push to gcr.io

### Test Plan

CI will show if it works. The Service Account itself was tested through impersonation.


brancz:
Not sure I understand, can you elaborate?

brancz:
Discussed on discord: The point was that since authentication is truly only done when pushing, we don't really know if this works until we merge it and it actually pushes.

PR #2177, opened 2023-10-26
Fix for 5.4 and older kernels running on arm64

javierhonduco:
Test Plan
=========

Tested on a `5.4.254-170.358.amzn2.aarch64` machine. Sumera is already working on the kernel tests that will catch these issues in the future.


brancz:
The container stuff is not a problem with this PR. Merging!

PR #2194, opened 2023-10-31
*: Optimize procinfo mapping updates

kakkoyun:
- Make sure process information updated only it is necessary
- Refresh process info mappings when needed

### Why?
We need to make sure process information is always up-to-date for correctness.
While aiming for correctness, we should make sure we prevent unnecessary work.
This PR makes sure that process information and interpreter information is up-to-date when mappings are changed.
Also, refactors the event handling a little bit to make it easier to reason about.

### <samp>🤖 Generated by Copilot at b920269</samp>

This pull request improves the performance and accuracy of the process profiling by adding caching and hashing mechanisms for the process mappings and interpreter information. It also refactors some code in the `cpu`, `bpfmaps`, and `unwind` packages to simplify and clarify the logic.

### Test Plan

* Test locally against interpreters

### DIFF
--- a/pkg/profiler/cpu/bpf/maps/maps.go
+++ b/pkg/profiler/cpu/bpf/maps/maps.go
@@ -28,7 +28,6 @@ import (
 	"syscall"
 	"unsafe"
 
-	"github.com/Masterminds/semver/v3"
 	libbpf "github.com/aquasecurity/libbpfgo"
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
@@ -184,6 +183,9 @@ type Maps struct {
 	pythonPIDToProcessInfo       *libbpf.BPFMap
 	pythonVersionSpecificOffsets *libbpf.BPFMap
 	pythonVersionToOffsetIndex   map[string]uint32
+
+	// Keeps track of synced process info and interpreter info.
+	syncedInterpreters map[int]runtime.Interpreter
 
 	unwindShards *libbpf.BPFMap
 	unwindTables *libbpf.BPFMap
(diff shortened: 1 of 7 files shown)


pkg/profiler/cpu/bpf/maps/maps.go
```diff
@@ -134,7 +133,7 @@ const (
 	compactUnwindRowSizeBytesArm64           = 16
 	minRoundsBeforeRedoingUnwindInfo         = 5
 	minRoundsBeforeRedoingProcessInformation = 5
-	maxCachedProcesses                       = 10_0000
+	maxCachedProcesses                       = 100_000
```
javierhonduco, line 136: What led to the 10x increase in the number of cached processes?

javierhonduco (commented):
Still reviewing it, but curious about the mem / CPU tradeoffs here -- did we see high resource consumption around these operations in the wild?

Something else I am thinking about is that we keep a mappings cache for executable objects for native unwinding with CFI. Should we perhaps unify all of this?

kakkoyun:
> Still reviewing it, but curious about the mem / CPU tradeoffs here -- did we see high resource consumption around these operations in the wild?
> 

Maybe "optimize" in the title is a little misleading. I wanted to make sure we don't do unnecessary updates. And make a little more clear to reason about the event handling. I will change the title.

The key part is to use the same idea of tracking changes in the mappings.

> 
> Something else I am thinking about is that we keep a mappings cache for executable objects for native unwinding with CFI. Should we perhaps unify all of this?

We should definitely do this. I started it but it was going to be too much change and it would touch multiple components,and we could break things so I reverted. We should only have one mapping implementation. Let's do it in another PR.

pkg/profiler/cpu/bpf/maps/maps.go
kakkoyun: If you look carefully it's not an increase. I have moved the underscore to proper location.

PR #2196, opened 2023-10-31
runtime/interpreter: Check the paths first for faster and cheaper detection

kakkoyun:

https://pprof.me/beb2cec956f42f3b78f38fc2a941ffdf
Ruby: https://pprof.me/4621cebc0f971f2e0f5a8bcfce78a90b
Python: https://pprof.me/14292dfb9a915878533098f53aa9b374

### DIFF
--- a/pkg/runtime/python/python.go
+++ b/pkg/runtime/python/python.go
@@ -22,6 +22,7 @@ import (
 	"path"
 	"regexp"
 	"strconv"
+	"strings"
 	"syscall"
 	"unsafe"
 
@@ -47,7 +48,7 @@ import (
 //	3.9:`Py_BytesMain`
 //	3.10:`Py_BytesMain`
 //	3.11:`Py_BytesMain`
-var pythonIdentifyingSymbols = [][]byte{
+var pythonExecutableIdentifyingSymbols = [][]byte{
 	[]byte("Py_Main"),
 	[]byte("_Py_UnixMain"),
 	[]byte("Py_BytesMain"),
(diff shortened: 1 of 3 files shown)


pkg/runtime/ruby/ruby.go
```diff
@@ -45,39 +44,61 @@ import (
...
+	if err != nil {
+		return false, fmt.Errorf("error reading process maps: %w", err)
 	}
 
-	if !ruby {
-		if ruby, err = runtime.IsSymbolNameInDynamicSymbols(ef, rubyIdentifyingSymbols); err != nil && !errors.Is(err, elf.ErrNoSymbols) {
-			return ruby, fmt.Errorf("search dynamic symbols: %w", err)
+	for _, mapping := range maps {
+		if isRubyLib(mapping.Pathname) {
+			// Let's make sure it's a ruby process by checking the ELF file.
```
javierhonduco, line 91: nit:
```suggestion
			// Let's make sure it's a Ruby process by checking the ELF file.
```

pkg/runtime/ruby/ruby.go
```diff
@@ -45,39 +44,61 @@ import (
...
 		return false, err
 	}
 
-	// Let's make sure it's a python process by checking the ELF file.
-	ef, err := elf.Open(absolutePath(proc, exe))
-	if err != nil {
-		return false, fmt.Errorf("open elf file: %w", err)
-	}
+	if isRubyBin(exe) {
+		// Let's make sure it's a ruby process by checking the ELF file.
```
javierhonduco, line 74: nit: 
```suggestion
		// Let's make sure it's a Ruby process by checking the ELF file.
```

pkg/runtime/python/python.go
```diff
@@ -60,35 +61,50 @@ const (
...
-	if python, err = runtime.IsSymbolNameInSymbols(ef, pythonIdentifyingSymbols); err != nil && !errors.Is(err, elf.ErrNoSymbols) {
-		return python, fmt.Errorf("search symbols: %w", err)
+	// If the executable is not a Python interpreter, let's check the memory mappings.
+	maps, err := proc.ProcMaps()
+	if err != nil {
+		return false, fmt.Errorf("error reading process maps: %w", err)
 	}
+	for _, mapping := range maps {
+		if isPythonLib(mapping.Pathname) {
+			// Let's make sure it's a python process by checking the ELF file.
```
javierhonduco, line 97: nit: 
```suggestion
			// Let's make sure it's a Python process by checking the ELF file.
```

pkg/runtime/python/python.go
```diff
@@ -60,35 +61,50 @@ const (
...
+	[]byte(pythonRuntimeSymbol),
+	[]byte(pythonThreadStateSymbol),
+}
+
 func absolutePath(proc procfs.Proc, p string) string {
 	return path.Join("/proc/", fmt.Sprintf("%d", proc.PID), "/root/", p)
 }
 
 func IsInterpreter(proc procfs.Proc) (bool, error) {
+	// First, let's check the executable`pathname since it's the cheapest and fastest.
```
javierhonduco, line 74: nit: 
```suggestion
	// First, let's check the executable's pathname since it's the cheapest and fastest.
```

javierhonduco (approved):
LGTM! Perhaps I missed it but I could not see Python or Ruby stacks in the profile, could you add them if they aren't present?

kakkoyun:
> LGTM! Perhaps I missed it but I could not see Python or Ruby stacks in the profile, could you add them if they aren't present?

The profile just from the agent, not the targets. 

I can add them as well.

javierhonduco:
We don't have integration tests for the interpreters, while I don't think this PR introduces any bugs, I think it would be a good idea to verify that everything still works as expected

kakkoyun:
> We don't have integration tests for the interpreters, while I don't think this PR introduces any bugs, I think it would be a good idea to verify that everything still works as expected

Yes, of course, I verified it. I'll attach a profile.

PR #2197, opened 2023-11-01
bpf/native: Don't add leading newline to logs

javierhonduco:
As it makes the output more obtuse. They were probably a typo.


kakkoyun (approved):
LGTM

PR #2199, opened 2023-11-02
unwinder/native: Unify native stacks

javierhonduco:
TL;DR: Unifying native stacks (fp, dwarf, kernel) decreases our BPF memory usage by ~70MB while reducing the amount of code. The effectiveness of stack storage and aggregation is improved and userspace might need to do less work thanks to it.

Context
=======

DWARF-derived unwinding stores its stack in a BPF hashmap where the stack hash (64 bits) is the key and the value is the stack formed by its addresses and length.

The kernel provided unwinder for user stacks with frame pointers and for kernel stacks was invoked via `bpf_get_stackid()` which only works with maps of type `BPF_MAP_TYPE_STACK_TRACE`.

This divergence led to our code needing two different maps for native stack storage, one for user code unwound without frame pointers `dwarf_stack_traces`, and another one for user code unwound with frame pointers, as well as kernel stacks `stack_traces`.

Having two maps is wasteful as this kernel memory is locked and can never be paged out. Tracing programs can't have lazily allocated maps either so ideally we would reduce memory usage.

Right now, each map roughly needs 66MB (excluding the overhead from internal bookeeping of the maps), 64000 entries * (8B key len + 8B len lenth + 8B size of frame * 127 frames).

Proposed solution
=================

This commit unifies all the native stack storage, the main changes are:

- Migrate user FP unwinder and kernel unwinder to `bpf_get_stack`, which received a byte buffer and returns the number of bytes of the unwound stack;
- Unify both stack storage maps into a single map, `stack_traces`;
- Remove custom stack readers for DWARF / FP / Kernel into a single helper as the stack layout is the same now;

As we now store all stacks in a single map with the same storage as before, now we ensure that the scratch stack is zeroed to ensure we have a better utilisation of the stack aggregation map. Before there could be garbage values after the length of the stack. This didn't affect correctness as we never read past it, but it resulted in worse aggregation as the stack hashes of two equal stacks with different garbage values would differ. This should also help make userspace faster in these cases as less data will have to be read / deleted from the stack agg maps.

Future work
===========

- We could also unify the interpreter stack here. This might be done in a different PR. A possibility would be to encode (frame_id << 32 | lineno) as a 64 bit value;
- The total stack storage is decreased by half. The effectiveness of it might be higher as we remove garbage values, but a way to tune the size of the maps would be ideal. This will be added in a forthcoming PR;

Test Plan
=========

- [x]: Integration tests must pass
- [x]: Kernel tests must pass
- [ ]: Run in Demo environment

### DIFF
--- a/pkg/profiler/cpu/bpf/maps/maps.go
+++ b/pkg/profiler/cpu/bpf/maps/maps.go
@@ -71,7 +71,6 @@ const (
 	PythonVersionSpecificOffsetMapName = "version_specific_offsets"
 
 	UnwindInfoChunksMapName = "unwind_info_chunks"
-	DWARFStackTracesMapName = "dwarf_stack_traces"
 	UnwindTablesMapName     = "unwind_tables"
 	ProcessInfoMapName      = "process_info"
 	ProgramsMapName         = "programs"
@@ -172,7 +171,6 @@ type Maps struct {
 	StackCounts            *libbpf.BPFMap
 	eventsCount            *libbpf.BPFMap
 	stackTraces            *libbpf.BPFMap
-	dwarfStackTraces       *libbpf.BPFMap
 	interpreterStackTraces *libbpf.BPFMap
 	symbolTable            *libbpf.BPFMap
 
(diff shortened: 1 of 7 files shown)


javierhonduco:
Note to self: just added some more commit, should squash when it's ready to merge

pkg/profiler/cpu/bpf/maps/maps.go
```diff
@@ -909,52 +906,27 @@ func (m *Maps) SetDebugPIDs(pids []int) error {
 	return nil
 }
 
-// ReadUserStack reads the user stack trace from the stacktraces ebpf map into the given buffer.
-func (m *Maps) ReadUserStack(userStackID int32, stack *bpfprograms.CombinedStack) error {
-	if userStackID == 0 {
+// ReadStack eads the walked stacktrace into the given buffer.
```
kakkoyun, line 909: s/eads/reads

kakkoyun (commented):
🥰 Looks good in a glance. 
I will have another look at it.

Issue #2202, opened 2023-11-04
Parca continous exporting profiles even when relabel config explicitly disallows it


isidentical:
**Describe the bug**
When parca-agent is configured to only keep profiles with a certain label, it still doesn't prevent ingesting mismatched profiles that lack that label. In our case we have a kubernetes cluster where parca-agent is deployed, and the agent is configured to only keep `app=$X` through a  `relabel_configs` entry. But I can still see profiling traces this being ingested in PolarSignals UI when I query `app!=$X`.

**To Reproduce**
This is our ConfigMap:
```
Name:         parca-config

Data
====
parca-agent.yaml:
----
relabel_configs:
- source_labels: [app]
  regex: fal-isolate-controller
  action: keep
```

And this is the Pod information:
```
[...]
    Mounts:
      /config from config (ro)
[...]
Volumes:
  config:
    Type:      ConfigMap (a volume populated by a ConfigMap)
    Name:      parca-config
    Optional:  false
```

**Expected behavior**
Not seeing any `app!=$X` on the UI.

**Screenshots** or **Profiles** (using https://pprof.me)
![image](https://github.com/parca-dev/parca-agent/assets/47358913/7afc0fea-a2a1-4a13-ade5-7bf77afa076c)

**Logs**

Don't see any errors/unusual warnings in the logs.

**Software (please complete the following information):**
 - Parca Agent Version: `ghcr.io/parca-dev/parca-agent:v0.26.0`
 - Parca Server Version (if applicable): PolarSignals

**Workload (please complete the following information):**
- Runtime (if applicable): Python
- Compiler (if applicable): -

**Environment (please complete the following information):**
- Linux Distribution: Same as the `parca-agent` docker image
- Linux Version: 5.15.107+
- Arch: x86_64
- Kubernetes Version (if applicable): v1.25.10-gke.2700
- Container Runtime (if applicable): COS

**Additional context**
Context from discord: https://discord.com/channels/813669360513056790/913454681878048868/1170356803712843946

kakkoyun:
@isidentical Could you please share the label set on your target workload? 

Hovering over a sample should do the trick. e.g.

![CleanShot 2023-11-09 at 18 20 55](https://github.com/parca-dev/parca-agent/assets/536449/6ba80ffc-a281-4cb7-b839-0c2b119f9b18)

isidentical:
This is a positive match (app="fal-isolate-controller")
[...]

PR #2232, opened 2023-11-14
config: Refactor config loading

kakkoyun:
Add warning messages for falling back to prevent #2202 


cmd/parca-agent/main.go
```diff
@@ -555,7 +555,13 @@ func run(logger log.Logger, reg *prometheus.Registry, flags flags) error {
 		if err != nil {
 			return fmt.Errorf("failed to read config: %w", err)
 		}
-		cfg = cfgFile
+		if cfg == nil {
+			level.Warn(logger).Log("msg", "config file is empty, using default config")
+		} else {
+			cfg = cfgFile
```
javierhonduco, line 561: Did you mean to move this to the block above?

pkg/config/config.go
```diff
@@ -36,13 +35,17 @@ func (c Config) String() string {
...
+	if len(b) == 0 {
+		return nil, nil
+	}
+
+	// TODO: Sanitize the input labels if it becomes a problem for the user.
+	// https://github.com/prometheus/prometheus/blob/1bfb3ed062e99bd3c74e05d9ff9a7fa4e30bbe21/util/strutil/strconv.go#L51
 	cfg := &Config{}
 
-	dec := yaml.NewDecoder(bytes.NewBufferString(s))
-	dec.KnownFields(true)
```
javierhonduco, line 43: Why was this removed?

kakkoyun: We are using a different API and it doesn't have this method.

kakkoyun:
@javierhonduco should be cleaner now.

PR #2233, opened 2023-11-14
metadata: Add runtime labels for NodeJS

kakkoyun:

The version information set in here https://github.com/nodejs/node/blob/36ecf8ca656063060421e5b5440ac056f3a51e20/src/node_version.h so they end up in `.data` or `.rodata` sections.

Tested against multiple versions https://pprof.me/75e08cdd91547a303b82065b9281bf1e

CleanShot 2023-11-15 at 13 09 51
CleanShot 2023-11-15 at 13 09 41
CleanShot 2023-11-15 at 13 09 36

### DIFF
--- a/pkg/metadata/runtime.go
+++ b/pkg/metadata/runtime.go
@@ -1,0 +1,71 @@
+// Copyright 2022-2023 The Parca Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+//nolint:dupl
+package metadata
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/procfs"
+
+	"github.com/parca-dev/parca-agent/pkg/cache"
+	"github.com/parca-dev/parca-agent/pkg/runtime"
+	"github.com/parca-dev/parca-agent/pkg/runtime/vm"
+)
+
+func Runtime(procfs procfs.FS, reg prometheus.Registerer) Provider {
+	cache := cache.NewLRUCache[int, *runtime.Runtime](
+		prometheus.WrapRegistererWith(prometheus.Labels{"cache": "metadata_runtime"}, reg),
+		512,
+	)
+	return &StatelessProvider{"runtime", func(ctx context.Context, pid int) (model.LabelSet, error) {
+		if ctx.Err() != nil {
+			return nil, ctx.Err()
+		}
+
+		p, err := procfs.Proc(pid)
+		if err != nil {
+			return nil, fmt.Errorf("failed to instantiate procfs for PID %d: %w", pid, err)
+		}
+
+		if rt, ok := cache.Get(pid); ok {
+			if rt == nil {
+				return nil, nil
+			}
+			return model.LabelSet{
+				"runtime":         model.LabelValue(rt.Type),
+				"runtime_version": model.LabelValue(rt.Version.String()),
+			}, nil
+		}
+
+		rt, err := vm.Fetch(p)
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch interpreter info for PID %d: %w", pid, err)
+		}
+		if rt == nil {
+			cache.Add(pid, nil)
+			return nil, nil
+		}
+		cache.Add(pid, rt)
+
+		return model.LabelSet{
+			"runtime":         model.LabelValue(rt.Type),
+			"runtime_version": model.LabelValue(rt.Version.String()),
+		}, nil
+	}}
+}
--- a/pkg/runtime/nodejs/nodejs.go
+++ b/pkg/runtime/nodejs/nodejs.go
@@ -1,0 +1,240 @@
+// Copyright 2023 The Parca Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package nodejs
+
+import (
+	"debug/elf"
+	"errors"
+	"fmt"
+	"os"
+	"path"
+	"regexp"
+	"strconv"
+
+	"github.com/Masterminds/semver/v3"
+	"github.com/prometheus/procfs"
+
+	"github.com/parca-dev/parca-agent/pkg/runtime"
+)
+
+var nodejsIdentifyingSymbols = [][]byte{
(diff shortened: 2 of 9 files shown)


pkg/metadata/runtime.go
```diff
@@ -0,0 +1,71 @@
...
+	cache := cache.NewLRUCache[int, *runtime.Runtime](
+		prometheus.WrapRegistererWith(prometheus.Labels{"cache": "metadata_runtime"}, reg),
+		512,
+	)
+	return &StatelessProvider{"runtime", func(ctx context.Context, pid int) (model.LabelSet, error) {
+		if ctx.Err() != nil {
+			return nil, ctx.Err()
+		}
+
+		p, err := procfs.Proc(pid)
```
javierhonduco, line 41: Should we move this after line ~55 so we don't have to do the work in the "is in cache" path?

pkg/metadata/runtime.go
```diff
@@ -0,0 +1,71 @@
...
+
+		rt, err := vm.Fetch(p)
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch interpreter info for PID %d: %w", pid, err)
+		}
+		if rt == nil {
+			cache.Add(pid, nil)
+			return nil, nil
+		}
+		cache.Add(pid, rt)
```
javierhonduco, line 64: `rt` can be an error, caching errors, is this intended?

pkg/runtime/nodejs/nodejs.go
```diff
@@ -0,0 +1,240 @@
...
+	return isNodeJS, nil
+}
+
+func RuntimeInfo(proc procfs.Proc) (*runtime.Runtime, error) {
+	isNodeJS, err := IsRuntime(proc)
+	if err != nil {
+		return nil, fmt.Errorf("is runtime: %w", err)
+	}
+	if !isNodeJS {
+		return nil, nil //nolint:nilnil
```
javierhonduco, line 107: nit: could this result in a misuse of the API? I think most would expect a nil error to carry a non-nil result. Perhaps a custom error / return value would be worth it? (I assume that this is what the lint above tries to catch)

pkg/metadata/runtime.go
kakkoyun: `rt` can be nil but not an error. Why do you think so?

pkg/runtime/nodejs/nodejs.go
```diff
@@ -0,0 +1,240 @@
...
+				if isNodeJSLib(pathname) {
+					found = true
+					lib = pathname
+					break
+				}
+			}
+		}
+	}
+	if !found {
+		return rt, fmt.Errorf("library %q not found in process maps", exe)
```
javierhonduco, line 156: TIL `%q` 😎

pkg/runtime/nodejs/nodejs.go
```diff
@@ -0,0 +1,240 @@
...
+	// If it is found, execution should never reach here.
+	if lastError != nil {
+		return "", lastError
+	}
+	return "", errors.New("version not found")
+}
+
+const semVerRegex string = `v([0-9]+)(\.[0-9]+)(\.[0-9]+)` +
+	`(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?` +
+	`(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?`
```
javierhonduco, line 224: nit: Perhaps removing everything after `-` or `+` and trimming a leading `v` could be easier to understand and evolve? This is a major nit though, I don't mind the approach and you added awesome tests already!

pkg/runtime/runtime.go
```diff
@@ -39,7 +41,15 @@ func (it InterpreterType) String() string {
...
+
 	// The address of the main thread state for Python.
 	MainThreadAddress  uint64
 	InterpreterAddress uint64
 }
+
+type Type string
+
+type Runtime struct {
+	Type    Type
```
javierhonduco, line 53: nit: Maybe "name" is clearer

pkg/runtime/nodejs/nodejs.go
```diff
@@ -0,0 +1,240 @@
...
+	ef, err := elf.NewFile(f)
+	if err != nil {
+		return "", fmt.Errorf("new file: %w", err)
+	}
+	defer ef.Close()
+
+	var lastError error
+	for _, sec := range ef.Sections {
+		if sec.Name == ".data" || sec.Name == ".rodata" {
+			data, err := sec.Data()
```
javierhonduco, line 202: perf nit: This might allocate a new buffer, I don't expect it to be very large, but might be worth checking some node js executables

pkg/metadata/runtime.go
javierhonduco: You are right, should have been `nil`, not an error!

pkg/runtime/nodejs/nodejs.go
kakkoyun: This is how we designed the providers in the first place. We could of course always change. The idea is some providers are opt-in and wouldn't contribute any labels to the label set of the given process.

javierhonduco (approved):
LGTM!! Mostly left some nits and questions to understand things a bit better.

Another thing I wanted to ask is where can I see the values retrieved in the pprof profile, I couldn't spot anything 😢 .

One more thing I noticed is that the provided file, https://github.com/nodejs/node/blob/36ecf8ca656063060421e5b5440ac056f3a51e20/src/node_version.h will only run in the preprocessor, so the version inclusion in the binary must happen elsewhere. Could be good to know exactly where it's embedded in case this shows us some opportunity to do things differently.

pkg/runtime/nodejs/nodejs.go
kakkoyun: We can be more precise about matching, but we can't be loose. Removing `v` gives false matches (I've tested, and we get things like `127.0.0.1). We must be strict since we search the whole space of constant strings. I wish there were a more straightforward way. We could evolve the approach when we know more.

pkg/runtime/runtime.go
kakkoyun: I'm also planning to merge or embed this to the `Interpreter` type. I'll iterate on it in subsequent PRs.

pkg/runtime/nodejs/nodejs.go
kakkoyun: You are spot on I have already an optimization lined up to optimize this part. Stay tuned ;)

kakkoyun:
Thanks for the review ❤️ 

> Another thing I wanted to ask is where can I see the values retrieved in the pprof profile, I couldn't spot anything 😢 .

This is my bad. For whatever reason, I thought we had the labels in there. I'm gonna add a screenshot of the labels.

kakkoyun:
> Thanks for the review ❤️
> 
> > Another thing I wanted to ask is where can I see the values retrieved in the pprof profile, I couldn't spot anything 😢 .
> 
> This is my bad. For whatever reason, I thought we had the labels in there. I'm gonna add a screenshot of the labels.

Shots are the descriptions.

javierhonduco:
Seems to be defined here: https://github.com/nodejs/node/blob/ff879cd493988e0bd8b8d7a44e3c0bd166719a51/src/node_metadata.h#L107-L128 and set here https://github.com/nodejs/node/blob/ff879cd493988e0bd8b8d7a44e3c0bd166719a51/src/node_metadata.cc#L135C3-L135C13

kakkoyun:
> Just if you are curious, seems to be defined [here](https://github.com/nodejs/node/blob/ff879cd493988e0bd8b8d7a44e3c0bd166719a51/src/node_metadata.h#L107-L128 ) and set [here](https://github.com/nodejs/node/blob/ff879cd493988e0bd8b8d7a44e3c0bd166719a51/src/node_metadata.cc#L135C3-L135C13). So it should be stored in the `.rodata` section of the `node_metadata.cc` compilation unit

Amazing 🤩 Thanks for digging it out

pkg/runtime/nodejs/nodejs.go
kakkoyun: @javierhonduco https://github.com/parca-dev/parca-agent/pull/2254

PR #2236, opened 2023-11-15
go.*: Upgrade all the packages

kakkoyun:

Renovate is having some problems upgrading packages individually so, I'm giving a hand.


kakkoyun:
Braking container builds are unrelated.

PR #2243, opened 2023-11-15
CI: Add kerneltests for arm64 kernels

Sylfrena:
CI: Add kerneltests for arm64 kernel

Add a separate goreleaser file [.goreleaser-test-binaries.yml](https://github.com/parca-dev/parca-agent/pull/2243/files#diff-043df9ae8265bd7d86706b3d1e756723aba02d425f4747df49e4a42f8c8f1148) for generating `arm64` and `x86` test binaries in CI

Add a new GH action workflow for kerneltests in `.github/workflows/kerneltest.yml` for running kerneltests. Tests for both `arm64` and `x86_64` architectures are included.

As of now, only `v6.5.0-14 arm64` linux kernel is used for testing. The `arm64` image is fetched from the [parca-ci-kernels](https://github.com/parca-dev/parca-ci-kernels) repo. More kernel versions will be added in follow-up PRs as they are added to `parca-ci-kernels`.

Tests can be run locally by running the `./kerneltest/vmtest.sh` script from `parca-agent` root. 
(Both `qemu-system-x86_64` and `qemu-system-aarch64` are needed to run the tests).

image


kerneltest/vmtest.sh
```diff
@@ -18,56 +18,93 @@ set -o pipefail
...
     kernel_version=$1
-    cat <<EOT >"kerneltest/logs/vm_log_$kernel_version.txt"
+    arch=$2
+    cat <<EOT >"kerneltest/logs/vm-log-$kernel_version-$arch.txt"
 ============================================================
 - date: $(date)
-- git revision: $(git rev-parse HEAD)-$(git diff-index --quiet HEAD || echo dirty)
-- vm kernel: $kernel_version
+- vm kernel: $kernel_version $arch
 - qemu version: $(qemu-system-x86_64 --version | head -1)
```
javierhonduco, line 56: Reminder: change this to the right qemu

kerneltest/vmtest.sh
```diff
@@ -78,23 +115,63 @@ check_executable() {
...
-            vm_run "$kernel" "1.5G"
-        fi
+    arch_versions=("amd64", "arm64")
+
+    for arch in "${arch_versions[@]}"; do
+        for kernel in "${kernel_versions[@]}"; do
+            use_kernel "$kernel" "$arch"
+            # Ensure that the adaptive unwind shard mechanism
+            # works in memory constrained environments.
+            if [[ "$kernel" == "5.4" ]]; then
```
javierhonduco, line 155: As discussed on our 1:1, let's reduce the if/else complexity here and use the arch passed to `vm_run`

.github/workflows/container.yml
```diff
@@ -319,3 +330,51 @@ jobs:
...
+          # Copy amd64 test
+          cp ./dist/parca-agent-test-amd64_linux_amd64_v1/kerneltest/cpu.test ./kerneltest/amd64/cpu.test
+
+          # Copy arm64 test
+          cp ./dist/parca-agent-test-arm64_linux_arm64/kerneltest/cpu.test ./kerneltest/arm64/cpu.test
+
+          go install "github.com/florianl/bluebox@${BLUEBOX_VERSION}"
+          echo $PATH
+
+
```
Sylfrena, line 370: make both test binaries executable first

Sylfrena:
Check if all relevant `make` commands still run. Maybe fix in a follow-up PR or wait till the new Makefile changes land.

.github/workflows/container.yml
```diff
@@ -161,7 +163,8 @@ jobs:
           find pkg/contained/bpf/*/ -type f | grep -v "pid_namespace.bpf.o" | xargs -I{} bash -c "rm {}"
 
       - name: Run Goreleaser
-        run: goreleaser release --clean --skip-validate --skip-publish --snapshot --debug
+        run: |
+          goreleaser release --clean --skip-validate --skip-publish --snapshot --debug -f .goreleaser.yml
```
Sylfrena, line 167:
```suggestion
        run: goreleaser release --clean --skip-validate --skip-publish --snapshot --debug -f .goreleaser.yml
```

.github/workflows/kerneltest.yml
```diff
@@ -0,0 +1,237 @@
...
+        shell: bash
+        run: |
+          ls -lR dist
+
+      # TODO(sylfrena): add this step as a target in Makefile for local dev
+      - name: Build initramfs
+        run: |
+          # Copy amd64 test
+          cp ./dist/parca-agent-test-amd64_linux_amd64_v1/kerneltest/cpu.test ./kerneltest/amd64/cpu.test
+
```
Sylfrena, line 217: Instructions here ideally should be part of the `make initramfs` target in out Makefile but since we might be switching to Magefiles or another build system as part of https://github.com/parca-dev/parca-agent/pull/2368#issuecomment-1862823498 we can do this as follow up if we need to.

Makefile
```diff
@@ -236,13 +237,18 @@ test: $(DOCKER_BUILDER)
 	$(call docker_builder_make,$@)
 endif
 
+hello-profiler: build
+$(OUT_BIN_HELPRO): go/deps
+	$(GO_ENV) $(GO) build $(SANITIZERS) $(GO_BUILD_FLAGS) -o $@ ./cmd/hello-profiler
+
```
Sylfrena, line 243: leftover code from debugging, remove this

kerneltest/amd64/README.md
```diff
@@ -0,0 +1 @@
+The x86_64 test binaries from goreleaser will be written in this directory.
```
Sylfrena, line 1:
```suggestion
The x86_64 test binaries from goreleaser will be written into this directory.
```

kerneltest/arm64/README.md
```diff
@@ -0,0 +1 @@
+The arm64 test binaries from goreleaser will be written in this directory.
```
Sylfrena, line 1:
```suggestion
The arm64 test binaries from goreleaser will be written into this directory.
```

kerneltest/kernels/README.md
```diff
@@ -1 +1,4 @@
-The kernel images will be downloaded in this directory.
+Dowloaded kernel images used in kerneltests are stored in this folder.
```
Sylfrena, line 1:
```suggestion
Dowloaded kernel images used for kerneltests are stored in this folder.
```

kerneltest/kernels/README.md
```diff
@@ -1 +1,4 @@
...
+
+NOTE: Do not delete this file as GitHub doesn't track empty folders and this causes CI scripts
+to fail.
```
Sylfrena, line 4:
```suggestion
NOTE: Do not delete this file as GitHub doesn't track empty folders causing CI scripts to fail.
```

.github/workflows/container.yml
```diff
@@ -201,11 +204,24 @@ jobs:
...
+        uses: actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe # v4.1.0
+        with:
+          go-version-file: .go-version
+
       - uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
         with:
           name: parca-agent-dist-container
           path: goreleaser/dist
 
+      #- name: Setup tmate session
```
javierhonduco, line 217: Pls remove it :D

Makefile
```diff
@@ -236,13 +237,18 @@ test: $(DOCKER_BUILDER)
...
 cputest-static: build
 	$(GO_ENV) $(CGO_ENV) $(GO) test -v ./pkg/profiler/cpu -c $(GO_BUILD_FLAGS) --ldflags="$(CGO_EXTLDFLAGS)"
-	mv cpu.test kerneltest/
+	mv cpu.test kerneltest/arm64/
 
 initramfs: cputest-static
-	bluebox -e kerneltest/cpu.test
-	mv initramfs.cpio kerneltest
+	bluebox -e kerneltest/cpu.test -o amd64-initramfs.cpio
```
brancz, line 250: Do we even still need this make target with all the instructions in the github action for now? I would prefer if we have one source of truth, either we have a script that we call from the github action, make target(s), or everything in the github action. It's a bit all over the place right now.

brancz (commented):
Very nice!!

I think largely we want to do a bit of deduplication so we have only one way to do things, other than that this looks great!

(I think like you already started, I would opt for not doing anything in make for now since we're in the process of moving so a script and call that from the github action would be my preferred option right now, but not feeling strongly as long as we have one way to do things I'm happy)

Makefile
Sylfrena: I agree with you. The targets are largely still helpful for local dev.

The reason we have things slightly different in github actions is that CI and goreleaser do some weird things (like make binaries non-executable in this case). 

We ultimately do end up calling the `vmtest.sh` script in both cases, once we are past the goreleaser part. I'll still see if we can deduplicate some more. 

Since Kemal's planning to reimplement the whole thing I just wanted to hold off.

Sylfrena: Nuking the make targets for now, will try to write a script in a follow up PR that we can run both locally and in CI

brancz: sounds good

PR #2253, opened 2023-11-16
metadata: Separate label sets per runtime

kakkoyun:

### Why?
According to our users' feedback individual sets of labels per runtime easier to reason about. This PR converts "runtime"=XXX and "runtime_version" to "XXX"=true and "XXX_version" labels.

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 0096c23</samp>

Refactored the runtime and metadata packages to support multiple languages and simplify the code. Added metadata providers for Python and Ruby. Renamed and embedded structs and functions for clarity and consistency. Improved error handling and tracing.

CleanShot 2023-11-16 at 17 13 02
CleanShot 2023-11-16 at 16 51 06
CleanShot 2023-11-16 at 16 50 09

### DIFF
--- a/pkg/metadata/python.go
+++ b/pkg/metadata/python.go
@@ -24,18 +24,21 @@ import (
 	"github.com/prometheus/procfs"
 
 	"github.com/parca-dev/parca-agent/pkg/cache"
-	"github.com/parca-dev/parca-agent/pkg/runtime"
-	"github.com/parca-dev/parca-agent/pkg/runtime/interpreter"
+	"github.com/parca-dev/parca-agent/pkg/runtime/python"
 )
 
-func Interpreter(procfs procfs.FS, reg prometheus.Registerer) Provider {
-	cache := cache.NewLRUCache[int, *runtime.Interpreter](
-		prometheus.WrapRegistererWith(prometheus.Labels{"cache": "metadata_interpreter"}, reg),
-		512,
+func Python(reg prometheus.Registerer, procfs procfs.FS) Provider {
+	cache := cache.NewLRUCache[int, model.LabelSet](
+		prometheus.WrapRegistererWith(prometheus.Labels{"cache": "metadata_python"}, reg),
+		128,
 	)
-	return &StatelessProvider{"interpreter", func(ctx context.Context, pid int) (model.LabelSet, error) {
+	return &StatelessProvider{"python", func(ctx context.Context, pid int) (model.LabelSet, error) {
 		if ctx.Err() != nil {
 			return nil, ctx.Err()
+		}
+
+		if lset, ok := cache.Get(pid); ok {
+			return lset, nil
 		}
 
 		p, err := procfs.Proc(pid)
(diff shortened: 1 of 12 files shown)


pkg/metadata/python.go
```diff
@@ -0,0 +1,77 @@
...
+		lset := model.LabelSet{
+			"python": model.LabelValue(strconv.FormatBool(true)),
+		}
+
+		rt, err := python.RuntimeInfo(p)
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch interpreter info for PID %d: %w", pid, err)
+		}
+		if rt == nil {
+			cache.Add(pid, lset)
```
javierhonduco, line 67: Why not returning the lset here / setting `nil`? Next time this is called it will return it anyways as `cache` will have an entry for it I think. What kind of behaviour should we have?

pkg/metadata/ruby.go
```diff
@@ -0,0 +1,77 @@
...
+		lset := model.LabelSet{
+			"ruby": model.LabelValue(strconv.FormatBool(true)),
+		}
+
+		rt, err := ruby.RuntimeInfo(p)
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch interpreter info for PID %d: %w", pid, err)
+		}
+		if rt == nil {
+			cache.Add(pid, lset)
```
javierhonduco, line 67: Same as in python's

javierhonduco (approved):
Reviewed half of it so far, will continue later with the rest!

pkg/metadata/python.go
kakkoyun: Good catch 🌮

pkg/runtime/runtime.go
```diff
@@ -38,18 +38,18 @@ func (it InterpreterType) String() string {
 	}
 }
 
-type Interpreter struct {
-	Type    InterpreterType
+type RuntimeName string
+
+type Runtime struct {
+	Name    RuntimeName
```
javierhonduco, line 44: That's a significantly better name!!

pkg/runtime/ruby/ruby.go
```diff
@@ -152,15 +268,14 @@ func InterpreterInfo(proc procfs.Proc) (*runtime.Interpreter, error) {
 		rubyExecutable = path.Join("/proc/", strconv.Itoa(pid), "/root/", librubyPath)
 	}
 
-	// Read the Ruby version.
-	//
 	// PERF(javierhonduco): Using Go's ELF reader in the stdlib is very
```
javierhonduco, line 271: This code was moved above, no? Could the extracted function be used here?

kakkoyun: Not, exactly as in current shape. Some of the variables like `symbols` are reused. I don't want to refactor too much code in this PR. We could adapt a similar approach as in `runtime/python` but after some refinements in that code.

pkg/runtime/ruby/ruby.go
```diff
@@ -102,6 +102,122 @@ func IsInterpreter(proc procfs.Proc) (bool, error) {
...
+	versionString, err := versionFromSymbol(f)
+	if err == nil {
+		version, err := semver.NewVersion(versionString)
+		if err != nil {
+			return rt, fmt.Errorf("new version: %q: %w", versionString, err)
+		}
+		rt.Version = version
+		return rt, nil
+	}
+	if _, err := f.Seek(0, io.SeekStart); err != nil {
```
javierhonduco, line 174: The file is closed in line 163. I think this seek is not needed and will always fail

kakkoyun: In line 163, it's used with `defer` so the file is open here.
The `Seek` here, because of ```go
			address := symbol.Value

			_, err = f.Seek(int64(address), io.SeekStart)
			if err != nil {
				return "", err
			}

			_, err = f.Read(rubyVersionBuf)
			if err != nil {
				return "", err
			}
``` in `versionFromSymbol`.

**Comment by kakkoyun** [comment]
> Could you add Python, Ruby and Nodejs pprofs as part of the test?

I can attach the files but they wouldn't have the metadata labels in them, thus the screenshots.

**Comment by javierhonduco** [comment]
Merging this PR means that we'll have to fix all the ELF symbol reading inefficiencies before the next version as now every Python / Ruby process on the system will make us pay a higher price during metadata extraction, than before was disabled.

Just wanted to point this out, it's not that I am against this change, but we should fix this sooner rather than later

**Comment by kakkoyun** [comment]
> Merging this PR means that we'll have to fix all the ELF symbol reading inefficiencies before the next version as now every Python / Ruby process on the system will make us pay a higher price during metadata extraction, than before was disabled.
> 
> Just wanted to point this out, it's not that I am against this change, but we should fix this sooner rather than later

There is always room for improvement in reading the symbols. However, this PR doesn't actually change anything regarding runtime costs. These are mere code refactorings and renaming the label names. If we have any inefficiencies, they are already on the main. But we heavily cache the results, and the metadata reading happens only once. 

I'd be happy to address any prominent inefficiencies immediately before the next release after enabling the metadata extraction for Ruby and Python processes by default.

**Comment by javierhonduco** [comment]
The main change is that before, by default, there was no cost to these checks, now there is

```go
	interpreterUnwindingEnabled := flags.Hidden.EnablePythonUnwinding || flags.Hidden.EnableRubyUnwinding
	if interpreterUnwindingEnabled { # <- these lines
		providers = append(providers, metadata.Interpreter(pfs, reg)) 
	}
```

kakkoyun:
Let's see them in action. I don't anticipate too much of a cost but I might be wrong. The compiler metadata checks are way more expensive, for example.

PR #2254, opened 2023-11-16
runtime/nodejs: Optimize .data version scans

kakkoyun:
CleanShot 2023-11-20 at 15 14 44@2x
CleanShot 2023-11-20 at 15 14 38@2x


javierhonduco (approved):
LGTM, great stuff!

javierhonduco:
Given that we have no integration tests yet, could you run some tests manually to ensure we are not regressing?

kakkoyun:
> Given that we have no integration tests yet, could you run some tests manually to ensure we are not regressing? 

I have already done that. I will attach some screenshots with extracted labels.

PR #2280, opened 2023-11-21
runtime: Optimize FindSymbol

kakkoyun:
- Add tests and benchmarks for isSymbolNameInSection
- Optimize FindSymbol
- Integrate runtime/ruby with FindSymbol


pkg/runtime/testdata/libpython3.11.so.1.0
javierhonduco, line 1: As we've discussed let's move to `testdata/`

PR #2286, opened 2023-11-23
bpf: Remove line number from frame

javierhonduco:
Line numbers for both Ruby and Python don't work great yet or are not even implemented.

Once we re-add them, we will encode them as the upper 32 bits of the frame ID that forms a stacktrace.


kakkoyun (commented):
👍 

> Once we re-add them, we will encode them as the upper 32 bits of the frame ID that forms a stacktrace.

I think we should add them as part of this PR or very worst case before cutting the next release.

PR #2290, opened 2023-11-23
Reduce number of the reads from the interpreter symbol table

kakkoyun:

- Uses `BPF_MAP_TYPE_LRU_HASH` for the symbol table. To eliminate cleaning up the map.
> Both BPF_MAP_TYPE_LRU_HASH and BPF_MAP_TYPE_LRU_PERCPU_HASH were introduced in version 4.10

- Caches the symbol table in the user space.
- Only updates the table if it encounters a new symbol ID.

It should work even better in combination with #2286 

### Benchmark
**Before:** 

1500-2500 CGo calls pm

Metrics:
CleanShot 2023-11-23 at 21 25 50

 
Profile:
- https://pprof.me/aebf36cac18a2f6c1725aec810ea87bc

**After:**

200-300 CGo calls pm

Metrics: 
CleanShot 2023-11-23 at 21 47 13

Profile:
- [TBA](https://pprof.me/d44684fefc312522443fa1c1747a1ac1)

**Comparison**
CleanShot 2023-11-23 at 21 44 22

### DIFF
--- a/bpf/unwinders/pyperf.bpf.c
+++ b/bpf/unwinders/pyperf.bpf.c
@@ -20,7 +20,6 @@
 //   ╚═════════════════════════════════════════════════════════════════════════╝
 //
 const volatile bool verbose = false;
-const volatile int num_cpus = 200; // Hard-limit of 200 CPUs.
 
 //
 //   ╔═════════════════════════════════════════════════════════════════════════╗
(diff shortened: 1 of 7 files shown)


javierhonduco (approved):
Thanks for catching the type mismatch between ID generation and the stored value and for reusing the code to insert / get frames.

Overall LGTM! As we discussed, I think the new logic to decide when to refresh the whole symbol table could be improved, e.g. only reading the missing symbols (if there aren't that many that are missing) and only redoing the whole symbol table entirely if it gets too large, for example

PR #2307, opened 2023-11-28
*: Use optimized symtabs for jitdump and perfmaps

brancz:
This patch utilizes the optimized symbol table previously only used for ksyms to move the results of parsing perfmaps and jitdumps to external memory and memory map it.

Since the overwhelming majority of symbols are never used this allows saving a large amount of memory while still keeping those that are used often in cache, which is handled by the operating system using mmap.

### Why?

Holding the results of parsing jitdump/perfmaps in-memory results in unnecessarily high memory usage.

### What?

Move the result to external memory and memory map it.

### How?

Use the same optimized symbol table as we use for ksyms already.

### Test Plan

Only unit tests so far, I will test it some more with actual workloads once merged.

### DIFF
--- a/cmd/parca-agent/main.go
+++ b/cmd/parca-agent/main.go
@@ -26,6 +26,7 @@ import (
 	"net/url"
 	"os"
 	"os/exec"
+	"path/filepath"
 	goruntime "runtime"
 	runtimepprof "runtime/pprof"
 	"strconv"
@@ -927,6 +928,14 @@ func run(logger log.Logger, reg *prometheus.Registry, flags flags) error {
 		})
 	}
 
+	optimizedSymtabs := filepath.Join(flags.Debuginfo.TempDir, "optimized_symtabs")
+	if err := os.RemoveAll(optimizedSymtabs); err != nil {
+		level.Warn(logger).Log("msg", "failed to remove optimized symtabs directory", "err", err)
+	}
+	if err := os.MkdirAll(optimizedSymtabs, 0o644); err != nil {
+		return fmt.Errorf("failed to create optimized symtabs directory: %w", err)
+	}
+
 	profilers := []Profiler{
 		cpu.NewCPUProfiler(
 			log.With(logger, "component", "cpu_profiler"),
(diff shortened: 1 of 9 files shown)


cmd/parca-agent/main.go
```diff
@@ -927,6 +928,14 @@ func run(logger log.Logger, reg *prometheus.Registry, flags flags) error {
 		})
 	}
 
+	optimizedSymtabs := filepath.Join(flags.Debuginfo.TempDir, "optimized_symtabs")
+	if os.RemoveAll(optimizedSymtabs); err != nil {
```
javierhonduco, line 932: Did you mean?

```suggestion
	if err := os.RemoveAll(optimizedSymtabs); err != nil {
```

cmd/parca-agent/main.go
```diff
@@ -927,6 +928,14 @@ func run(logger log.Logger, reg *prometheus.Registry, flags flags) error {
...
+		level.Warn(logger).Log("msg", "failed to remove optimized symtabs directory", "err", err)
+	}
+	if os.MkdirAll(optimizedSymtabs, 0o755); err != nil {
```
javierhonduco, line 935: Same as above

cmd/parca-agent/main.go
brancz: oh wow, I didn't even expect this to compile

javierhonduco: (unrelated to the PR feedback) I prefer to avoid this pattern because I find it unreadable and it's too easy to make mistakes like this one. Perhaps a good opportunity for a new linter? 😄

javierhonduco:
Getting closer, some CI failures (`make lint` locally should report the same)

```
Error: test/integration/profiler_test.go:292:2: tmpDir declared and not used
Error: test/integration/profiler_test.go:298:10: not enough return values
	have (error)
	want (*"github.com/parca-dev/parca-agent/pkg/profiler/cpu".CPU, *prometheus.Registry, *objectfile.Pool) (typecheck)
```

brancz:
Yeah the problem was that I was doing all of this directly on OSX which prevents me from running some of the tests.

pkg/perf/dump.go
```diff
@@ -83,21 +87,53 @@ func ReadJITdump(logger log.Logger, fileName string) (Map, error) {
...
+func (p *JITDumpCache) path(pid int, fileName string) string {
+	return p.pathForKey(key(pid, fileName))
+}
+
+func (p *JITDumpCache) pathForKey(key string) string {
+	return filepath.Join(p.tmpDir, key)
+}
+
+func key(pid int, fileName string) string {
+	return fmt.Sprintf("/proc/%d/root%s", pid, fileName)
```
javierhonduco, line 130: Just in case the filename doesn't contain a trailing path, can we use `filepath.Join`?

javierhonduco (approved):
LGTM! Just a minor comment

cmd/parca-agent/main.go
javierhonduco: For future refrences, this wasn't caught in CI because golangci-lint excludes it https://github.com/golangci/golangci-lint/blob/855d443351417cfec74f1c832cab2c52ecedcb71/pkg/config/issues.go#L13-L14

PR #2310, opened 2023-11-29
profiler/cpu: Move metrics registration out of retry

brancz:
This previously had the potential to cause a duplicate metrics registration panic.


javierhonduco (approved):
Great stuff!

Issue #2311, opened 2023-11-29
Panic while attempting to profile Python


Garbett1:
**Describe the bug**
Parca-agent panics while attempting to profile a Python binary

**To Reproduce**
Can't replicate with a minimal example yet, but happens consistently with the specific binary.

**Expected behavior**
Ideally, getting the trace, but at least an error rather than a panic.

**Logs**
```panic: runtime error: index out of range [0] with length 0

goroutine 192 [running]:
github.com/parca-dev/parca-agent/pkg/runtime/python.interpreterExecutableFile.copyMemory({0xc000d94038, 0xc000ef2320, 0x5222, 0x202000, 0xc0000c6960}, 0xc000672a20?, {0x3ea73c0?, 0x28?, 0x242f800?})
        github.com/parca-dev/parca-agent/pkg/runtime/python/python.go:238 +0x28e
github.com/parca-dev/parca-agent/pkg/runtime/python.versionFromBSS(0xc0000c6a50)
        github.com/parca-dev/parca-agent/pkg/runtime/python/python.go:119 +0x233
github.com/parca-dev/parca-agent/pkg/runtime/python.newInterpreter({0xc000ef2280?, {{0x2467ec8?, 0x31?}, 0x90?}})
        github.com/parca-dev/parca-agent/pkg/runtime/python/python.go:560 +0x655
github.com/parca-dev/parca-agent/pkg/runtime/python.InterpreterInfo({0xc001f880c0?, {{0x2467ec8?, 0xc001199698?}, 0x10?}})
        github.com/parca-dev/parca-agent/pkg/runtime/python/python.go:588 +0x4b
github.com/parca-dev/parca-agent/pkg/runtime/interpreter.Fetch({0xc000000c00?, {{0x2467ec8?, 0xc000da7950?}, 0x80?}})
        github.com/parca-dev/parca-agent/pkg/runtime/interpreter/interpreter.go:43 +0x72
github.com/parca-dev/parca-agent/pkg/process.(*InfoManager).fetch(0xc000000c00, {0x2b2b1d0, 0xc000da7950}, 0x5222, 0x0)
        github.com/parca-dev/parca-agent/pkg/process/info.go:299 +0x807
github.com/parca-dev/parca-agent/pkg/process.(*InfoManager).Fetch(0xc000000c00, {0x2b2b1d0, 0xc0024845d0}, 0xc000d8ab00?)
[...]

PR #2312, opened 2023-11-29
Skip empty sections

javierhonduco:
Not sure what conditions have to happen for this to be the case, but we've seen some suspicious panics that can be due to an empty buffer.


brancz (approved):

kakkoyun:
Fixed #2311

PR #2316, opened 2023-11-29
pkg/metadata/compiler: Use filepath.Join

brancz:


javierhonduco (approved):
Awesome, this should add a slash if needed!

Issue #2314, opened 2023-11-29
Python stack unwinding fails, but with `null` error


Garbett1:
**Describe the bug**
Parca agent stack unwinding fails to add the unwind table, but also has a `null` error which is surprising.

**To Reproduce**
Can't replicate with a minimal example yet, but happens consistently with the specific binary.

**Logs**
```
kubectl logs -n parca parca-agent-rb55j | grep 18846

level=debug name=parca-agent ts=2023-11-29T12:45:32.972851349Z caller=cpu.go:536 component=cpu_profiler msg="adding unwind tables" pid=18846
level=error name=parca-agent ts=2023-11-29T12:45:32.975991544Z caller=cpu.go:557 component=cpu_profiler msg="failed to add unwind table" pid=18846 err=null
level=debug name=parca-agent ts=2023-11-29T12:45:33.170072189Z caller=cpu.go:531 component=cpu_profiler msg="prefetching process info" pid=18846
level=debug name=parca-agent ts=2023-11-29T12:45:33.170377246Z caller=cpu.go:536 component=cpu_profiler msg="adding unwind tables" pid=18846
level=error name=parca-agent ts=2023-11-29T12:45:33.173663715Z caller=cpu.go:557 component=cpu_profiler msg="failed to add unwind table" pid=18846 err=null
level=debug name=parca-agent ts=2023-11-29T12:45:35.182409458Z caller=cpu.go:531 component=cpu_profiler msg="prefetching process info" pid=18846
level=debug name=parca-agent ts=2023-11-29T12:45:35.182471246Z caller=cpu.go:536 component=cpu_profiler msg="adding unwind tables" pid=18846
level=error name=parca-agent ts=2023-11-29T12:45:35.184726234Z caller=cpu.go:557 component=cpu_profiler msg="failed to add unwind table" pid=18846 err=null
level=debug name=parca-agent ts=2023-11-29T12:45:44.121114761Z caller=cpu.go:531 component=cpu_profiler msg="prefetching process info" pid=18846
level=debug name=parca-agent ts=2023-11-29T12:45:44.12125953Z caller=cpu.go:536 component=cpu_profiler msg="adding unwind tables" pid=18846
level=error name=parca-agent ts=2023-11-29T12:45:44.123850309Z caller=cpu.go:557 component=cpu_profiler msg="failed to add unwind table" pid=18846 err=null
[...]

PR #2318, opened 2023-11-29
profiler: Fix error handling

javierhonduco:
The err variable was clobbered so we were reading the wrong value

Test Plan
========

```
--- a/pkg/profiler/cpu/bpf/maps/maps.go
+++ b/pkg/profiler/cpu/bpf/maps/maps.go
@@ -1140,6 +1140,7 @@ func (m *Maps) AddUnwindTableForProcess(pid int, executableMappings unwind.Execu
        // is challenging if the process name contains spaces, etc).
        //  - PIDs can be recycled.

+       return fmt.Errorf("le error")
        m.mutex.Lock()
        defer m.mutex.Unlock()
```

Before:
```
msg="failed to add unwind table" pid=18846 err=null
```

After:
```
msg="failed to add unwind table" pid=18846 err="le err"
```

### DIFF
--- a/pkg/profiler/cpu/cpu.go
+++ b/pkg/profiler/cpu/cpu.go
@@ -549,7 +549,7 @@ func (p *CPU) addUnwindTableForProcess(ctx context.Context, pid int) {
 	}
 
 	level.Debug(p.logger).Log("msg", "adding unwind tables", "pid", pid)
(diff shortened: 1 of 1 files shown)


javierhonduco:
Fixes https://github.com/parca-dev/parca-agent/issues/2314

brancz (approved):

PR #2321, opened 2023-11-30
pkg/symtab: Use bufio.Writer to reduce write syscalls

brancz:
### Why?

Users have reported CPU spikes with workloads that have perfmaps/jitdumps.

### What?

CPU profiling data showed clearly that the most time is spent in write syscalls during those spikes.

### How?

Buffering writes should amortize most of the CPU time spent on write syscalls.

### Test Plan

Tested against some jitdump/perfmap workloads.

### DIFF
--- a/pkg/symtab/symtab.go
+++ b/pkg/symtab/symtab.go
@@ -14,6 +14,7 @@
 package symtab
 
 import (
+	"bufio"
 	"encoding/binary"
 	"errors"
 	"fmt"
@@ -86,6 +87,7 @@ type entry struct {
 
 type FileWriter struct {
 	file         *os.File
+	w            *bufio.Writer
 	entries      []entry
 	stringOffset uint32
 	finalized    bool
@@ -104,6 +106,7 @@ func NewWriter(path string, preallocate int) (*FileWriter, error) {
 	}
 	return &FileWriter{
 		file:    file,
+		w:       bufio.NewWriter(file),
 		entries: make([]entry, 0, preallocate),
 	}, nil
 }
@@ -113,12 +116,12 @@ func (fw *FileWriter) AddSymbol(name string, address uint64) error {
 		return errAlreadyFinalized
 	}
 
-	_, err := fw.file.WriteString(name)
+	_, err := fw.w.WriteString(name)
 	if err != nil {
 		return fmt.Errorf("WriteString: %w", err)
 	}
 	// Append nil to make debugging easier.
-	_, err = fw.file.WriteString("\000")
+	_, err = fw.w.WriteString("\000")
 	if err != nil {
 		return fmt.Errorf("WriteString: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


javierhonduco:
Amazing! Had a very similar patch ready and the default buffer size didn't work in my experiments, could you please run the tests for the symtab package under `strace -fc` (to follow and summarise syscalls) and under `perf stat` in main and on your branch?

pkg/symtab/symtab.go
```diff
@@ -165,17 +170,20 @@ func (fw *FileWriter) Write() error {
...
 		}
 	}
 
-	err := fw.writeHeader()
-	if err != nil {
+	if err := fw.writeHeader(); err != nil {
 		return fmt.Errorf("writeHeader: %w", err)
 	}
 
+	if err := fw.w.Flush(); err != nil {
```
javierhonduco, line 183: Should this be in the defer lambda above?

javierhonduco:
The perf stat numbers look good (but it might be noise, not sure if you run it multiple times to ensure everything's cached by the OS). One small tip is to run the generated binary directly so we are not accounting for all the time spent by the `go test` infrastructure.

Main:
```
  0.21    0.010071           3      2975           write
```

With this PR:
```
  0.25    0.011625           4      2880           write
```

The number of writes seems similar, perhaps this could mean that an increase in the buffer size would be beneficial? What do you think?

brancz:
The existing tests weren't sufficient to show what happens with larger files (which is more representative in practice), so I added a new benchmark, and the improvement is quite noticeable:

```
goos: linux
goarch: amd64
pkg: github.com/parca-dev/parca-agent/pkg/symtab
cpu: AMD Ryzen 5 3400GE with Radeon Vega Graphics
                               │    old.txt    │              new2.txt               │
                               │    sec/op     │    sec/op     vs base               │
OptimizedSymbolizerManyWrite-8   54.039m ± ∞ ¹   2.817m ± ∞ ¹  -94.79% (p=0.008 n=5)
¹ need >= 6 samples for confidence interval at level 0.95

                               │    old.txt    │               new2.txt               │
                               │     B/op      │     B/op       vs base               │
OptimizedSymbolizerManyWrite-8   551.2Ki ± ∞ ¹   164.5Ki ± ∞ ¹  -70.16% (p=0.008 n=5)
¹ need >= 6 samples for confidence interval at level 0.95

                               │    old.txt     │              new2.txt              │
                               │   allocs/op    │  allocs/op   vs base               │
OptimizedSymbolizerManyWrite-8   40015.00 ± ∞ ¹   17.00 ± ∞ ¹  -99.96% (p=0.008 n=5)
¹ need >= 6 samples for confidence interval at level 0.95
```

Almost 95% better!

javierhonduco (approved):
LGTM, great stuff!! Thanks for making this faster now that the use-cases have grown!

PR #2332, opened 2023-12-03
 weekly build script to test on latest clang version

### LINKED ISSUE #1208: ci: Test and build with more clang versions
We use [Clang (and related LLVM toolchain) version 11](https://github.com/parca-dev/parca-agent/blob/50aec92dba54d0358a633e17d23ebfba3c1a76ab/.github/workflows/build.yml#L66) released the 12 October 2020	in CI for builds and tests.

Locally, some of us run more modern versions of Clang and LLVM. My system reports:
``` 
$ clang --version
clang version 14.0.5 (Fedora 14.0.5-2.fc36)
Target: x86_64-redhat-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

It would be great to have a matrix of Clang / LLVM versions that we want to support. I think we should have:
- The minimum version of the LLVM we want to support (e.g. LLVM 11)
- The "stable" version of LLVM that most of us run (e.g. LLVM 14);
- Latest LLVM, either a release candidate or some recent version from their main branch, to ensure that there are no issues with the latest changes upstream and our BPF code

By doing this we will ensure that everything works as expected with new LLVM releases. It will also help increase our confidence when bumping the  LLVM version in CI, taking advantage of the latest improvements in our BPF compilation toolchain.

Given that this would increase the build time, perhaps this could be a periodic action that we run on a weekly basis.

Namanl2001:
### Why?

Fixes #1208
By doing this we will ensure that everything works as expected with new LLVM releases.

### What?

Newly added weekly-build.yml is same as build.yml except this new script will run weekly every Saturday at midnight and for clang version "17.0.6" (which is latest version as of 2023/12/02)

### Test Plan

I guess this can be tested once merged to `main`


.github/workflows/weekly-build.yml
```diff
@@ -0,0 +1,168 @@
...
+          go-version-file: .go-version
+
+      # Installs gofumpt, golangci-lint, and govulncheck.
+      - name: Set up environment
+        run: ./env.sh
+
+      - name: Set up Clang
+        uses: KyleMayes/install-llvm-action@be40c5af3a4adc3e4a03199995ab73aa37536712 # v1.9.0
+        with:
+          version: "17.0.6"
```
kakkoyun, line 108: Maybe we should use https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs and test against several versions.

kakkoyun (commented):
Thanks for tackling this.

Could we make this runnable on demand as well? https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow

.github/workflows/weekly-build.yml
Namanl2001: sounds good; since this GH action will run once a week, we can manage to test on multiple versions

In [build.yml](https://github.com/parca-dev/parca-agent/blob/main/.github/workflows/build.yml#L109-L112) we regularly test on clang version `14`, what versions would you recommend for this new weekly action, perhaps, [`15.0.7`, `16.0.6`, `17.0.6`] which are the latest patch releases of major release versions. WDYT?

kakkoyun (commented):
Thanks a lot for your contribution.

I'm introducing some new changes to the CI/CD pipeline. Let's hold onto this one for a little more. We might need to rewrite it.

Stay tuned.

brancz:
Outdated since #2958 

The bpf programs are no longer built in this repository.

PR #2367, opened 2023-12-14
unwinder/native: Increase size of mappings

javierhonduco:
This commit increases the maximum mappings a process without FP can have. This limitation affected a small number of users, but it was still a blocker for some.

Additionally, rather than a linear search, we switch to using binary search on the mappings as otherwise it's too many instructions for the BPF verifier to check.

Test Plan
=========

```
=== RUN   TestCPUProfilerWorks/mixed_mode_unwinding
=== RUN   TestCPUProfilerWorks/unwinder_metrics_work
--- PASS: TestCPUProfilerWorks (20.79s)
    --- PASS: TestCPUProfilerWorks/dwarf_unwinding (0.08s)
    --- PASS: TestCPUProfilerWorks/fp_unwinding (0.67s)
    --- PASS: TestCPUProfilerWorks/mixed_mode_unwinding (0.04s)
    --- PASS: TestCPUProfilerWorks/unwinder_metrics_work (0.00s)
PASS
ok      github.com/parca-dev/parca-agent/test/integration       20.810s
```

and kernel tests pass

### DIFF
--- a/bpf/unwinders/native.bpf.c
+++ b/bpf/unwinders/native.bpf.c
@@ -47,17 +47,19 @@ _Static_assert(MAX_TAIL_CALLS *MAX_STACK_DEPTH_PER_PROGRAM >= MAX_STACK_DEPTH, "
 #define MAX_PROCESSES 5000
 // Binary search iterations for dwarf based stack walking.
 // 2^19 can bisect ~524_288 entries.
-#define MAX_BINARY_SEARCH_DEPTH 19
+#define MAX_UNWIND_INFO_BINARY_SEARCH_DEPTH 19
 // Size of the unwind table.
 // 250k * sizeof(stack_unwind_row_t) = 2MB
 #define MAX_UNWIND_TABLE_SIZE 250 * 1000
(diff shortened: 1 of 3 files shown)


javierhonduco:
I thought of generalising binary search for both mappings and unwind information but unfortunately, the result would be harder to maintain (macros in C, ugh) than just having two implementations. The code is compact enough that making modifications to both, if needed, shouldn't be a big overhead. We should be careful not to forget to keep both binary searches updated, though.

brancz (approved):
Maybe we should comment on the two implementations that they need to be kept in sync? Otherwise lgtm.

kakkoyun (approved):

javierhonduco:
> Maybe we should comment on the two implementations that they need to be kept in sync? Otherwise lgtm.

They don't necessarily have to be kept 100% in sync, e.g. if we want to change something for one of them in the future that only affects mappings or unwind info, it's mostly in case we find out that the current algorithm has an off by one or integer overflow or something similar.

Now that I am writing this, we might have one, will prep a fix!

javierhonduco:
Can add the comment later if you feel strongly about it!

brancz:
Nope all good, thanks for the explanation!

PR #1602, opened 2023-05-04
Locally download and use libbpf dependencies

kakkoyun:
### Why?

After `libbpf` started requiring `zstd` as a dependency, my system can't build libbpf locally. That's why I have started changes. (Thanks @javierhonduco, for the pointers). 

This is a first step towards to system-independent way of reproducibly building our binary. This will also simplify our CI dependencies. We need to enhance the script to handle cross-platform builds and configure goreleaser accordingly.

related: https://github.com/parca-dev/parca-agent/issues/1304

### <samp>🤖 Generated by Copilot at f5b7e98</samp>

This pull request adds support for building the parca-agent binary with static libraries for libbpf and its dependencies. It introduces a new script `scripts/download-and-build-libbpf-deps.sh` and modifies the `Makefile` and the `Tiltfile` accordingly. This enables the parca-agent binary to run on systems without dynamic libraries for libbpf.

### <samp>🤖 Generated by Copilot at f5b7e98</samp>

*  Define and use variables for libbpf dependencies in `Makefile` ([link](https://github.com/parca-dev/parca-agent/pull/1602/files?diff=unified&w=0#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R54-R56), [link](https://github.com/parca-dev/parca-agent/pull/1602/files?diff=unified&w=0#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L69-R77), [link](https://github.com/parca-dev/parca-agent/pull/1602/files?diff=unified&w=0#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L105-R108), [link](https://github.com/parca-dev/parca-agent/pull/1602/files?diff=unified&w=0#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L111-R114), [link](https://github.com/parca-dev/parca-agent/pull/1602/files?diff=unified&w=0#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L173-R194))
* Add a new script `scripts/download-and-build-libbpf-deps.sh` to download and build the static libraries for libbpf dependencies ([link](https://github.com/parca-dev/parca-agent/pull/1602/files?diff=unified&w=0#diff-2df6b8021f4eb6f37f6ecd921284f2f70e7c340c237d6292a477127278fb4d6eR1-R100))
[...]

PR #2368, opened 2023-12-14
New build pipeline: Isolate dependencies using devbox

kakkoyun:

> [!NOTE]
> This targets https://github.com/parca-dev/parca-agent/tree/new_build_pipeline and there will be a PR in the end targets the `main`.

From the users' perspective, this PR shouldn't change anything. If you want to use specified dependencies please consider installing the devbox and direnv (simply `make init`)

This brings the result of the experimentation for a new build pipeline from https://github.com/parca-dev/testdata/pull/27 to the agent.

At first, we will start by introducing the https://www.jetpack.io/devbox

> Now, we use https://www.jetpack.io/devbox for managing external and tool dependencies, such as the Go toolchain or C compiler. This tool has been chosen because it has all the benefits of [Nix](https://nixos.org/), but we don't have to deal with all the intricacies of Nix. It also has an excellent integration with [direnv](https://direnv.net/) to load the environment automatically when you enter the directory.

Moreover, this PR:
- Changes the directory structure a little bit to reduce the number of top-level directories (I will introduce more) and group them.
- It also vendors in the dependencies of libbpf as git submodules and compiles it. It's the same strategy as in https://github.com/parca-dev/parca-agent/pull/1602 (that PR will be deprecated). It is easier to manage than using some external tool. Especially when we want to cross-compile them. (Makefile changes)
- The GH actions are updated to utilize the devbox.

### Test Plan
- See that CI is green.
- (Locally) Just run the make targets as before.


bpf/Makefile
```diff
@@ -49,6 +56,9 @@ c/fmt:
 .PHONY: format-check
 format-check:
 
+# TODO(kakkoyun): Fix.
+# clang-14: error: argument unused during compilation: '--gcc-toolchain=/nix/store/hf2gy3km07d5m0p1lwmja0rg9wlnmyr7-gcc-12.3.0' [-Werror,-Wunused-command-line-argument]
+
```
Sylfrena, line 61: let's fix this first

bpf/Makefile
```diff
@@ -135,3 +146,10 @@ $(OUT_PYPERF): $(PYPERF_SRC) $(LIBBPF_HEADERS) $(BPF_HEADERS) | $(OUT_DIR)
 		-O2 -emit-llvm -c -g $< -o $(@:.o=.ll)
 	$(CMD_LLC) -march=bpf -filetype=obj -o $@ $(@:.o=.ll)
 	rm $(@:.o=.ll)
+
+.PHONY: env
+env:
+	$(foreach v,                                        \
+		$(filter-out $(VARS_OLD) VARS_OLD,$(.VARIABLES)), \
+		$(info $(v) = $($(v))))
+	@echo "---"
```
Sylfrena, line 155: I might be missing something, but can't really understand what's happening here?
I think we should put a comment in here, mostly because it's difficult to understand context from just the variable names

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/parca-dev/parca-agent
 
-go 1.21.5
+go 1.21.1
+
+toolchain go1.21.4
```
Sylfrena, line 5: ah, any reason we went back to a previous version? 
maybe worth commenting why to avoid regressions if someone updates it

test/kernel/kernels/README.md
Sylfrena, line 1: nit: I think `test/kerneltests` is a better idea (and leaves less room for path typos)

bpf/Makefile
kakkoyun: As I explained in other communication channels, this is a part of a series of PRs, which will be deprecated. This is an inherent issue because of the nix toolchain interfering with what we do. The clang picks up and environment variable and adds this implicit flag. I don't think it's productive to waster time on this at this stage, hence the TODO.

bpf/Makefile
kakkoyun: This is just for debugging. And soon you don't have to deal with makefiles. Stay tuned.

go.mod
kakkoyun: The nix package doesn't have the latest Golang version. I can upgrade it to the latest version, and the go toolchain will download the missing tooling. However, we want reproducible builds, thus the downgrade. I believe these particular security fixes don't concern us.

test/kernel/kernels/README.md
kakkoyun: Agree to disagree 😊 I don't like the stutter. We have a similar situation with `test/testdata` but `testdata` is special.

go.mod
Sylfrena: just to check I got it correct - nix doesn't have latest golang and we won't have reproducible builds without nix so we are using the previous version?

if so, and someone bumps the version just the same - do we have tests for reproducible builds (in this or another PR)?

.envrc
```diff
@@ -0,0 +1,7 @@
+# Automatically sets up your devbox environment whenever you cd into this
+# directory via our direnv integration:
+
+eval "$(devbox generate direnv --print-envrc)"
```
maxbrunet, line 4: I would move this into the `shellHook` of the Nix shell, and replace the content of `.envrc` with:

```bash
use flake

watch_file nix/devshell.nix
```

This is currently what I have in my local `.envrc` and not doing `use flake` here would likely make the Nix shell with `direnv` unusable. This also has the advantage of setting up `devbox` for you when using `direnv`.

.github/workflows/container.yml
```diff
@@ -98,27 +103,41 @@ jobs:
...
+          key: devbox-go-tools.cache-${{ runner.arch }}-${{ hashFiles('install-tools.sh') }}
+          restore-keys: |
+            devbox-go-tools.cache-${{ runner.arch }}
+            devbox-go-tools.cache-
+
+      - name: Setup devbox
+        run: devbox run -- echo "done!"
+
+      - name: Load devbox shellenv
+        uses: HatsuneMiku3939/direnv-action@v1
```
maxbrunet, line 137:
```suggestion
        uses: HatsuneMiku3939/direnv-action@v1.0.4
```

Multiple occurrences, need to that for all of them

.github/workflows/container.yml
```diff
@@ -98,27 +103,41 @@ jobs:
...
+        with:
+          direnvVersion: 2.32.3
```
maxbrunet, line 139: Probably want to set this version like the others for Renovate

```suggestion
          direnvVersion: ${{ DIRENV_VERSION }}
```

Multiple occurrences, need to that for all of them

.github/workflows/pre-commit.yml
```diff
@@ -74,12 +81,9 @@ jobs:
             golangci-lint.cache-${{ env.INTERVAL_KEY }}-
             golangci-lint.cache-
 
-      - name: Set up pre-commit
-        run: pipx install "pre-commit${PRE_COMMIT_VERSION}"
-
       - name: Set pre-commit cache key parts
         run: |
-          echo "GO_VERSION=$(<.go-version)" >> $GITHUB_ENV
+          echo "GO_VERSION=$(go version)" >> $GITHUB_ENV
```
maxbrunet, line 86: Let's do that in 2 steps to avoid masking the exit code of `go version` if it fails

```suggestion
          GO_VERSION="$(go version)"
          echo "GO_VERSION=${GO_VERSION}" >> $GITHUB_ENV
```

https://www.shellcheck.net/wiki/SC2312

.github/workflows/container.yml
```diff
@@ -266,13 +275,28 @@ jobs:
       - name: Check out the code
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
 
-      - name: Set up Go
-        uses: actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe # v4.1.0
+      - name: Install devbox
+        uses: jetpack-io/devbox-install-action@v0.7.0
```
maxbrunet, line 279: It might be possible to remove this action if `direnv` uses the Nix flake

Would still require to install Nix tho, [cachix/install-nix-action](https://github.com/cachix/install-nix-action) could do that

Dockerfile.builder
```diff
@@ -0,0 +1,17 @@
+FROM jetpackio/devbox:latest
```
maxbrunet, line 1: nit

```suggestion
FROM docker.io/jetpackio/devbox:latest
```

devbox.json
```diff
@@ -0,0 +1,40 @@
...
+    "autoconf@2.71",
+    "gettext@0.21.1",
+    "automake@1.16.5",
+    "libtool@2.4.7",
+    "glibc.dev",
+    "glibc.static",
+    "llvmPackages_14.stdenv@latest",
+    "llvmPackages_14.bintools@latest",
+    "llvmPackages_14.lld@latest"
+  ],
```
maxbrunet, line 20: What's the process to update the packages?

Maybe we need to configure a new [regex manager](https://docs.renovatebot.com/modules/manager/regex/) with Renovate? The [Repology datasource](https://docs.renovatebot.com/modules/datasource/repology/) could be of use

install-tools.sh
maxbrunet, line 1: I would move all these packages to `devbox`, it's more robust way of installing these dev dependencies

nix/devshell.nix
```diff
@@ -1,39 +1,23 @@
-{ pkgs }:
+{ pkgs ? import <nixpkgs> { } }:
```
maxbrunet, line 1: The uses a Nix channel, which is not pinned and hopefully this feature will soon be deprecated. For reproducibility, we should continue using the Flake as source: 

```suggestion
{ pkgs }:
```

Flakes are the future of Nix, if the feature is not enabled on your Nix installation yet, here is how to: https://nixos.wiki/wiki/Flakes

Also if Nix Flakes are not enabled, the old `nix shell` command should still work as [a compatibility layer](https://github.com/edolstra/flake-compat) is configured

nix/go-tools.nix
maxbrunet, line 1: I would suggest keeping this file, this is more robust than `go install`, and with a few simple changes, easy to integrate with `devbox`: https://www.jetpack.io/devbox/docs/guides/using_flakes/#using-a-local-flake

maxbrunet (commented):
Exciting!

I think the key to make this change great is to use `devbox` with Nix Flake. As using Nix Flake in parallel of Devbox as it is currently suggested in this PR is likely impossible. And using Devbox without Nix Flake lacks some elements of reproducibility. Used together, it should still be possible to use the user-friendly `devbox` commands as well as its JSON config for most operations, while getting the benefits of Nix Flake.

kakkoyun:
> I think the key to make this change great is to use `devbox` with Nix Flake. As using Nix Flake in parallel of Devbox as it is currently suggested in this PR is likely impossible. And using Devbox without Nix Flake lacks some elements of reproducibility. Used together, it should still be possible to use the user-friendly `devbox` commands as well as its JSON config for most operations, while getting the benefits of Nix Flake.

@maxbrunet Thank you so much for the review! You're awesome!

I should have clarified several points. Maybe you can give another round of review after this and maybe with some suggestions.

I want to use `devbox` as our strict dependencies for building the applications. I'd like to keep using `devshell.nix` for extended dev dependencies for tools like `kubectl`, etc. It is not necessarily needed to build the application, but maybe testing it locally.

I'm no Nix expert by all means (not even a user), so I might have missed some best practices. Feel free to guide me.

I will introduce another set of changes to eliminate all the custom bash scripts and makefiles using [Mage](https://magefile.org), so in a sense, this PR is incomplete. I'll send PRs on top of this one. So please stay tuned.

I want a simple layered approach for the build pipeline, tooling, and CI/CD. It will be more verbose because we will be using Go. However, it will be more straightforward to understand.

kakkoyun:
I will address the concerns in subsequent PRs before everything gets merged into the main branch.

Issue #2382, opened 2023-12-25
[BUG] IMPORTANT! Default option sending  AnalyticsOptOut to parca now


joeyleeeeeee97:
**Describe the bug**
	AnalyticsOptOut bool `default:"false" help:"Opt out of sending anonymous usage statistics."`

But https://github.com/parca-dev/parca-agent/blob/f7e8ce31846b49ebf0acb836cf84c01c1a09b076/cmd/parca-agent/main.go#L694

        if !flags.AnalyticsOptOut {

This should be  if flags.AnalyticsOptOut.

**I think this is super important... We need people's consent to collect their analytic info like machines.**

**To Reproduce**
Run a parca demo you can easily see this

parca-agent 2455983 root  227u     IPv4          520413940      0t0       TCP xxx:49518->51-158-227-244.lb.nl-ams.scw.cloud:https (ESTABLISHED)

**Expected behavior**
No report by default

**Screenshots** or **Profiles** (using https://pprof.me)

**Logs**

**Software (please complete the following information):**
 - Parca Agent Version: [`parca-agent --version`]
 - Parca Server Version (if applicable): [`parca --version`]

**Workload (please complete the following information):**
- Runtime (if applicable):
- Compiler (if applicable):

**Environment (please complete the following information):**
- Linux Distribution: [`cat /etc/*-release`]
- Linux Version: [`uname -r`]
- Arch: [`uname -m`]
- Kubernetes Version (if applicable): [`kubectl --version`]
- Container Runtime (if applicable):

**Additional context**

joeyleeeeeee97:
@brancz

PR #2383, opened 2023-12-25
Disable sending analytics report by default

joeyleeeeeee97:
### Why?

As in https://github.com/parca-dev/parca-agent/issues/2382 , We need people's consent to collect their analytic info like machines.

### What?

change check logic

### DIFF
--- a/cmd/parca-agent/main.go
+++ b/cmd/parca-agent/main.go
@@ -691,7 +691,7 @@ func run(logger log.Logger, reg *prometheus.Registry, flags flags) error {
 		})
 	}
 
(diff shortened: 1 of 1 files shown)


brancz:
While I understand where you’re coming from, it’s entirely customary now that open source projects send a small amount of anonymized data to inform development decisions. Parca agent has to run on an extremely huge matrix of combinations of parameters and at least understanding a little bit of what our user base looks like helps us prioritize the right things.

We’re not changing this default.

joeyleeeeeee97:
ok

PR #2395, opened 2024-01-02
New Build Pipeline: Use Zig as compiler

kakkoyun:
This PR compiles dependencies, using [Zig Build System](https://ziglang.org/learn/build-system).

As of now, it doesn't address cross-compilation concerns.

- It uses `musl` and statically links everything. 
- It reduces build time dependecies.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,3 +13,6 @@
 [submodule "github.com/parca-dev/testdata"]
 	path = test/testdata
 	url = https://github.com/parca-dev/testdata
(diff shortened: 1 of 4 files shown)


brancz:
devbox + direnv is a pretty neat combo! The overall strategy of this change makes a ton of sense to me.

javierhonduco:
For what I see in the PR, it seems like we are using musl for the native C dependencies as our libc, not GNUs, is that correct?

PR #2411, opened 2024-01-09
New Build Pipeline

kakkoyun:
TBA

This is a very early-stage PR. There will be PRs that are going to be merged onto this and then to the main. Stay tuned.


javierhonduco:
Failing with ```-buildmode=pie not supported when -race is enabled```

PR #2476, opened 2024-02-05
fix: bpf struct sizes fixes

korniltsev:
### Why?

1. I noticed `mappingInfoSizeBytes` is off by `0x10` which may lead to reading little bit of garbage out of bounds of the allocation.
2. Buffer growing in various places seems incorrect - it should not take pointers
https://go.dev/play/p/wSGZw-SAocA

### DIFF
--- a/pkg/profiler/cpu/bpf/maps/maps.go
+++ b/pkg/profiler/cpu/bpf/maps/maps.go
@@ -102,7 +102,7 @@ const (
 			mapping_t mappings[MAX_MAPPINGS_PER_PROCESS];
 		} process_info_t;
 	*/
-	mappingInfoSizeBytes = 8 + 8 + (maxMappingsPerProcess * 8 * 5)
+	mappingInfoSizeBytes = 8*4 + (maxMappingsPerProcess * 8 * 5)
 	/*
 		TODO: once we generate the bindings automatically, remove this.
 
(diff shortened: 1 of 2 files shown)


pkg/profiler/cpu/bpf/maps/maps.go
```diff
@@ -102,7 +102,7 @@ const (
 			mapping_t mappings[MAX_MAPPINGS_PER_PROCESS];
 		} process_info_t;
 	*/
-	mappingInfoSizeBytes = 8 + 8 + (maxMappingsPerProcess * 8 * 5)
+	mappingInfoSizeBytes = 8*4 + (maxMappingsPerProcess * 8 * 5)
```
brancz, line 105: I don't think I understand this change, `(maxMappingsPerProcess * 8 * 5)` makes sense because it's the mapping struct (each field is 8 bytes), but where do the static `8 * 4` come from?

korniltsev: four extra fields  8 bytes each here https://github.com/parca-dev/parca-agent/blob/7a532053d10a220dc32362481f82487c80e55665/bpf/unwinders/native.bpf.c#L205-L208

see how in go comment above it is 
```
typedef struct {
	u64 is_jit_compiler;
	u64 len;
	mapping_t mappings[MAX_MAPPINGS_PER_PROCESS];
} process_info_t;
```

and in kernel it is 

```
typedef struct {
  u64 should_use_fp_by_default;
  u64 is_jit_compiler;
  u64 interpreter_type;
  u64 len;
  mapping_t mappings[MAX_MAPPINGS_PER_PROCESS];
} process_info_t;
```

so it used to be `8 + 8 + (maxMappingsPerProcess * 8 * 5)`
and now it should be `8 + 8 + 8 + 8 + (maxMappingsPerProcess * 8 * 5)`

Does it make sense?

brancz: Can you update the comment with that?

korniltsev: updated

kakkoyun (approved):

Issue #2459, opened 2024-01-29
Support DWARF CFA opcode: 0x26


brancz:
Found in the wild by a Polar Signals Cloud customer:

```
panic: Encountered an unexpected DWARF CFA opcode: 0x26
goroutine 240 [running]:
github.com/parca-dev/parca-agent/internal/dwarf/frame.lookupFunc(0x50?, 0x1bdbac4?)
	github.com/parca-dev/parca-agent/internal/dwarf/frame/table.go:404 +0x360
github.com/parca-dev/parca-agent/internal/dwarf/frame.executeDWARFInstruction(0x4000c34a50)
	github.com/parca-dev/parca-agent/internal/dwarf/frame/table.go:309 +0x84
github.com/parca-dev/parca-agent/internal/dwarf/frame.(*InstructionContextIterator).Next(0x40024fb730)
	github.com/parca-dev/parca-agent/internal/dwarf/frame/table.go:90 +0x5c
github.com/parca-dev/parca-agent/pkg/stack/unwind.BuildCompactUnwindTable({0x4000bf6000, 0xcdd, 0x1000?}, 0xadd0?)
	github.com/parca-dev/parca-agent/pkg/stack/unwind/compact_unwind_table.go:157 +0x2c8
github.com/parca-dev/parca-agent/pkg/stack/unwind.GenerateCompactUnwindTable({0x40023f69c0?, 0x4b0?})
	github.com/parca-dev/parca-agent/pkg/stack/unwind/compact_unwind_table.go:283 +0x60
github.com/parca-dev/parca-agent/pkg/profiler/cpu/bpf/maps.(*Maps).setUnwindTableForMapping(0x40005ee600, 0x37?, 0x0?, 0x4000924840)
	github.com/parca-dev/parca-agent/pkg/profiler/cpu/bpf/maps/maps.go:1557 +0xab8
github.com/parca-dev/parca-agent/pkg/profiler/cpu/bpf/maps.(*Maps).AddUnwindTableForProcess(0x40005ee600, 0xd1c, {0x0, 0x0, 0x0}, 0x1, 0x0)
	github.com/parca-dev/parca-agent/pkg/profiler/cpu/bpf/maps/maps.go:1211 +0x604
github.com/parca-dev/parca-agent/pkg/profiler/cpu.(*CPU).addUnwindTableForProcess(0x400232da40, {0x29dbae8, 0x4000afd9e0}, 0x0?)
[...]

PR #2479, opened 2024-02-06
DWARF: Add logging for unexpected CFA opcodes

Sylfrena:
### Why?
An end-user found an unexpected DWARF opcode in #2459 causing the agent to panic.

As there was no logging, the binary which has this opcode could not be determined

### Test Plan

CI tests

### DIFF
--- a/internal/dwarf/frame/table.go
+++ b/internal/dwarf/frame/table.go
@@ -265,7 +265,7 @@ func NewContext(logger log.Logger, fullExecutablePath string) *Context {
 		buf:                &bytes.Reader{},
 		rememberedState:    newStateStack(),
 		fullExecutablePath: fullExecutablePath,
-		logger:             logger,
+		logger:             log.With(logger, "component", "dwarf unwind table context"),
 	}
 }
 
@@ -413,8 +413,7 @@ func lookupFunc(instruction byte, ctx *Context) instruction {
 	case DW_CFA_GNU_window_save:
 		fn = gnuwindowsave
 	default:
-		//panic(fmt.Sprintf("Encountered an unexpected DWARF CFA opcode: %#v", instruction))
-		level.Error(ctx.logger).Log("msg", "Encountered an unexpected DWARF CFA opcode", "opcode instruction", instruction, "PID", getPid(ctx.fullExecutablePath), "comm", ctx.fullExecutablePath)
+		level.Error(ctx.logger).Log("msg", "encountered an unexpected DWARF CFA opcode", "opcode instruction", instruction, "PID", getPid(ctx.fullExecutablePath), "comm", ctx.fullExecutablePath)
 	}
 	return fn
 }
--- a/internal/dwarf/frame/table.go
+++ b/internal/dwarf/frame/table.go
@@ -4,7 +4,10 @@ package frame
 import (
 	"bytes"
 	"encoding/binary"
-	"fmt"
+	"strings"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
 
 	"github.com/parca-dev/parca-agent/internal/dwarf/util"
 )
@@ -144,8 +147,10 @@ type Context struct {
 	lastInsCtx      *InstructionContext
 	rememberedState *StateStack
 	// The buffer where we store the dwarf unwind entries to be parsed for this function.
-	buf   *bytes.Reader
-	order binary.ByteOrder
+	buf                *bytes.Reader
+	order              binary.ByteOrder
+	fullExecutablePath string
+	logger             log.Logger
 }
 
 func (ctx *Context) currentInstruction() *InstructionContext {
@@ -253,18 +258,20 @@ const low_6_offset = 0x3f
 
 type instruction func(ctx *Context)
 
-func NewContext() *Context {
+func NewContext(logger log.Logger, fullExecutablePath string) *Context {
 	return &Context{
-		currInsCtx:      &InstructionContext{},
-		lastInsCtx:      &InstructionContext{},
-		buf:             &bytes.Reader{},
-		rememberedState: newStateStack(),
+		currInsCtx:         &InstructionContext{},
+		lastInsCtx:         &InstructionContext{},
+		buf:                &bytes.Reader{},
+		rememberedState:    newStateStack(),
+		fullExecutablePath: fullExecutablePath,
+		logger:             logger,
 	}
 }
 
 func executeCIEInstructions(cie *CommonInformationEntry, context *Context) *Context {
 	if context == nil {
-		context = NewContext()
+		context = NewContext(context.logger, context.fullExecutablePath)
 	}
 
 	context.reset(cie)
@@ -308,6 +315,11 @@ func executeDWARFInstruction(ctx *Context) {
 
 	fn := lookupFunc(instruction, ctx)
 	fn(ctx)
+}
+
+// getPid gets the Process ID as a string from the fullExecutablePath provided
+func getPid(fullExecutablePath string) string {
+	return strings.Split(fullExecutablePath, "/")[2]
 }
 
 func lookupFunc(instruction byte, ctx *Context) instruction {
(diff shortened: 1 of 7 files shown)


internal/dwarf/frame/table.go
```diff
@@ -401,9 +413,9 @@ func lookupFunc(instruction byte, ctx *Context) instruction {
 	case DW_CFA_GNU_window_save:
 		fn = gnuwindowsave
 	default:
-		panic(fmt.Sprintf("Encountered an unexpected DWARF CFA opcode: %#v", instruction))
-	}
 
+		level.Error(ctx.logger).Log("msg", "Encountered an unexpected DWARF CFA opcode", "opcode instruction", instruction, "PID", getPid(ctx.fullExecutablePath), "comm", ctx.fullExecutablePath)
```
kakkoyun, line 417:
```suggestion
		level.Error(ctx.logger).Log("msg", "encountered an unexpected DWARF CFA opcode", "opcode instruction", instruction, "PID", getPid(ctx.fullExecutablePath), "comm", ctx.fullExecutablePath)
```

pkg/stack/unwind/compact_unwind_table_test.go
```diff
@@ -455,8 +456,10 @@ func TestIsSorted(t *testing.T) {
 	matches, err := filepath.Glob("../../../testdata/vendored/x86/*")
 	require.NoError(t, err)
 
+	logger := logger.NewLogger("error", logger.LogFormatLogfmt, "compact-unwind-table-tests")
```
kakkoyun, line 459: You're misusing the API here. What you need is `log.With`. You can find a lot of examples in the codebase.

https://github.com/go-kit/log?tab=readme-ov-file#contextual-loggers

pkg/stack/unwind/unwind_table.go
```diff
@@ -79,7 +79,7 @@ func registerToString(reg uint64, arch elf.Machine) string {
 }
 
 // PrintTable is a debugging helper that prints the unwinding table to the given io.Writer.
-func (ptb *UnwindTableBuilder) PrintTable(writer io.Writer, path string, compact bool, pc *uint64) error {
+func (ptb *UnwindTableBuilder) PrintTable(writer io.Writer, logger log.Logger, path string, compact bool, pc *uint64) error {
```
kakkoyun, line 82:
```suggestion
func (ptb *UnwindTableBuilder) PrintTable(logger log.Logger, writer io.Writer, path string, compact bool, pc *uint64) error {
```

Putting the injected dependencies first would be better. This is what we usually do throughout the codebase.

pkg/stack/unwind/compact_unwind_table_test.go
Sylfrena: TIL about contextual logging, thanks!🤩
Have updated the PR

internal/dwarf/frame/table.go
```diff
@@ -401,9 +413,8 @@ func lookupFunc(instruction byte, ctx *Context) instruction {
 	case DW_CFA_GNU_window_save:
 		fn = gnuwindowsave
 	default:
-		panic(fmt.Sprintf("Encountered an unexpected DWARF CFA opcode: %#v", instruction))
+		level.Error(ctx.logger).Log("msg", "encountered an unexpected DWARF CFA opcode", "opcode instruction", instruction, "PID", getPid(ctx.fullExecutablePath), "comm", ctx.fullExecutablePath)
```
brancz, line 416: I would prefer if we didn't pass the logger around like this and instead have this function return an error that we then pass up and layer on context as we pass the error further up.

internal/dwarf/frame/table.go
```diff
@@ -253,38 +258,49 @@ const low_6_offset = 0x3f
 
 type instruction func(ctx *Context)
 
-func NewContext() *Context {
+func NewContext(fullExecutablePath string) *Context {
```
brancz, line 261: We don't need to pass this down, we just wrap the error returned by `BuildCompactUnwindTable` with `fmt.Errorf("build compact unwind table for executable %q: %w", fullExecutablePath, err)`

internal/dwarf/frame/table.go
```diff
@@ -296,21 +312,30 @@ func (ctx *Context) Execute(instructions []byte) *InstructionContextIterator {
...
 	}
 
 	if instruction == DW_CFA_nop {
-		return
+		return nil
 	}
 
-	fn := lookupFunc(instruction, ctx)
+	fn, err := lookupFunc(instruction, ctx)
 	fn(ctx)
```
brancz, line 326: always first handle the error then continue

internal/dwarf/frame/table.go
```diff
@@ -296,21 +312,30 @@ func (ctx *Context) Execute(instructions []byte) *InstructionContextIterator {
...
+	if err != nil {
+		return fmt.Errorf(" DWARF CFA rule is not valid. This should never happen :%w", err)
+	}
+	return nil
 }
 
-func lookupFunc(instruction byte, ctx *Context) instruction {
+// getPid gets the Process ID as a string from the fullExecutablePath provided.
+func getPid(fullExecutablePath string) string {
+	return strings.Split(fullExecutablePath, "/")[2]
```
brancz, line 335: this is likely to panic, we already get this information by looking at the path, so I wouldn't try to parse it out of the path again, a human is going to look at this error anyway

internal/dwarf/frame/table.go
```diff
@@ -253,38 +258,49 @@ const low_6_offset = 0x3f
...
-func ExecuteDWARFProgram(fde *FrameDescriptionEntry, context *Context) *InstructionContextIterator {
-	ctx := executeCIEInstructions(fde.CIE, context)
+func ExecuteDWARFProgram(fde *FrameDescriptionEntry, context *Context) (*InstructionContextIterator, error) {
+	ctx, err := executeCIEInstructions(fde.CIE, context)
 	ctx.order = fde.order
 	frame := ctx.currentInstruction()
 	frame.loc = fde.Begin()
-	return ctx.Execute(fde.Instructions)
+	if err != nil {
+		return ctx.Execute(fde.Instructions), err
```
brancz, line 291: why execute in the error case?

internal/dwarf/frame/table.go
```diff
@@ -253,38 +258,49 @@ const low_6_offset = 0x3f
...
+	}
+	return context, nil
 }
 
 // ExecuteDWARFProgram evaluates the unwind opcodes for a function.
-func ExecuteDWARFProgram(fde *FrameDescriptionEntry, context *Context) *InstructionContextIterator {
-	ctx := executeCIEInstructions(fde.CIE, context)
+func ExecuteDWARFProgram(fde *FrameDescriptionEntry, context *Context) (*InstructionContextIterator, error) {
+	ctx, err := executeCIEInstructions(fde.CIE, context)
 	ctx.order = fde.order
```
brancz, line 287: ctx is likely to be nil so this will panic, always handle the error first

internal/dwarf/frame/table.go
Sylfrena: good catch, thanks!

brancz (commented):
Pretty close, let's also add error wrapping in each place where we return an error up the chain, that way we'll have context and know exactly where the error happened.

internal/dwarf/frame/table.go
```diff
@@ -296,21 +309,26 @@ func (ctx *Context) Execute(instructions []byte) *InstructionContextIterator {
...
 
 	if instruction == DW_CFA_nop {
-		return
+		return nil
 	}
 
-	fn := lookupFunc(instruction, ctx)
+	fn, err := lookupFunc(instruction, ctx)
+	if err != nil {
+		return fmt.Errorf(" DWARF CFA rule is not valid. This should never happen :%w", err)
```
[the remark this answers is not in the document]
Sylfrena: good catch, thanks!

brancz (approved):

PR #2493, opened 2024-02-12
Use OS num CPUs instead of runtime schedulable ones

brancz:
### Why?

`runtime.NumCPU()` does not report the present/available CPUs it reports the schedulable ones.

### What?

Read available CPUs from sysfs.

### How?

Using the [`numcpus`](https://github.com/tklauser/numcpus) package.

### Test Plan

Tested locally and it works, and added a metric to report the detected number as well.

### DIFF
--- a/cmd/parca-agent/main.go
+++ b/cmd/parca-agent/main.go
@@ -48,6 +48,7 @@ import (
 	vtproto "github.com/planetscale/vtprotobuf/codec/grpc"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/collectors"
+	"github.com/prometheus/client_golang/prometheus/promauto"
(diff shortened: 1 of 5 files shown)


kakkoyun (approved):

brancz:
Odd, since the library we now use seems to do it correctly: https://github.com/tklauser/numcpus

umanwizard:
@brancz , I was referring to this code: https://github.com/parca-dev/parca-agent/blob/2812709a/pkg/profiler/cpu/cpu.go#L812-L819

Note that if the available CPUs is not a contiguous range from 0 to n, this logic is wrong.

I think this is what's causing https://github.com/parca-dev/parca-agent/issues/2639

brancz:
Ooooooh I see, yes makes sense!

PR #2516, opened 2024-02-16
pkg/analytics: Add support for report arbitrary metrics

metalmatze:
This should be super helpful when looking at the overall performance of the parca-agent across various environments and sizes.

The first set of metrics is still up for discussion.


kakkoyun (approved):

brancz:
I think this facility is nice, but I don't think there are any additional metrics we want to send at the moment.

metalmatze:
This was an experiment. Given there isn't a direct need, let's close for now and see if we ever want to pick it up.

PR #2515, opened 2024-02-16
internal/dwarf/table.go: fix typo

Sylfrena:

PR #2521, opened 2024-02-19
Revert "internal/dwarf/table.go: fix typo"

kakkoyun:
Reverts parca-dev/parca-agent#2515


kakkoyun:
False alert! The issue seems to be caused by some GH action-specific environmental pollution!

PR #2531, opened 2024-02-20
python: Use new initial state API from runtime-data

kakkoyun:
- Use new initial state API
- Use more precise label matches 05d3b97af3d0d759c35e24d7387fd2b78ecf0800


brancz:
Assuming you tested this and the python tests pass lgtm

kakkoyun:
> Assuming you tested this and the python tests pass lgtm

Tests are green. The caching plugin fails. I'll remove it.

PR #2550, opened 2024-02-23
make process filtering work

gnurizen:


kakkoyun (approved):
This looks good to me for now.

I guess the initial intention was to be able to filter by thread id.
But we can figure things out later.

PR #2574, opened 2024-03-01
Add ability to read trace id from goroutine labels of Go processes

brancz:
### Why?

It's already possible to attach trace ids to profiling data with instrumented profilers, but it would be great if we could also create the connection between distributed tracing and profiling data with collection from eBPF. Well this is exactly that.

With the trace ID added to profiled stacks it's possible to view all profiling data of a request in one flamegraph/iciclegraph, which makes it much easier to identify bottlenecks of a request than having to view it on a per process basis.

### What?

Attempt to read the `otel.traceid` goroutine label from Go processes. Plus a flag to enable it, by default this will be disabled as it significantly increases the amount of data produced.

### How?

Go processes store the current goroutine in thread local store. From there this reads the `g` (aka goroutine) struct, then the `m` (the actual operating system thread) of that goroutine, and finally `curg` (current goroutine).

This chain is necessary because `getg().m.curg` points to the current user g assigned to the thread (`curg == getg()` when not on the system stack). `curg` may be nil if there is no user `g`, such as when running in the scheduler.

### Test Plan

Tested with a test binary.

Screenshot 2024-03-01 at 10 11 34

### DIFF
--- a/bpf/unwinders/go_traceid.h
+++ b/bpf/unwinders/go_traceid.h
@@ -1,0 +1,199 @@
+// +build ignore
+// ^^ this is a golang build tag meant to exclude this C file from compilation
+// by the CGO compiler
+//
+// SPDX-License-Identifier: GPL-2.0-only
+// Copyright 2024 The Parca Authors
+
+#include "vmlinux.h"
+#include "basic_types.h"
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include "tls.h"
+
+struct go_string
+{
+    char *str;
+    s64 len;
+};
+
+struct go_slice
+{
+    void *array;
+    s64 len;
+    s64 cap;
+};
+
+struct map_bucket {
+    char tophash[8];
+    struct go_string keys[8];
+    struct go_string values[8];
+    void *overflow;
+};
+
+struct
+{
+    __uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+    __uint(key_size, sizeof(u32));
+    __uint(value_size, sizeof(struct map_bucket));
+    __uint(max_entries, 1);
+} golang_mapbucket_storage_map SEC(".maps");
+
+// length of "otel.traceid" is 12
+#define TRACEID_MAP_KEY_LENGTH 12
+#define TRACEID_MAP_VAL_LENGTH 32
+#define MAX_BUCKETS 8
+
+static __always_inline bool bpf_memcmp(char *s1, char *s2, s32 size)
+{
+	for (int i = 0; i < size; i++)
+	{
+		if (s1[i] != s2[i])
+		{
+			return false;
+		}
+	}
+
+	return true;
+}
+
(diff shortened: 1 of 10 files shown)


bpf/unwinders/go_traceid.h
```diff
@@ -0,0 +1,210 @@
...
+
+    unsigned char log_2_bucket_count;
+    res = bpf_probe_read(&log_2_bucket_count, sizeof(log_2_bucket_count), labels_map_ptr + 9);
+    if (res < 0)
+    {
+        return false;
+    }
+    u64 bucket_count = 1 << log_2_bucket_count;
+    void *label_buckets;
+    res = bpf_probe_read(&label_buckets, sizeof(label_buckets), labels_map_ptr + 16);
```
gnurizen, line 161: Should we declare a map struct and collapse these 3 reads into one?   Would probably turn a 8 byte read, followed by 1 byte read, followed by 8 byte read into a memcpy I'm guessing?  Probably a wash performance wise but maybe smaller code?  Feel free to file under "thing for Tom to investigate".

bpf/unwinders/native.bpf.c
```diff
@@ -1110,6 +1115,22 @@ static __always_inline bool set_initial_state(struct bpf_perf_event_data *ctx) {
...
+  unwind_state->stack_key.trace_id[6] = 0;
+  unwind_state->stack_key.trace_id[7] = 0;
+  unwind_state->stack_key.trace_id[8] = 0;
+  unwind_state->stack_key.trace_id[9] = 0;
+  unwind_state->stack_key.trace_id[10] = 0;
+  unwind_state->stack_key.trace_id[11] = 0;
+  unwind_state->stack_key.trace_id[12] = 0;
+  unwind_state->stack_key.trace_id[13] = 0;
+  unwind_state->stack_key.trace_id[14] = 0;
+  unwind_state->stack_key.trace_id[15] = 0;
```
gnurizen, line 1133: I would think a bzero or memset call would be more efficient here but I haven't investigated it.

gnurizen (approved):
LGTM!

bpf/unwinders/go_traceid.h
```diff
@@ -0,0 +1,210 @@
...
+typedef long long unsigned int __u64;
+
+typedef __s8 s8;
+typedef __u8 u8;
+typedef __s16 s16;
+typedef __u16 u16;
+typedef __s32 s32;
+typedef __u32 u32;
+typedef __s64 s64;
+typedef __u64 u64;
```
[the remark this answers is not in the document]
brancz: Whips, yes! A leftover from my standalone development.

bpf/unwinders/go_traceid.h
brancz: it might be possible to optimize, but the bpf stack is only 512bytes, which already caused me to have to move one of these to be stored in a per-cpu bpf map instead

kakkoyun (approved):

PR #2578, opened 2024-03-01
debuginfo: Change find order

kakkoyun:

### Why?

`.debug` pattern is more likely to exists compared to other candidates.


kakkoyun:
> How did we come to this conclusion?

While working on the runtime-data bits, I had to test and try to extract debug info on various operating systems (debian, ubuntu, alpine, etc.) And I have seen this pattern repeatedly over the alternatives.

brancz:
Got it!

brancz (approved):

PR #2591, opened 2024-03-05
java: Add groundwork for the naive unwinder

kakkoyun:
- **Unify the names for unwinder information**
- **Initialize ~~dtrace~~ java unwinder**
- **Fetch and sync the required initial data for unwinding**


brancz (commented):
I don't think we should call it dtrace, I realize we take plenty of inspiration, but dtrace was not at all java specific.

The other thing I wonder about, is how can this work with a quite typical case of `jvm->java->native`? If we unwind separately how would we know how to jump from the native frames over the jit frames to the JVM frames? I frankly think this is the time where we need to make Java a new "native" unwinder.

kakkoyun:
> I don't think we should call it dtrace, I realize we take plenty of inspiration, but dtrace was not at all java specific.

I'll rename.

> The other thing I wonder about, is how can this work with a quite typical case of `jvm->java->native`? If we unwind separately how would we know how to jump from the native frames over the jit frames to the JVM frames? I frankly think this is the time where we need to make Java a new "native" unwinder and not the same kind as python or ruby (even these we should eventually convert, this was the whole point of why we introduced the per-frame switching).

I agree. Again, this is not even close to the finish line. I'm trying to reduce the number of unknowns. As we did with mode switching, we should switch between unwinders for Java.

WIP.

PR #2596, opened 2024-03-05
Fix benign but bogus main executable assumption

gnurizen:


kakkoyun (approved):
LGTM 🚀 

Great find! Why the PR is draft though?

PR #2604, opened 2024-03-07
Various fixes for aarch64

umanwizard:
### Why?

I was trying to run parca-agent on my new work laptop which has Fedora 39 in a VM on a mac host, i.e. aarch64. I encountered various problems trying to do so, which this PR fixes.

Probably the reason nobody has fixed this before is that Fedora is building their OS using some special compiler flags for security hardening.

Note that this PR contains two logically distinct changes in separate commits (plus one commit for a minor logging fix).

### What?

Please see the individual commit messages for details.

### How?

Please see the individual commit messages for details.

### Test Plan

parca-agent works on my machine now and generates reasonable-looking output.

### DIFF
--- a/bpf/unwinders/native.bpf.c
+++ b/bpf/unwinders/native.bpf.c
@@ -724,6 +724,20 @@ static __always_inline void add_frame(unwind_state_t *unwind_state, u64 frame) {
   }
 }
 
+#if __TARGET_ARCH_arm64
+static __always_inline u64 canonicalize_addr(u64 addr) {
+  // aarch64 has a 48-bit address space; one bit (in position 56)
+  // indicates whether it points into kernel or user space.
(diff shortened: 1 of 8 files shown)


kakkoyun (approved):
Awesome findings and fixes, thanks a lot!

Also I love the comments. It made the review a piece of cake 😀 

I wonder if our arm64 tests will pass now.

We have been chasing a sneaky issue that causes unwinding to fail.

kakkoyun:
It fixes Ruby failure on arm64 🙌🙌 There are still some more issues we need to track down for some Python versions.

umanwizard:
Thanks for the quick review. I pushed a new version which should get rid of the unused function warning on x86.

kakkoyun (approved):

PR #2625, opened 2024-03-18
unwinder/native: Per mapping unwinder switching (including runtime unwinders)

kakkoyun:

TBD

There are still a couple of more things that I'd like to add.


gnurizen:
Is there a discussion somewhere of what this is fixing?  For the record I think mixing non-functional cosmetic changes with functional changes is bad practice and just makes things harder on reviewers.

kakkoyun:
> Is there a discussion somewhere of what this is fixing? For the record I think mixing non-functional cosmetic changes with functional changes is bad practice and just makes things harder on reviewers.

"Cosmetic" changes are my attempt to make the code easy to reason about myself. It started as finding some bugs in the JIT unwinding phase. And then I realized some things that we might need to do better. This requires a lot of tests (which I plan to add). I would be happy to split the PR into easy-to-ingest smaller PRs, but first, I need to finish and ensure the behavior I intended works as expected.

However, since I don't work on the project full-time anymore, the progress will come slower. I'd be happy to discuss my intentions online during office hours, and we can come up with a plan together.

PR #2636, opened 2024-03-21
Fix function end detection

umanwizard:
When constructing unwind tables, we add a synthetic entry marking the end of functions. This serves to allow the unwinder to detect program counters that fall between two functions. We only insert this synthetic row if it's possible to give it a program counter that doesn't overlap with anything else; that is, if there is a gap between the end of the function in question and the beginning of the next.

It's fine to omit the synthetic entries in the case where there is no gap, because in those cases, there isn't any way the PC can lie between to functions.

However, currently we _also_ use the synthetic entries for another purpose: determining where it's okay to break the unwind table into shards (as it's not acceptable for a function to cross a shard boundary). This is imprecise, because the synthetic entries are not added in all cases, as described above. Indeed, binaries that have no gaps in a long enough sequence of functions can't be split anywhere and thus fail to be profiled.

This commit fixes that issue by looking at the original FDEs (which have actual function begin and end information) directly, rather than trying to find function boundaries by using the synthetic rows. It also factors out the chunk-splitting logic into a separate function so it can be tested.

### DIFF
--- a/pkg/profiler/cpu/bpf/maps/maps.go
+++ b/pkg/profiler/cpu/bpf/maps/maps.go
@@ -26,6 +26,7 @@ import (
 	"os"
 	"path"
 	goruntime "runtime"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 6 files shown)


pkg/profiler/cpu/bpf/maps/maps.go
```diff
@@ -1821,6 +1823,44 @@ func (m *Maps) allocateNewShard() error {
 	return nil
 }
 
+// Get the largest chunk of `ut` (up to `maxLen`) that
```
gnurizen, line 1826: FYI some linters require the first word of a function doc comment to be the function name itself, apparently we don't have such a linter which surprises me.

gnurizen (approved):
Nice work!

Issue #2639, opened 2024-03-23
parca-agent does not run in systems with offline CPUs


ekotik:
**Describe the bug**

Perf could not procceed to open offline CPUs (disabled by kernel, system or user) and parca-agent stops its execution
even it could continue to work with other CPUs (online cores).

**To Reproduce**
1. Try one of the following
-  echo 0 > /sys/devices/kernel/cpu/cpuX/online
- boot linux kernel with `mds=full,nosmt` parameter
2. run agent
3. check if parca-agent terminates with error

**Expected behavior**
1. parca-agent does not terminate
3. there is no "open perf event" error in log

**Logs**
`level=error name=parca-agent ts=2024-03-22T17:32:05.924131353Z caller=main.go:539 err="open perf event: no such device"
`

**Software (please complete the following information):**
 - Parca Agent Version: `<= 1b3ce753740834c2370b7e42c281798f65df51d9`

**Environment (please complete the following information):**
- Linux Distribution: `debian`
- Linux Version: `6.7.9-amd64`
- Arch: `amd64`

**Additional context**
Basic patch and some details of this issue is in #2638.

PR #2646, opened 2024-03-26
Only attempt to profile online CPUs

umanwizard:
Previously we counted the number of present CPUs, e.g. n, and then turned on all CPUs from 0 to n-1 inclusive.

This has two defects:

1. It counts present CPUs, not online CPUs, whereas we don't want to attempt to profile an offline-but-present CPU.
2. It doesn't account for the possiblity that there can be gaps in the set of CPUs; e.g., the ranges might be 0-11,13-15 if cpu 12 is offline.

numcpus doesn't have a public API to get the set of online CPUs, so let's implement it ourselves (based on reading numcpus internal code).

### Why?

Fixes https://github.com/parca-dev/parca-agent/issues/2639

### Test Plan
`sudo chcpu -d 12 && sudo dist/parca-agent --log-level=debug` and manually verify that all CPUs except 12 are profiled. Before this would exit with an error.

### DIFF
--- a/pkg/profiler/cpu/cpu.go
+++ b/pkg/profiler/cpu/cpu.go
@@ -135,6 +135,8 @@ type CPU struct {
 	lastSuccessfulProfileStartedAt time.Time
 	lastProfileStartedAt           time.Time
 	objFilePool                    *objectfile.Pool
+
+	cpus cpuinfo.CPUSet
 }
 
 func NewCPUProfiler(
@@ -147,6 +149,7 @@ func NewCPUProfiler(
 	config *Config,
 	bpfProgramLoaded chan bool,
 	objFilePool *objectfile.Pool,
+	cpus cpuinfo.CPUSet,
 ) *CPU {
 	return &CPU{
 		config: config,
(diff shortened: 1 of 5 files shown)


pkg/profiler/cpu/cpu.go
```diff
@@ -803,49 +807,48 @@ func (p *CPU) Run(ctx context.Context) error {
...
-			Type:   unix.PERF_TYPE_SOFTWARE,
-			Config: unix.PERF_COUNT_SW_CPU_CLOCK,
-			Size:   uint32(unsafe.Sizeof(unix.PerfEventAttr{})),
-			Sample: p.config.ProfilingSamplingFrequency,
-			Bits:   unix.PerfBitDisabled | unix.PerfBitFreq,
-		}, -1 /* pid */, i /* cpu id */, -1 /* group */, 0 /* flags */)
-		if err != nil {
-			return fmt.Errorf("open perf event: %w", err)
-		}
+	for _, cpuRange := range p.cpus {
```
umanwizard, line 812: Review this file with "hide whitespace changes" on

kakkoyun (approved):

PR #2649, opened 2024-03-27
Surface unwind failure reasons in status page

umanwizard:
Each time we return from the eBPF program without a stack, bump a counter describing the reason for doing so. In user space, every time through the profiling loop, slurp these counters and surface them in the status page.

My hope is that this will be a first step towards helping us debug why some processes are not being unwound in some user environments.

### Why?

We are repeatedly hearing from users that some processes are not being profiled in various situations and they don't know why. It's a bit hard to figure this out from just the `/metrics` output, because that is not broken down per process.

### Test Plan

Tested by looking at the status page locally

### DIFF
--- a/bpf/unwinders/native.bpf.c
+++ b/bpf/unwinders/native.bpf.c
@@ -253,7 +253,9 @@ typedef struct {
     u32 unsupported_fp_action;
     u32 unsupported_cfa;
     u32 truncated;
-    u32 catchall;  // TODO more granular?
+    u32 previous_rsp_zero;
+    u32 previous_rip_zero;
+    u32 previous_rbp_zero;
     u32 internal_error;
 } unwind_failed_reasons_t;
 
@@ -1005,7 +1007,7 @@ int native_unwind(struct bpf_perf_event_data *ctx) {
         if (previous_rsp == 0) {
             LOG("[error] previous_rsp should not be zero.");
             bump_unwind_error_catchall();
-            BUMP_UNWIND_FAILED_COUNT(per_process_id, catchall);
+            BUMP_UNWIND_FAILED_COUNT(per_process_id, previous_rsp_zero);
             return 1;
         }
 
@@ -1045,13 +1047,14 @@ int native_unwind(struct bpf_perf_event_data *ctx) {
                 LOG("[warn] mapping not added yet");
                 request_refresh_process_info(ctx, user_pid);
 
+                BUMP_UNWIND_FAILED_COUNT(per_process_id, mapping_not_found);
                 bump_unwind_error_jit_unupdated_mapping();
                 return 1;
             }
 
             LOG("[error] previous_rip should not be zero. This can mean that the read failed, ret=%d while reading previous_rip_addr", err);
             bump_unwind_error_catchall();
-            BUMP_UNWIND_FAILED_COUNT(per_process_id, catchall);
+            BUMP_UNWIND_FAILED_COUNT(per_process_id, previous_rip_zero);
             return 1;
         }
 
@@ -1068,7 +1071,7 @@ int native_unwind(struct bpf_perf_event_data *ctx) {
                     "that the read has failed %d.",
                     ret);
                 bump_unwind_error_catchall();
-                BUMP_UNWIND_FAILED_COUNT(per_process_id, catchall);
+                BUMP_UNWIND_FAILED_COUNT(per_process_id, previous_rbp_zero);
                 return 1;
             }
         }
--- a/bpf/unwinders/native.bpf.c
+++ b/bpf/unwinders/native.bpf.c
@@ -240,6 +240,23 @@ typedef struct {
     stack_unwind_row_t rows[MAX_UNWIND_TABLE_SIZE];
 } stack_unwind_table_t;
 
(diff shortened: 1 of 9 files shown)


pkg/profiler/cpu/bpf/maps/maps.go
```diff
@@ -1485,6 +1505,82 @@ func (m *Maps) RefreshProcessInfo(pid int, shouldUseFPByDefault bool) {
...
+		var pid int32
+		if err := binary.Read(bytes.NewBuffer(key), m.byteOrder, &pid); err != nil {
+			return nil, err
+		}
+		val, err := m.unwindFailedReasons.GetValue(unsafe.Pointer(&key[0]))
+		if err != nil {
+			return nil, err
+		}
+		var reasons profiler.UnwindFailedReasons
+		if err := binary.Read(bytes.NewBuffer(val), m.byteOrder, &reasons.PcNotCovered); err != nil {
```
gnurizen, line 1522: I'd just create one reader and use it for all calls to binary.Read.  Also I'd probably use reflection in a loop to set the values instead of hand writing each one.

pkg/profiler/profiler.go
```diff
@@ -35,6 +35,45 @@ type StackID struct {
 	TID PID
 }
 
+
+// TODO[btv]
```
gnurizen, line 38: I would think Go should be the source of truth and go generate should be used to generate a header.  But others have probably trodden this path, I wonder what cilium/ebpf does.

bpf/unwinders/native.bpf.c
```diff
@@ -240,6 +240,23 @@ typedef struct {
...
 
+typedef struct {
+    u32 pc_not_covered;
+    u32 no_unwind_info;
+    u32 missed_filter;
+    u32 mapping_not_found;
+    u32 chunk_not_found;
+    u32 null_unwind_table;
+    u32 table_not_found;
+    u32 rbp_failed;
```
gnurizen, line 251: This name probably shouldn't be architecture specific.

pkg/profiler/cpu/bpf/maps/maps.go
umanwizard: I didn't know Go supported reflection, that is definitely way nicer than what I have here (and also the string generation stuff in main.go). Will restructure.

bpf/unwinders/native.bpf.c
umanwizard: I wanted to keep things as close as possible to the log messages / comments in the unwinder. We can clean up all of them at the same time in a separate PR, I think.

gnurizen (approved):

pkg/profiler/cpu/cpu.go
```diff
@@ -195,6 +196,12 @@ func (p *CPU) ProcessLastErrors() map[int]error {
 	return p.processLastErrors
 }
 
+func (p *CPU) FailedReasons() map[int]profiler.UnwindFailedReasons {
```
brancz, line 199: maybe just a quick comment that the key is the PID?

cmd/parca-agent/main.go
```diff
@@ -1032,6 +1036,25 @@ func run(logger log.Logger, reg *prometheus.Registry, flags flags, numCPU int) e
 					default:
 						profilingStatus = profilerStatusInactive
 					}
+					failedReasons := unwindFailedReasons[prflr.Name()][pid]
+					failedReasonsStrs := make([]string, 0)
+					v := reflect.ValueOf(failedReasons)
```
brancz, line 1041: I think the idiomatic thing in Go here is to list all fields individually, I realize what you're trying to do here, but I think most people reading this stumble over it.

pkg/profiler/cpu/bpf/maps/maps.go
```diff
@@ -1485,6 +1506,36 @@ func (m *Maps) RefreshProcessInfo(pid int, shouldUseFPByDefault bool) {
...
+		if err != nil {
+			return nil, err
+		}
+
+		buf := bytes.NewBuffer(val)
+		reasons := new(profiler.UnwindFailedReasons)
+		v := reflect.ValueOf(reasons)
+		for i := 0; i < v.Elem().NumField(); i++ {
+			fv := v.Elem().Field(i)
+			if err := binary.Read(buf, m.byteOrder, fv.Addr().Interface()); err != nil {
```
brancz, line 1529: You can just call a single `binary.Read` on the whole struct, no need for reflection here.

brancz (commented):
Loooooooove what this will enable!

cmd/parca-agent/main.go
umanwizard: To clarify, are you suggesting I do it like I had in a previous commit (see [here](https://github.com/parca-dev/parca-agent/blob/77aadf96f5086dd6130925e115120c513f9bc7ac/cmd/parca-agent/main.go#L1040)) ? Or are you suggesting something else?

I'm a Go novice, so I'm happy to do whatever you think is the most idiomatic.

brancz: Yeah I think that would be the typical way to write this in Go.

PR #2667, opened 2024-04-04
Fix unwinder loading on 6.8 kernels

umanwizard:
See comments, and mailing list thread linked therein, for details. Before this change, recent kernels (6.8 and later) cannot verify the native unwinder, because improvements in the precision of register bounds tracking cause an explosion in state space resulting in exceeding the limit of 1 million instructions processed per load.

This commit works around that issue, as the comment in the code explains.

### DIFF
--- a/bpf/unwinders/native.bpf.c
+++ b/bpf/unwinders/native.bpf.c
@@ -379,7 +379,6 @@ DEFINE_COUNTER(total_filter_misses);
 static __always_inline u32 opaquify(u32 val) {
     // We use inline asm to make sure clang doesn't optimize it out
     asm volatile(
-        /* "%0 = %1\n" */
         "%0 ^= 0xffffffff\n"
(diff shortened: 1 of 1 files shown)


kakkoyun (approved):
LGTM 🚀 

Wow, interesting.

Happy to merge when CI is green!

kakkoyun:
Is the failure of the native unwinder integration tests on arm64 expected? That should have worked, but the rest could fail.

umanwizard:
No, it's not expected. I need to look into that, thanks.

umanwizard:
OK, this finally. @kakkoyun and @brancz , can you please have another look?

We can't use function calls; everything has to be inlined in order to work on old kernels. But if we let clang/llvm inline the function, it will optimize it away...

I solved this problem by just using inline volatile asm, which the compiler won't optimize. It's an ugly solution, but it seems to work.

PR #2682, opened 2024-04-10
Add "unwind failed" synthetic stacks

umanwizard:
For each unwind failed reason, add a fake stack with two frames: one for a fake function named "unwind failed" and the other for a fake function whose name describes the failure reason.

This will let users see the true CPU usage of their processes in the UI, even when unwinding failed for many stacks. It will also give people an idea that something is going wrong without having to look at the status page.

### Test Plan

pprof from my machine while compiling some rust software: https://pprof.me/96be0e68ef27e71e3621c163c02e2e2b

### DIFF
--- a/pkg/pprof/pprof.go
+++ b/pkg/pprof/pprof.go
@@ -34,6 +34,7 @@ import (
 	"github.com/parca-dev/parca-agent/pkg/perf"
 	"github.com/parca-dev/parca-agent/pkg/process"
 	"github.com/parca-dev/parca-agent/pkg/profile"
+	"github.com/parca-dev/parca-agent/pkg/profiler"
 	"github.com/parca-dev/parca-agent/pkg/symtab"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/pprof/pprof.go
```diff
@@ -271,6 +272,61 @@ func (c *Converter) Convert(ctx context.Context, rawData []profile.RawSample) (*
...
+	var unwindFailedLocation *pprofprofile.Location
+	var unwindFailedMapping *pprofprofile.Mapping
+	addUnwindFailed := func(name string, value uint32) {
+		if value == 0 {
+			return
+		}
+		if unwindFailedLocation == nil {
+			unwindFailedMapping = &pprofprofile.Mapping{
+				ID: uint64(len(c.result.Mapping)) + 1,
+				// File: "unwind_failed",
```
umanwizard, line 284:
```suggestion
```

gnurizen:
LGTM!

brancz (approved):
Very cool!

brancz:
Looks like the integration tests need to be adapted.

PR #2686, opened 2024-04-11
Avoid filling uploadJob channel w/ NoopDebugInfoManager

gnurizen:


kakkoyun (approved):

pkg/process/info.go
```diff
@@ -369,6 +369,10 @@ func (im *InfoManager) ensureDebuginfoUploaded(ctx context.Context, mappings Map
 			continue
 		}
 
+		if shouldInitiateUpload, _ := im.debuginfoManager.ShouldInitiateUpload(ctx, m.BuildID); !shouldInitiateUpload {
```
brancz, line 372: This would cause a network request in a non-testing setting, which we want to avoid. Are we sure this part is necessary in addition to the other change?

brancz (commented):
lgtm

PR #2688, opened 2024-04-11
Fix bug where event processors exit on error

umanwizard:
When there are errors in processing `prefetch` events, the goroutine that processes them would exit silently. If this happened to all the workers, the entire agent would seize up as no more events could be processed.
This commit fixes that behavior.

### Why?
It's easy to reproduce this if I run something that creates a lot of short-lived processes (e.g. C/C++/Rust compilation). The agent eventually gets into a permanently broken state.

### DIFF
--- a/pkg/profiler/cpu/cpu.go
+++ b/pkg/profiler/cpu/cpu.go
@@ -456,19 +456,21 @@ func (p *CPU) listenEvents(ctx context.Context, eventsChan <-chan []byte, lostCh
 				select {
 				case pid, open := <-prefetch:
 					if !open {
+						level.Info(p.logger).Log("msg", "event loop ended, ending worker loop")
 						return
 					}
 					_ = p.prefetchProcessInfo(ctx, pid)
 					fetchInProgress.Delete(pid)
 				case pid, open := <-refresh:
 					if !open {
+						level.Info(p.logger).Log("msg", "event loop ended, ending worker loop")
 						return
 					}
 
-					func() {
+					err := func() error {
 						defer refreshInProgress.Delete(pid)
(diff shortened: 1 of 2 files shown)


pkg/profiler/cpu/cpu.go
```diff
@@ -464,25 +464,28 @@ func (p *CPU) listenEvents(ctx context.Context, eventsChan <-chan []byte, lostCh
...
 
-					executable := fmt.Sprintf("/proc/%d/exe", pid)
-					shouldUseFPByDefault, err := p.framePointerCache.HasFramePointers(executable) // nolint:contextcheck
-					if err != nil {
-						// It might not exist as reading procfs is racy. If the executable has no symbols
-						// that we use as a heuristic to detect whether it has frame pointers or not,
-						// we assume it does not and that we should generate the unwind information.
-						level.Debug(p.logger).Log("msg", "frame pointer detection failed", "executable", executable, "err", err)
-						if !errors.Is(err, os.ErrNotExist) && !errors.Is(err, elf.ErrNoSymbols) {
+					func() {
```
brancz, line 468: Ultimately it doesn't matter, but below we fix this by using `continue` statements. Whichever way we prefer, I think we should be consistent.

gnurizen: I would have func return an error and log it, the err != nil { return } pattern should be avoided I think.  Swallowing errors like that is no beuno.

brancz: I see! Nevermind then, lgtm

umanwizard:
@kakkoyun 
> Do these errors are regular ones of panics?

They are regular errors
> Also why don't we add the mentioned metrics in this PR?

I have now added them

gnurizen (approved):

PR #2697, opened 2024-04-12
Fix off-by-one error in chunk splitting

umanwizard:
If the PC of the "end-of-function" marker was exactly one greater than the PC of the last real table entry, and if the unwind entries of the function (including the end-of-function marker) were too big to fit in the current chunk also by exactly one, then we would erroneously take the entire function _except_ the marker, and start the next chunk with the marker. This commit fixes that defect.

### Test Plan
Added a test for this case to maps_test.go


brancz (approved):
Nice!!

PR #2717, opened 2024-04-17
Use both .eh_frame and .debug_frame

umanwizard:
.debug_frame is quite similar to .eh_frame, but some binaries have one and not the other, or have a mix of things in both. In this commit, we construct things from both.

We don't have much experience with our .debug_frame parser in the real world, so for now if anything is correpted or fails parsing that one, we just fall back to the old behavior of using only .eh_frame and log a warning / bump a metric. After we gain some more confidence, we can reconsider that logic.

### Why?
Various old versions of gcc apparently didn't put everything in .eh_frame by default, at least on aarch64. This seems to be why our python tests are failing.

### Test Plan
Not tested yet, but after we fix the issue with failing to unwind past main due to missing dwarf info, this should improve the state of aarch64 python integration tests.

### DIFF
--- a/cmd/eh-frame/main.go
+++ b/cmd/eh-frame/main.go
@@ -64,7 +64,8 @@ func main() {
 	}
 
 	if flags.Final {
-		ut, arch, _, err := unwind.GenerateCompactUnwindTable(file, logger, prometheus.NewCounter(prometheus.CounterOpts{}))
+		g := unwind.NewCompactUnwindTableGenerator(logger, prometheus.NewRegistry())
+		ut, arch, _, err := g.Gen(file)
 		if err != nil {
 			// nolint:forbidigo
 			fmt.Println("failed with:", err)
--- a/cmd/eh-frame/main.go
+++ b/cmd/eh-frame/main.go
@@ -64,7 +64,7 @@ func main() {
 	}
 
 	if flags.Final {
-		ut, arch, _, err := unwind.GenerateCompactUnwindTable(file)
+		ut, arch, _, err := unwind.GenerateCompactUnwindTable(file, logger, prometheus.NewCounter(prometheus.CounterOpts{}))
 		if err != nil {
 			// nolint:forbidigo
 			fmt.Println("failed with:", err)
@@ -79,7 +79,7 @@ func main() {
 	}
 
 	ptb := unwind.NewUnwindTableBuilder(logger)
-	if err := ptb.PrintTable(os.Stdout, file, flags.Compact, pc); err != nil {
+	if err := ptb.PrintTable(os.Stdout, file, flags.Compact, pc, true); err != nil {
 		// nolint:forbidigo
 		fmt.Println("failed with:", err)
 	}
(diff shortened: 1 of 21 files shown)


cmd/eh-frame/main.go
```diff
@@ -64,7 +64,7 @@ func main() {
 	}
 
 	if flags.Final {
-		ut, arch, _, err := unwind.GenerateCompactUnwindTable(file)
+		ut, arch, _, err := unwind.GenerateCompactUnwindTable(file, logger, prometheus.NewCounter(prometheus.CounterOpts{}))
```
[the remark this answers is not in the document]
kakkoyun: Rather than passing a no-op counter, I would prefer to pass a no-op registerer.

I think we need a new type `CompactUnwindGenerator` and that can hold on to logger and register and crate it's own metrics.

pkg/stack/unwind/compact_unwind_table.go
```diff
@@ -290,23 +295,71 @@ func CompactUnwindTableRepresentation(unwindTable UnwindTable, arch elf.Machine)
...
 	var ut CompactUnwindTable
+	isUnexpected := func(err error) bool {
+		return err != nil && !errors.Is(err, ErrEhFrameSectionNotFound) && !errors.Is(err, ErrNoFDEsFound)
+	}
 
 	// Fetch FDEs.
-	fdes, arch, err := ReadFDEs(file)
-	if err != nil {
-		return ut, arch, fdes, err
+	ehFrameFdes, arch, err := ReadFDEs(file, true)
```
gnurizen, line 305: I'm not a fan of boolean parameters.  Also I think the extra .debug_frame reading should be buried below the ReadFDEs function.   Seems unnecessary to complicate higher level layers with that complexity.

pkg/stack/unwind/unwind_table.go
```diff
@@ -199,29 +199,38 @@ func (ptb *UnwindTableBuilder) PrintTable(writer io.Writer, file *objectfile.Obj
...
 	if err != nil {
-		return nil, arch, fmt.Errorf("failed to read .eh_frame section: %w", err)
+		return nil, arch, fmt.Errorf("failed to read %s section: %w", secName, err)
 	}
 
 	// TODO: Byte order of a DWARF section can be different.
-	fdes, err := frame.Parse(ehFrame, obj.ByteOrder, 0, pointerSize(obj.Machine), sec.Addr, arch)
+	var ehFrameAddr uint64
+	if ehFrame {
+		ehFrameAddr = sec.Addr
```
gnurizen, line 231: Interesting!   My musl changes don't do this, I wonder if .debug_frame's in .debug files are different, what clued you in that you needed to do this?

gnurizen: I included this change in my musl stuff and the eh-frame output didn't change, probably good to include but apparently the .debug_frame's in the musl libc doesn't include anything different that the eh/debug conditionals trigger on.

umanwizard:
I've updated the PR to use Go runtime VDSO data for unwinding from within VDSO stacks.

umanwizard:
don't merge, this still is broken somehow

brancz (approved):

gnurizen (approved):

PR #2721, opened 2024-04-18
remove no-op Close statement

umanwizard:
We always use `elf.NewFile` to open elf files, so closing them is always a no-op.


kakkoyun (approved):
Awesome 🏆 🥇 💯

https://github.com/parca-dev/parca-agent/blob/main/pkg/objectfile/object_file.go

The close logic in objectfile remnants of an archaic attempt to implement reference counting for these files. Feel free to remove the whole implementation.

The goal was to prevent leaking file descriptors. When I say leak not in the sense of runtime leaking those FDs. Th Go runtime is pretty good at it. 

We were just holding to FDs too long, more than we should be. I think this is not the case anymore. But still approach with caution.

PR #2732, opened 2024-04-22
Fix pointer authentication for FP-based unwinding

umanwizard:
This was missed in 872d82a740f8ceff4fce806d4cfa8daea83e5395. We need similar logic for FP unwinding as we already added there for DWARF unwinding.

See that commit message for details about what this is.


kakkoyun (approved):

kakkoyun:
I just realized native mixed mode testing fails on arm https://github.com/parca-dev/parca-agent/actions/runs/8791927660/job/24127129336

PR #2739, opened 2024-04-25
Fix verification on 6.8 kernels (second try).

umanwizard:
See PR #2667 for details about why this is needed, as well as the mailing list thread
https://lore.kernel.org/bpf/874jci5l3f.fsf@taipei.mail-host-address-is-not-set/ .

The approach tried in that PR caused a regression on older kernels; hopefully, this one will work while also being less confusing.

### Test Plan
See if it passes CI


brancz:
Looks like all ARM64 integration tests fail with this.

umanwizard:
apparently old kernel versions don't support function calls at all, so this is a non-starter.

PR #2748, opened 2024-04-30
Dynamically discover offsets for finding custom labels map

umanwizard:
This reads the DWARF info of a Go program to discover the offsets of the map containing custom pprof labels, and passes them to the unwinder, rather than hardcoding them manually. As before, the unwinder walks this label map to discover the OTEL thread ID, if it has been saved.

Future work will include extending this to Rust as well as making it work in more exotic situations (e.g. when the Go runtime has been loaded from a shared library rather than being part of the main executable)

### DIFF
--- a/bpf/unwinders/go_traceid.h
+++ b/bpf/unwinders/go_traceid.h
@@ -36,6 +36,16 @@ struct {
     __uint(max_entries, 1);
 } golang_mapbucket_storage_map SEC(".maps");
 
+typedef struct {
+    u32 m;
+    u32 curg;
+    u32 labels;
+
+    u32 hmap_count;
+    u32 hmap_log2_bucket_count;
+    u32 hmap_buckets;
+} go_custom_label_offsets_t;
+
 // length of "otel.traceid" is 12
 #define TRACEID_MAP_KEY_LENGTH 12
 #define TRACEID_MAP_VAL_LENGTH 32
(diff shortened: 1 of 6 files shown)


brancz:
Since for go there is only one compiler and these are runtime structs, I think we should actually just extract these once, store the results in [runtime data](https://github.com/parca-dev/runtime-data) and based on the Go version decide which offsets to load. The nice thing about that is that debugger-level debuginfo don't have to be present in order to do this, we just need to find which Go version a binary is.

gnurizen:
> The nice thing about that is that debugger-level debuginfo don't have to be present in order to do this, we just need to find which Go version a binary is (which we already do anyway).

Can a go binary be stripped of this information?  I think it can but I'm not sure.  I almost wonder if we want the best of both worlds, dynamic lookup when dwarf symbols are present to support new releases w/o having to re-release our software w/ runtime-data to fallback on when dwarf symbols aren't present.  For reference luajit release binaries are fully stripped and dwarf lookup would be a non-starter if we needed to dynamically lookup field offsets (which so far I haven't needed but suspect might be necessary down the road).

brancz:
Yeah Go binaries don't even have the DWARF required for this enabled by default, you need to pass special flags for it. The only thing that cannot be stripped from Go binaries is the gopclntab section.

I like the idea of a hybrid approach here, we use the pre-extracted offsets for known versions and for custom Go builds (which do exist) or those that we don't have offsets for yet, we fall back to trying to extract it from debuginfo.

gnurizen:
> Yeah Go binaries don't even have the DWARF required for this enabled by default, you need to pass special flags for it. The only thing that cannot be stripped from Go binaries is the gopclntab section.

Wait a minute, if go binaries don't have this information then I think we are doing it wrong.  The Go binary has to know where pointers live in the "g" struct or the garbage collector couldn't work.  I _think_ maybe we need to get this information from pclntab?  

This project seems interesting: https://github.com/mandiant/GoReSym

I also have a faint recollection that dlv can debug stripped binaries because it also uses the pclntab to figure out this stuff.

I think we should fully understand what the solution space is here before proceeding, right now my thinking is this information has to be in the go binary, we should always be able to get it and we have no need for ahead of time runtime-data for Go (rust/C/C++ may be different).   I'm prepared to be wrong!

brancz:
Closing in favor of https://github.com/parca-dev/parca-agent/pull/2808

PR #2761, opened 2024-05-07
Properly get return address for leaf frames on ARM during syscalls

umanwizard:
In the leaf frame on ARM, we read the return address directly from x30. But this is wrong during syscalls, where we need to read it from the user task state, like we do for all the other registers during initial startup.

To test, run this program, which spins with high CPU in kernel mode:

```c
// kernspin.c
#include <unistd.h>

int main() {
    while (1) {
        (void)getpid();  // Repeatedly call getpid to consume CPU in kernel mode
    }
    return 0;
}
```


brancz (approved):
Nice find!!

PR #2792, opened 2024-05-16
don't hang on panic in Run()

umanwizard:
Right now we just hang forever waiting on this wg if we panic anywhere under `Run`, so the panic will never be reported, the process will never be restarted, etc.

To fix this, re-raise the panic and don't call `wg.Wait()` if we're panicking.


gnurizen:
I don't get it, why wouldn't the panic blow right passed the Wait()?   Is there another recover involved?

umanwizard:
@gnurizen Defers do still run when unwinding the stack for a panic.

gnurizen:
Ah that makes sense, I would have looked at fixing it by removing the defer, calling Wait in a defer seems questionable.

PR #2794, opened 2024-05-16
Fix Python version detection

umanwizard:
We were looking for the .bss section by parsing the ELF file and getting its address. Unfortunately, we used the `Offset` field instead, which points to a section's file offset rather than its location in memory. These can be the same, but don't have to be.

This was causing python version detection to break in at least some cases.

With this change, I am able to profile Python programs run from the python-test:3.7.17-slim container, which didn't work before.


brancz (approved):

kakkoyun:
Good catch!

PR #2799, opened 2024-05-17
feat: add procfs cmdline

dreamerlzl:
### Why?
Add an extra label for distinguishing processes, especially for java processes where the `comm`s are the same.

### What?
Add a procfs label cmdline from `/proc/[pid]/cmdline`

### How?
The code is short and self-explanatory.

### Test Plan
See the modified integration test.

### DIFF
--- a/cmd/parca-agent/main.go
+++ b/cmd/parca-agent/main.go
@@ -185,7 +185,7 @@ type FlagsMetadata struct {
 	ContainerRuntimeSocketPath string            `help:"The filesystem path to the container runtimes socket. Leave this empty to use the defaults."`
 
 	DisableCaching       bool `default:"false" help:"Disable caching of metadata."`
-	EnableProcessCmdline bool `default:"false" help:"add /proc/[pid]/cmdline as a label."`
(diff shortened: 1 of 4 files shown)


kakkoyun (approved):

dreamerlzl:
Wondering whether the profiler timeouts for Python and Java are not enough... seeing similar check failures in other PRs.

brancz:
I'm worried that this can easily expose secrets in profiling data. I think we should only expose this with a flag on that is off by default.

cmd/parca-agent/main.go
```diff
@@ -184,7 +184,8 @@ type FlagsMetadata struct {
 	ExternalLabels             map[string]string `help:"Label(s) to attach to all profiles."`
 	ContainerRuntimeSocketPath string            `help:"The filesystem path to the container runtimes socket. Leave this empty to use the defaults."`
 
-	DisableCaching bool `default:"false" help:"Disable caching of metadata."`
+	DisableCaching       bool `default:"false" help:"Disable caching of metadata."`
+	EnableProcessCmdline bool `default:"false" help:"add /proc/[pid]/cmdline as a label."`
```
brancz, line 188: "add" -> "Add" and please add a note about the potential security implications.

brancz (approved):

PR #2811, opened 2024-05-21
fix duplicated leaf frame for dwarf unwinding

umanwizard:
We already record the leaf frame in `set_initial_state`, so we don't need to record it again during the main unwinding loop.

The FP unwinder handles this by recording the frame that was reached in each iteration of the loop, whereas the DWARF unwinder records the frame from the _previous_ iteration of the loop. This PR changes the DWARF unwinder to match the behavior of the FP unwinder, fixing the issue.

FYI @metalmatze


metalmatze (approved):
Nice find!

PR #2833, opened 2024-05-30
Collect custom labels

umanwizard:
This causes the agent to collect arbitrary Go trace IDs, not just `otel.traceid`.

The maximum size of keys and values is 64, and the maximum number of custom labels is 16. The maximum number of unique sets of custom labels is 1000, which we store in a map by their hash. Thus the total increase in memory usage for the native unwinder is ~2MB.

### DIFF
--- a/bpf/Makefile
+++ b/bpf/Makefile
@@ -24,7 +24,7 @@ BPF_BUNDLE := $(OUT_DIR)/parca-agent.bpf.tar.gz
 LIBBPF_HEADERS := $(OUT_DIR)/libbpf/$(ARCH)/usr/include
 
 VMLINUX_INCLUDE_PATH := $(SHORT_ARCH)
-BPF_SRC := unwinders/native.bpf.c unwinders/go_runtime.h unwinders/hash.h unwinders/common.h unwinders/shared.h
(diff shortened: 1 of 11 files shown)


bpf/Makefile
```diff
@@ -24,7 +24,7 @@ BPF_BUNDLE := $(OUT_DIR)/parca-agent.bpf.tar.gz
 LIBBPF_HEADERS := $(OUT_DIR)/libbpf/$(ARCH)/usr/include
 
 VMLINUX_INCLUDE_PATH := $(SHORT_ARCH)
-BPF_SRC := unwinders/native.bpf.c unwinders/go_runtime.h
+BPF_SRC := unwinders/native.bpf.c unwinders/go_runtime.h unwinders/hash.h unwinders/common.h unwinders/shared.h
```
gnurizen, line 27: Maybe this should be 

`BPF_SRC := unwinders/native.bpf.c $BPF_HEADERS`

And we should move BPF_HEADERS up.  Or the way the rules are written looks like BPF_SRC doesn't need to include the headers.

gnurizen (approved):
Looks good to me, we should probably have an integration test for this so it doesn't break again.

PR #2851, opened 2024-06-04
Remove synthetic unwinder error frames from agent and instead generate them in the unwinder with more context

gnurizen:
The main benefit here is getting line numbers from the errors, there is roughly a couple dozen error sites so having line numbers is nice.  The symbol "class" is the error message, a little clunky but works.  Additionally we add the samples from exitted processes now instead of throwing them away.

### DIFF
--- a/bpf/unwinders/native.bpf.c
+++ b/bpf/unwinders/native.bpf.c
@@ -948,6 +948,7 @@ int native_unwind(struct bpf_perf_event_data *ctx) {
             LOG("[warn] mapping not found");
             request_refresh_process_info(ctx, per_process_id);
             BUMP_UNWIND_FAILED_COUNT(per_process_id, mapping_not_found);
+            ERROR_SAMPLE(unwind_state, "mapping not found");
             return 1;
         } else if (unwind_table_result == FIND_UNWIND_CHUNK_NOT_FOUND || unwind_table_result == FIND_UNWIND_CHUNK_NOT_FOUND_FOR_PC) {
             if (proc_info->should_use_fp_by_default) {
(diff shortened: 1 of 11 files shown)


pkg/pprof/pprof.go
```diff
@@ -437,9 +394,20 @@ func (c *Converter) AddUnwinderInfoLocation(frameID uint64) *pprofprofile.Locati
 		return l
 	}
 
+	mapping := c.interpreterMapping
+	for idx, prog := range bpfprograms.ProgNames {
+		if base, found := strings.CutSuffix(interpreterSymbol.Filename, ".c"); found {
+			_, file := path.Split(base)
+			if strings.HasSuffix(prog, file+".o") {
+				mapping = c.bpfProgMappings[idx]
```
gnurizen, line 402: This is kinda hacky, open to suggestions.

bpf/unwinders/native.bpf.c
```diff
@@ -948,6 +948,7 @@ int native_unwind(struct bpf_perf_event_data *ctx) {
             LOG("[warn] mapping not found");
             request_refresh_process_info(ctx, per_process_id);
             BUMP_UNWIND_FAILED_COUNT(per_process_id, mapping_not_found);
```
umanwizard, line 950: We can probably get rid of `BUMP_UNWIND_FAILED_COUNT`, `failed_unwind_reasons`, etc. now

pkg/pprof/pprof.go
umanwizard: We use `StackFrame.Status` in `profile.go` to track truncated frames today. You could add a new possible value for bpf error frames, then pass it through to here.

bpf/unwinders/native.bpf.c
gnurizen: I was thinking these could still be useful for surveillance purposes (ie if we got them exported into prometheus).

umanwizard: We already have a bunch of Prometheus metrics from the unwinder, which doesn't work via this `BUMP_UNWIND_FAILED_COUNT` mechanism. See e.g. `unwinder_stats_t` and the `DEFINE_COUNTER` mechanism.

umanwizard (approved):

PR #2857, opened 2024-06-07
memset entire unwind_state

gnurizen:
Properly memset the entire unwind_state.  This in theory should improve our stack hit rate as uninitialized fields wont introduce noise into the hashing.  This by itself isn't fixing any bugs.


umanwizard (approved):
Good catch!

PR #2877, opened 2024-06-13
fix incorrect zeroing of unwind_state

umanwizard:
### Why?
The original code was wrong for a few reasons (most importantly using `sizeof(ptr)` instead of `sizeof(*ptr)`).

Fixing it causes clang to replace the entire loop with a call to `__builtin_memset`, which is broken for bpf, so we write our own `zero_mem` function that can't be optimized in that way.


umanwizard:
Closing in favor of https://github.com/parca-dev/parca-agent/pull/2878

PR #2878, opened 2024-06-13
Revert "memset the entire unwind_state to make sure no"

gnurizen:
This reverts commit cce2bd4700764fe242e44a30ee1bda0ac55eec4a.

Instead use the bpf_perf_prog_read_value trick for all large memzeros

### DIFF
--- a/bpf/unwinders/shared.h
+++ b/bpf/unwinders/shared.h
@@ -135,3 +135,13 @@ static __always_inline void *bpf_map_lookup_or_try_init(void *map, const void *k
       }                                                                                                                                                        \
     }                                                                                                                                                          \
   })
+
+
+// HACK: On failure, bpf_perf_prog_read_value() zeroes the buffer. We ensure that this always
+// fail with a compile time assert that ensures that the struct size is different to the size
+// of the expected structure.
+#define bpf_large_memzero(_d, _l)                                                                                        \
+    ({                                                                                                                   \
+        _Static_assert(_l != sizeof(struct bpf_perf_event_value), "stack size must be different to the valid argument"); \
+        bpf_perf_prog_read_value(ctx, _d, _l);                                                                           \
+    })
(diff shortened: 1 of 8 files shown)


umanwizard (approved):

bpf/unwinders/shared.h
```diff
@@ -10,128 +10,140 @@
...
+// fail with a compile time assert that ensures that the struct size is different to the size
+// of the expected structure.
+//
+// By zeroing the stack we will ensure that stack aggregates work more effectively as otherwise
+// previous values past the stack length will hash the stack to a different value in the map.
+#define bpf_large_memzero(_d, _l)                                                                                        \
+    ({                                                                                                                   \
+        _Static_assert(_l != sizeof(struct bpf_perf_event_value), "stack size must be different to the valid argument"); \
+        bpf_perf_prog_read_value(ctx, _d, _l);                                                                           \
+    })
```
umanwizard, line 149: You could compute the size in this macro (e.g. `u64 _l = sizeof(*_d);`) rather than forcing it to be passed in.

bpf/unwinders/shared.h
```diff
@@ -10,128 +10,140 @@
...
+            }                                                                                         \
+        }                                                                                             \
+    })
+
+// HACK: On failure, bpf_perf_prog_read_value() zeroes the buffer. We ensure that this always
+// fail with a compile time assert that ensures that the struct size is different to the size
+// of the expected structure.
+//
+// By zeroing the stack we will ensure that stack aggregates work more effectively as otherwise
+// previous values past the stack length will hash the stack to a different value in the map.
```
umanwizard, line 144: This comment seems misplaced now

umanwizard (approved):
LGTM, just two minor nits

bpf/unwinders/shared.h
gnurizen: I'll leave it like this in the hopes that it gets changed to some compiler/bpf helper in the future.

bpf/unwinders/shared.h
gnurizen: Good point, fixed.

PR #2886, opened 2024-06-20
fix x86 verification issues with custom labels

umanwizard:
Rewrite a problematic section of code in assembly language to force the verifier to accept our program.

Otherwise, clang does something with this code (on x86) that is too clever for the verifier to understand.


bpf/unwinders/common.h
```diff
@@ -73,7 +77,7 @@ typedef struct {
         aggregate_stacks();                                                           \
     })
 
-// These must be divisible by 8
+// These must be powers of 2 and divisible by 8
```
umanwizard, line 80:
```suggestion
// These must be divisible by 8
```

brancz (approved):

PR #2887, opened 2024-06-20
fix wrong computation of current goroutine in kernel code on arm

umanwizard:
If we are in kernel code, the registers will have the kernel values of various registers, rather than the user-mode ones. We need to get x28 (the current goroutine register) at the beginning of unwinding the same way we do for other registers.


gnurizen (approved):
lgtm

PR #2891, opened 2024-06-24
update clang to v18

umanwizard:
Update clang version.

The changes to rbperf.h are due to clang-format changes.


umanwizard:
Shelving this for now. We need to wait for https://github.com/KyleMayes/install-llvm-action/issues/75 to be fixed.

PR #2892, opened 2024-06-24
Increase warning level when failing to fetch unwinder info

umanwizard:
If this happens, it's a real issue that can break a lot of functionality, so we want to see it in the logs


brancz (approved):

umanwizard:
this ends up being pretty noisy, so I changed my mind about landing it

umanwizard:
I changed my mind again... this shouldn't be _too_ noisy in the steady state. We can re-evaluate if it proves to be too noisy later.

PR #2895, opened 2024-06-27
Fix various issues with debuginfo uploading

umanwizard:
See individual commit messages for details.


pkg/process/info.go
```diff
@@ -108,6 +109,10 @@ func newMetrics(reg prometheus.Registerer) *metrics {
 			Help:                        "Duration of metadata fetches.",
 			NativeHistogramBucketFactor: 1.1,
 		}),
+		uploadBacklog: promauto.With(reg).NewGauge(prometheus.GaugeOpts{
```
brancz, line 112: I think a gauge func would be more appropriate.

brancz:
Glen other than the comment feel free to merge once changed

PR #2958, opened 2024-08-08
Parca Agent as extension to open-telemetry/opentelemetry-ebpf-profiler

brancz:
As discussed on the maintainers mailing list, this is the Parca Agent based on the [open-telemetry/opentelemetry-ebpf-profiler](https://github.com/open-telemetry/opentelemetry-ebpf-profiler).

The projects have identical goals, so it makes sense to just have everyone pull on the same string and keep the tight integration and extensions that are special to the Parca project as a thick wrapper around it.

This change includes several large changes, most prominently:

* **Apache Arrow-based protocol**: A new Parca-native Apache Arrow-based protocol that is significantly more efficient than what was previously used. It is cheaper to produce, cheaper to consume by a server, and less bytes on the wire.
* **Metadata relabeling**: It introduces Prometheus-style relabeling for metadata. Parca Agent no longer adds all metadata it can find, which previously resulted in overwhelming amounts of metadata automatically being attached to profiling data. Using relabeling users can choose exactly what metadata is valuable to them, and also use relabeling to selectively choose which profiling data to potentially drop.

While the Parca Agent already supported native unwinding with [DWARF](https://www.polarsignals.com/blog/posts/2022/11/29/dwarf-based-stack-walking-using-ebpf) (x86 and [arm64](https://www.polarsignals.com/blog/posts/2023/10/17/profiling-arm64-with-ebpf-in-parca-agent)), [Python, Ruby](https://www.polarsignals.com/blog/posts/2023/10/04/profiling-python-and-ruby-with-ebpf), and many more, by using the opentelemetry-ebpf-profiler under the hood, it gains JVM, PHP, Perl, and .NET support. We've also been working on [LuaJIT support](https://github.com/parca-dev/parca-agent/pull/2784), which we plan to upstream to the opentelemetry-ebpf-profiler. We're also working on upstreaming our custom labels support, which among other things, [allows correlating with distributed tracing data](https://www.polarsignals.com/blog/posts/2024/03/05/correlating-tracing-with-profiling-using-ebpf).

Since this has already been discussed on the maintainers mailing list, I'm going to merge this right away, and then we'll work on finishing everything up for a release.


bobrik:
It seems that labeling docs predate this change:

* https://www.parca.dev/docs/parca-agent-labelling

> Using relabeling users can choose exactly what metadata is valuable to them, and also use relabeling to selectively choose which profiling data to potentially drop.

It's unclear how to do this exactly. From what I see, `parca-agent` only pushes `comm` and `node`.

PR #2969, opened 2024-08-13
build: reintroduce snap packaging, test, and release flows

jnsgruk:
This PR reintroduces the snap package, and the test/release workflow.

Unfortunately, we cannot use the plain `goreleaser` setup for building/releasing the snap, as the project is currently using a Goreleaser Docker image, which is unable to run `snapcraft`.

That said, it makes this job slightly easier, as much of what was in place before will still work. There are a couple of places where we might need to tweak artifact paths and suchlike, but otherwise this should work.

@brancz - at the moment I've tagged on the snap build, and test, and release into one workflow to avoid too much duplication, but we can adjust if you'd prefer.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -89,6 +89,14 @@ jobs:
           path: |
             deploy/manifests.tar.gz
 
+      - name: Archive generated artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: parca-agent-dist-release
+          if-no-files-found: error
+          path: |
+            dist/
+
       - name: Release
         uses: softprops/action-gh-release@v0.1.15
         if: startsWith(github.ref, 'refs/tags/')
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -98,6 +106,126 @@ jobs:
...
+        with:
+          name: built-snaps
+
+      - name: Install snapcraft
+        run: |
+          sudo snap install snapcraft --classic --channel=8.x/stable
+
+      - name: Release to latest/edge
+        env:
+          SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.SNAPCRAFT_STORE_CREDENTIALS }}
```
jnsgruk, line 224: We need to make sure this is still set - though I'm guessing it's probably still in the secrets config for the repo

brancz: yep, it is!

brancz (approved):

brancz:
Thank you very much!!

PR #2970, opened 2024-08-13
Support TID

umanwizard:
Bump to our fork of opentelemetry-ebpf-profiler that supports thread IDs, and add the thread ID as a label.

Previously we were just getting the `comm` of the first thread we saw a sample for and saving that as the entire process's `comm`. This PR saves that as `thread_name` instead and gets the main thread's comm (i.e. process name) from procfs, saving that as `comm` instead.

### DIFF
--- a/reporter/parca_reporter.go
+++ b/reporter/parca_reporter.go
@@ -223,7 +223,7 @@ func (r *ParcaReporter) labelsForKey(key labelsKey) labelRetrievalResult {
 
 	lb := &labels.Builder{}
 	lb.Set("node", r.nodeName)
-	lb.Set("comm", key.comm)
+	lb.Set("thread_name", key.comm)
 	lb.Set("tid", fmt.Sprint(key.tid))
 	r.addMetadataForPID(key.pid, lb)
 
--- a/reporter/parca_reporter.go
+++ b/reporter/parca_reporter.go
@@ -10,8 +10,10 @@ import (
 	"bytes"
 	"context"
 	"debug/elf"
+	"encoding/binary"
 	"errors"
 	"fmt"
+	"hash/fnv"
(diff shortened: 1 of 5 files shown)


reporter/parca_reporter.go
```diff
@@ -69,6 +72,31 @@ type stack struct {
 	frameTypes []libpf.FrameType
 }
 
+type labelsKey struct {
+	pid util.PID
+	tid util.PID
+	comm string
```
brancz, line 78: does comm really need to be a part of this key, isn't pid+tid already uniquely identifying the thread?

reporter/parca_reporter.go
```diff
@@ -182,15 +216,16 @@ func (r *ParcaReporter) addMetadataForPID(pid util.PID, lb *labels.Builder) {
...
+	if labels, exists := r.labels.Get(key); exists {
 		return labels
 	}
 
 	lb := &labels.Builder{}
 	lb.Set("node", r.nodeName)
-	lb.Set("comm", comm)
-	r.addMetadataForPID(pid, lb)
+	lb.Set("thread_name", key.comm)
+	lb.Set("tid", fmt.Sprint(key.tid))
```
brancz, line 227: If we make these on by default as opposed to opt-in via relabeling, then we should probably also have the `pid` by default.

Also, can you add documentation about these here: https://github.com/parca-dev/parca-agent?tab=readme-ov-file#metadata-labels

reporter/parca_reporter.go
umanwizard: yes, you're right.

However, we should still not cache comm, since it can change (e.g. if `pthread_setname_np` is called). But it doesn't need to be part of the key.

PR #2972, opened 2024-08-26
goreleaser: Don't prefix docker image with v

metalmatze:
The version should have the v as part of the name for a release.

Additionally renaming the deprecated name_template to version_template.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -78,7 +78,7 @@ release:
   prerelease: auto
   draft: false
 dockers:
-- image_templates: ["ghcr.io/parca-dev/{{ .ProjectName }}:v{{ .Version }}-amd64"]
+- image_templates: ["ghcr.io/parca-dev/{{ .ProjectName }}:{{ .Version }}-amd64"]
   dockerfile: Dockerfile
   use: buildx
   build_flag_templates:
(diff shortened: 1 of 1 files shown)


metalmatze:
Hey @jnsgruk, the builds are failing for snap. What is the best way to resolve this for snaps? 
Cannot pack snap: error: cannot validate snap "parca-agent": invalid snap version "goreleaser-docker-version-d12fff49": cannot be longer than 32 characters (got: 34)
```

kakkoyun (approved):

jnsgruk:
> Hey @jnsgruk, the builds are failing for snap. What is the best way to resolve this for snaps? 
> 
> ```
> 
> Cannot pack snap: error: cannot validate snap "parca-agent": invalid snap version "goreleaser-docker-version-d12fff49": cannot be longer than 32 characters (got: 34)
> 
> ```

It's because I've got snapcraft.yaml to grab the version from the manifest, which in this case is a big long string.

It should be safe to merge, in that it'll work fine for actual releases - I can try and make the logic a bit cleverer so it just uses the short hash as the version in cases where the commit isn't tagged at some point

jnsgruk:
Other option is to have goreleaser configured to do neater versions for unreleased stuff like "0.32.0-next" or "0.34.1-deadbeef"

PR #2973, opened 2024-08-28
Point to latest version of our otel profiler fork

umanwizard:
There are several innocuous changes, but please take a look at the comment beginning `!!FIXME!!`. We will have to address that soon, I think.


umanwizard:
Don't merge this yet -- it's not uploading any profiles when I run it locally; something must be wrong.

umanwizard:
Now ready for a look.

reporter/parca_reporter.go
```diff
@@ -244,7 +241,9 @@ func (r *ParcaReporter) ExecutableMetadata(ctx context.Context,
 	}
 
 	// Always attempt to upload, the uploader is responsible for deduplication.
-	r.uploader.Upload(ctx, fileID, buildID, open)
+	// !! FIXME !! Apparently uploading stuff in this function is a "really bad idea";
+	// see https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/126#pullrequestreview-2248554710
```
brancz, line 245: The recommendation is to specifically not *block* this Go routine, but instead open the file and pass the file handle to an asynchronous mechanism. Apart from immediately opening the file, that's exactly how our mechanism works, so we can remove the context here without a problem.

brancz (approved):

reporter/parca_reporter.go
umanwizard, line 245:
```suggestion
```

PR #2983, opened 2024-09-13
Search for k8s containers by name, not position in array.

umanwizard:
It is not necessarily true that the items in ContainerStatuses are in the same order as those in Spec.Containers, so the old logic was wrong.

container names _are_ unique within a pod, so we can find the container in the spec for a given ID based on its name.


brancz (approved):
Nice find!!

PR #2987, opened 2024-09-19
fix(snap): ensure arm64 build can be performed from an amd64 machine

jnsgruk:
The primary reason for this PR is to address the [failing snap release job](https://github.com/parca-dev/parca-agent/actions/runs/10943881277/job/30384811867).

This job fails because there is no `arm64` build to upload, despite the build artifacts being fetched from the previous steps in the process.

Further investigation showed that the `snapcraft pack` for the `arm64` snap [finished abruptly](https://github.com/parca-dev/parca-agent/actions/runs/10943881277/job/30384611839#step:6:48) having not actually generated an artefact. This turns out to be a [known bug](https://github.com/canonical/craft-application/issues/225), which should be addressed in the next snapcraft release. In the mean time, we can work around by just explicitly stating that the `arm64` snap can be `built-on` an `amd64` machine.

As a drive-by, https://github.com/parca-dev/parca-agent/commit/a3fefff51e11603f5f5451832c14b92919324eb3 ensures that when there is a tag present in the goreleaser `metadata.json,` it's used to set the version of the snap accordingly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all crossbuild build build-debug
+.PHONY: all crossbuild build build-debug snap
 
 all: crossbuild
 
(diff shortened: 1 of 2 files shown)


jnsgruk:
[This line](https://github.com/parca-dev/parca-agent/actions/runs/10945230563/job/30389113505#step:6:83) is the indicator that the build fix has worked :)

thorfour (approved):

PR #2990, opened 2024-10-08
Allow running without syscall tracepoints

umanwizard:
Currently we fail to run if tracepoints can't be set on various sub-events of /sys/kernel/debug/tracing/events/syscalls/ . Currently these are used to detect various offsets when BTF is not available.

Since some kernel configurations don't have syscall tracepoints, but _do_ have BTF info, things might still work if we just keep running here.

I have confirmed that with this patch, the agent starts up and seems to work normally on a kernel without CONFIG_FTRACE_SYSCALLS, which fails without this patch.

### DIFF
--- a/flags/flags.go
+++ b/flags/flags.go
@@ -123,6 +123,7 @@ type Flags struct {
 	Hidden    FlagsHidden    `embed:"" hidden:""           prefix:""`
 
 	BPF FlagsBPF `embed:"" prefix:"bpf-"`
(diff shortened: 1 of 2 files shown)


umanwizard:
The Otel agent has exactly the same issue: https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/173

We don't necessarily need to do the same thing as them, since this is entirely controlled within our own `main` function. But just in case, let's wait for the outcome of that discussion before merging this, in case I missed something.

brancz (approved):

umanwizard:
There were no strong objections on the OTel issue so I think this can be merged.

PR #2993, opened 2024-10-09
Ensure nulls are written for non-present labels

umanwizard:
When we write some labels for a trace, we need to make sure to write nulls for all the labels that are _not_ present, so the arrow arrays are still the correct length.


umanwizard:
Marking as draft for now as there's some doubt about whether this is correct

PR #3000, opened 2024-11-13
Set CGO_ENABLED in build

umanwizard:
Without this flag, Go will silently determine whether to enable Cgo or not based on whether a program called "gcc" exists in the PATH. This is a bad default that leads to very difficult-to-understand error messages when it's not available.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,10 +14,10 @@ crossbuild:
 		release --snapshot --clean --skip=publish --verbose
 
 build:
-	go build -o parca-agent -buildvcs=false -ldflags="-extldflags=-static" -tags osusergo,netgo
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -14,10 +14,10 @@ crossbuild:
 		release --snapshot --clean --skip=publish --verbose
 
 build:
-	go build -o parca-agent -buildvcs=false -ldflags="-extldflags=-static" -tags osusergo,netgo
+	CGO_ENABLED=1 go build -o parca-agent -buildvcs=false -ldflags="-extldflags=-static" -tags osusergo,netgo
```
gnurizen, line 17: Should we just export CGO_ENABLED = 1 higher up?

PR #3009, opened 2024-12-16
docs: fix README links leading nowhere

BunningsWarehouseOfficial:


BunningsWarehouseOfficial:
The website at https://www.parca.dev/docs/parca-agent/ is also giving an inconsistent message as to what the default sampling rate is:
> Parca Agent is an always-on sampling profiler that uses [eBPF](https://ebpf.io/) to capture raw profiling data with very low overhead. It observes user-space, and kernel-space stack traces 100 times per second...

https://www.parca.dev/docs/parca-agent-design/#cpu-sampling-frequency
> We sample at 19Hz (19 times per second)...

BunningsWarehouseOfficial:
The [Java Support page](https://www.parca.dev/docs/java-support/) also seems out of date given the current status of #1 and my own personal experience playing around with Java profiling.

brancz (approved):
Thank you!

Yes, there are various things out of date, everything you mentioned needs to be fixed in the docs. Want to give us a hand? :)

BunningsWarehouseOfficial:
Yeah sure, I could help out with docs, I just wasn't confident on my ability to do so accurately given that I used Parca for the first time only 2 days ago 😅

PR #3024, opened 2025-01-22
Update parca-agent to include debug tracer

gnurizen:


main.go
```diff
@@ -459,3 +465,49 @@ func maxElementsPerInterval(monitorInterval time.Duration, samplesPerSecond int,
 	presentCPUCores uint16) uint32 {
 	return uint32(samplesPerSecond) * uint32(monitorInterval.Seconds()) * uint32(presentCPUCores)
 }
+
+func getTracePipe() (*os.File, error) {
+	for _, mnt := range []string{
+		"/sys/kernel/debug/tracing",
+		"/sys/kernel/tracing",
+		"/tracing",
+		"/trace"} {
```
umanwizard, line 474: what's the deal with this? Do different systems have it at different places?

main.go
```diff
@@ -459,3 +465,49 @@ func maxElementsPerInterval(monitorInterval time.Duration, samplesPerSecond int,
...
+		tp.Close()
+	}()
+
+	r := bufio.NewReader(tp)
+	for {
+		line, err := r.ReadString('\n')
+		if err != nil {
+			if errors.Is(err, io.EOF) {
+				continue
+			} else {
```
umanwizard, line 503: don't need `else` here since there is a `continue` just before

umanwizard (approved):

main.go
gnurizen: Apparently?   I copied this list from the bpftool code base: https://github.com/libbpf/bpftool/blob/main/src/tracelog.c#L59

main.go
gnurizen: :+1:

PR #3026, opened 2025-01-22
Initial implementation of offline mode

umanwizard:
This patch introduces *offline mode*, a mechanism for running the agent without a network connection.

`--offline-mode-storage-path` takes a path to a directory in which data will be stored; it will be created if it does not exist. Samples and stack traces are written to a file with name `<current time>-<agent PID>.padata`. Every 10 minutes (by default), this file is rotated and compressed with zstd.

`--offline-mode-upload` causes the agent to look for these files in the path passed via `--offline-mode-storage-path`, and upload them to the remote store configured with the usual remote store options.

Note that significant parts of `log_uploader.go` -- specifically the `locationsReader` and associated functions -- were mostly copied from Parca.

### DIFF
--- a/flags/flags.go
+++ b/flags/flags.go
@@ -22,8 +22,8 @@ import (
 	"time"
 
 	"github.com/alecthomas/kong"
-	"go.opentelemetry.io/ebpf-profiler/tracer"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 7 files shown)


brancz (approved):
Very nice!

For a moment I was thinking that maybe we should export the things that were copied from Parca, but this is probably easier so we can evolve the wire protocol without having to first modify Parca.

PR #3048, opened 2025-04-10
Validate that custom labels are utf-8.

umanwizard:
Also, re-truncate a value that the tracer
might have truncated at a byte boundary at a
rune boundary instead.

### DIFF
--- a/reporter/parca_reporter.go
+++ b/reporter/parca_reporter.go
@@ -174,17 +174,17 @@ func (r *ParcaReporter) SupportsReportTraceEvent() bool { return true }
 //
 // It returns the correctly truncated utf-8 string if possible;
 // otherwise "", false.
-func maybeFixTruncation(s string, maxLen uint) (string, bool) {
+func maybeFixTruncation(s string, maxLen int) (string, bool) {
 	if utf8.ValidString(s) {
 		return s, true
 	}
 	// maybe we truncated in the middle of a rune -- if that's the case,
 	// truncate the entire rune.
 	plausibleTruncatedRuneBegin := -1
-	if len(s) == support.CustomLabelMaxValLen {
+	if len(s) == maxLen {
 		i := 0
 		for ; i < 2; i += 1 {
-			idx := support.CustomLabelMaxValLen - i - 1
+			idx := maxLen - i - 1
 			if s[idx]&0xC0 != 0x80 {
 				plausibleTruncatedRuneBegin = idx
 				break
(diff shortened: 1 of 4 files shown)


brancz (approved):

reporter/parca_reporter.go
```diff
@@ -167,6 +169,39 @@ func hashString(s string) uint32 {
...
+// It returns the correctly truncated utf-8 string if possible;
+// otherwise "", false.
+func maybeFixTruncation(s string, maxLen uint) (string, bool) {
+	if utf8.ValidString(s) {
+		return s, true
+	}
+	// maybe we truncated in the middle of a rune -- if that's the case,
+	// truncate the entire rune.
+	plausibleTruncatedRuneBegin := -1
+	if len(s) == support.CustomLabelMaxValLen {
```
gnurizen, line 184: I think this will never be true since we only read CustomLabelMaxValLen-1 in order to have a null terminated C string.

reporter/parca_reporter.go
```diff
@@ -167,6 +169,39 @@ func hashString(s string) uint32 {
...
+		i := 0
+		for ; i < 2; i += 1 {
+			idx := support.CustomLabelMaxValLen - i - 1
```
gnurizen, line 187: I think it may be easier to just use Go's rune iteration to iterate over runes and keep track of length, ie:

```go
len := 0
for _, r := range s {
    len += utf8.RuneLen(r)
}
return s[0:len]
```

reporter/parca_reporter.go
```diff
@@ -196,6 +231,15 @@ func (r *ParcaReporter) ReportTraceEvent(trace *libpf.Trace,
 	}
 
 	for k, v := range trace.CustomLabels {
+		if !utf8.ValidString(k) {
+			log.Warnf("ignoring non-UTF8 label: %s", hex.EncodeToString([]byte(k)))
```
gnurizen, line 235: I think we should warn if any truncation takes place that isn't because we hit the string byte limit.  My concern is we have labels that are complete garbage but we don't warn because there happened to be 1 or 2 valid ascii chars at the beginning.

umanwizard: That is what we're doing. `maybeFixTruncation` doesn't return a valid value unless the truncation happened at the boundary.

reporter/parca_reporter.go
umanwizard: Good call, updated the logic to subtract 1.

reporter/parca_reporter.go
umanwizard: As you said in another comment, we _only_ want to log something valid if the truncation happened at the max len boundary. Which is what this more complex logic does.

PR #3051, opened 2025-04-24
ci(docker): Update build tag to include commit timestamp

metalmatze:
This modifies the build workflow to append the commit timestamp to the generated tag. It ensures better uniqueness and traceability of snapshot image tags.
This can fix https://github.com/parca-dev/demo-deployments/issues/327


metalmatze:
As far as I can tell, this shouldn't impact the other build artifacts.

Issue #3019, opened 2025-01-14
No ttl for label cache


dreamerlzl:
In previous versions of parca-agent (e.g. 0.31.0) there is a ttl for the label cache like 3 * profiling duration(so it's easier to profile specific pods at will by attaching a label to a pod + relabel_configs). Now after refactoring based on otel ebpf profiler, the TTL has been removed. Is this by design?

https://github.com/parca-dev/parca-agent/blob/v0.35.1/reporter/parca_reporter.go#L450

brancz:
No particular reason, we can definitely add TTL again!

dreamerlzl:
https://github.com/parca-dev/parca-agent/pull/3063

PR #3063, opened 2025-06-25
add TTL for label cache

dreamerlzl:
see https://github.com/parca-dev/parca-agent/issues/3019


dreamerlzl:
> If I recall correctly I think we had this cache in the old agent because we wanted to re-resolve metadata every now and then, eg. if an annotation was added to a pod at runtime. I feel like a 15s TTL might be a little too aggressive for this though. I think maybe a 10min TTL is more appropriate as this should really only be necessary quite rarely.

Yeah we have the similar use cases. I have changed the default value to 10min

PR #3064, opened 2025-07-06
add flag for reading pods from kube-apiserver cache

dreamerlzl:
Current listing of pods can go to etcd directly without hitting cache in apiserver. Adding an optional flag to reduce the load on etcd side by setting `resourceVersion=0`.


brancz:
Is there a good reason not to make this default?

Another thing I've recently been thinking about, we already have a trigger where we list pods when we don't have metadata. What if we used the kubelet's local pods endpoint instead of hitting the apiserver at all?

dreamerlzl:
> Is there a good reason not to make this default?
> 
> Another thing I've recently been thinking about, we already have a trigger where we list pods when we don't have metadata. What if we used the kubelet's local pods endpoint instead of hitting the apiserver at all?

I believe that's even better by using kubelet's local pods.

brancz:
Sounds good to me!

dreamerlzl:
Btw, can we merge this pr first?

PR #3068, opened 2025-07-18
Fix goroutines leak at reportDataToBackend

secfree:
Below is the monitor panel of the number of parca-agent goroutines before and after applying this patch.

image


brancz (approved):
Great find! Thank you very much!

PR #3097, opened 2025-09-17
fix: upload size when stripping is disabled

iamKunal:
Same as https://github.com/parca-dev/parca-agent/pull/3031 with fixed merge conflicts


brancz:
Merged the original one, but thank you for bringing this up again!

PR #3148, opened 2026-02-26
Fixing the bug where wrong cpu label values were getting assigned to the samples

manojVivek:
Bug: 
When viewing the data collected from the agent as flamechart, samples grouped by node + cpu, there were overlapping stacks, which cannot happen ideally. 

Root cause: 
The labels lru cache in the `parca-reporter` is only using tid as the cache key and when a thread gets scheduled across different cpu cores across the profiling ticks, we only use the labels cached from the first occurrence of the tid. 
So, if the tid when first seen was on cpu `1`, then it gets reported the same for all future occurrences, even though it can be seen on a different cpu core. 

Fix:
Made `pid` the label cache key and patching cpu, thread_id and thread_name labels on the fly. This reduced the memory footprint of this cache as well. 

Flags:
Added 3 more flags to control these labels, as they don't belong to the relabeling path anymore: 
- `--metadata-disable-cpu-label`
- `--metadata-disable-thread-id-label`
- `--metadata-disable-thread-comm-label`

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,6 +256,9 @@ Parca Agent supports [Prometheus relabeling](https://prometheus.io/docs/promethe
 
 * `node`: The name of the node that the process is running on as specified by the `--node` flag.
 * `comm`: The command name of the process being profiled.
+* `cpu`: The CPU the sample was taken on. Can be disabled with `--metadata-disable-cpu-label`.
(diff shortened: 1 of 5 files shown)


gnurizen:
I looked at this but don't really get what's going on and why the 2nd commit undoes a bunch of stuff in the first commit.  What's the motivation for the 3 new flags?

PR #3168, opened 2026-04-12
oomprof: prefer GNU build IDs for profiles

### LINKED ISSUE #3152: oomprof improvements
`Read record: rpc error: code = internal desc = flight error: externalerror(external(tonic(status { code: unknown, message: "failed to symbolize: get readers: get reader: ensure buildid file: stat cache file: open /var/symbolizer/cache/debuginfo/2be1042d-6479-4b07-be2d-b55fc3629439/ulmkek7612iaaqfbreyp/-4mu379iyuzob0xgpy3t/9kh6jff18qk9d8bvhog6/9qjs2oi1fyl3i_jqzae7\0/optimized.debuginfo.v6: invalid argument", metadata: metadatamap { headers: {"content-type": "application/grpc"} }, source: none })))`

this is an oomprof profile from a customers go 1.26 process
which we couldn't read back
a couple of things .. it seems like this uses the go build ID, not the gnu build ID
and second .. it looks like there's a null-terminator included in the string?

officialasishkumar:
OOMProf can capture Go build IDs with a trailing NUL byte when reading ELF notes directly. Passing those IDs through profile metadata makes Parca construct invalid cache paths during symbolization.

This resolves OOMProf mappings through the executable's GNU build ID when the file is still available, and otherwise falls back to a NUL-trimmed captured build ID. The same normalization is applied to raw OOM profiles and reporter-converted memory samples.

Fixes #3152

Tests:
- go test ./...
- go test -race ./...

### DIFF
--- a/internal/buildid/buildid.go
+++ b/internal/buildid/buildid.go
@@ -1,0 +1,131 @@
+package buildid
+
+import (
+	"bytes"
+	"debug/elf"
+	"encoding/binary"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+)
+
+const (
+	gnuBuildIDNoteName = "GNU\x00"
+	gnuBuildIDNoteType = 3
+)
+
+var ErrNotFound = errors.New("GNU build ID not found")
+
+// Clean treats build IDs as C strings. This removes the trailing NUL byte that
+// Go build IDs can carry when read directly from ELF notes.
+func Clean(buildID string) string {
+	if idx := strings.IndexByte(buildID, 0); idx >= 0 {
+		return buildID[:idx]
+	}
+	return buildID
+}
+
+// Resolve prefers the executable's GNU build ID when the file is available,
+// falling back to a cleaned build ID captured earlier by the caller.
+func Resolve(executablePath, fallback string) string {
+	if executablePath != "" {
+		if buildID, err := GNU(executablePath); err == nil && buildID != "" {
+			return buildID
+		}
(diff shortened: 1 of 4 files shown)


gnurizen:
Thanks for tackling this — the NUL-byte build ID issue is a real problem and the fix in the reporter/oom paths looks correct.
  
 However, the new internal/buildid package reimplements GNU build ID extraction from ELF files which already exists in our dependency go.opentelemetry.io/ebpf-profiler/libpf/pfelf:
 
  - https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/libpf/pfelf/file.go + https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/libpf/pfelf/file.go#L585-L600 does the same thing as your buildid.GNU() — opens an ELF file, finds .note.gnu.build-id (or .notes), and parses the GNU note to return a hex-encoded build ID.
  - The underlying note parsing (getNoteDescBytes / getNoteHexString in pfelf.go) handles the same namesz/descsz/type/alignment logic as your gnuFromNotes().
 
Since parca-agent already depends on this module, we can avoid the duplication and re-use that code.

gnurizen:
Actually thinking about this further the null byte issue should probably be fixed in oomprof not here.   I think when I created issue 3152 I meant to create it in parca-dev/oomprof not here.

gnurizen:
Wanna redo your fix relative to this: https://github.com/parca-dev/oomprof/pull/4     Thanks!

PR #3176, opened 2026-05-07
v8: Fix the broken passthrough of column and buildId fields

manojVivek:


reporter/parca_reporter.go
```diff
@@ -520,10 +520,18 @@ func (r *ParcaReporter) appendLocationV2(frame libpf.Frame) uint32 {
...
+		// the interpreter type (e.g. "v8js").
 		b.locFrameType.AppendString(frame.Type.String())
-		b.locMappingFile.AppendString(frameKind.String())
-		b.locMappingID.AppendNull()
+		if frame.Mapping.Valid() && frame.Mapping.Value().File.Value().GnuBuildID != "" {
+			file := frame.Mapping.Value().File.Value()
+			b.locMappingFile.AppendString(file.FileName.String())
+			b.locMappingID.AppendString(file.GnuBuildID)
+		} else {
+			b.locMappingFile.AppendString(frameKind.String())
```
brancz: I don't see how this is helpful, why not just append null?

reporter/parca_reporter.go
```diff
@@ -520,10 +520,18 @@ func (r *ParcaReporter) appendLocationV2(frame libpf.Frame) uint32 {
...
-		// Interpreted frames (Python, Ruby, etc.)
+		// Interpreted frames: surface the bundle FileName + GnuBuildID
+		// when present (to support sourcemap resolution, etc.), else fall back to
+		// the interpreter type (e.g. "v8js").
 		b.locFrameType.AppendString(frame.Type.String())
-		b.locMappingFile.AppendString(frameKind.String())
-		b.locMappingID.AppendNull()
+		if frame.Mapping.Valid() && frame.Mapping.Value().File.Value().GnuBuildID != "" {
+			file := frame.Mapping.Value().File.Value()
+			b.locMappingFile.AppendString(file.FileName.String())
```
brancz: I don't understand why this is helpful, why not just append null?

manojVivek: True this is not really helpful except to show the file name as binary on tooltip, which I guess must be the intention in the v1 schema path.

reporter/parca_reporter.go
manojVivek: Same as below, can remove if you feely strongly against it.

brancz: yeah let's append null in both cases

manojVivek: done.

brancz (approved):

PR #3181, opened 2026-05-14
probes: BPF uprobe service for entry/exit duration tracking via OTLP logs

gnurizen:
## Summary

Adds a generic eBPF uprobe service that attaches paired entry/exit probes to declared symbols and emits per-call duration records through the OTLP logs pipeline introduced in #3190.

The motivating use case is finding long JS execution blocks in Node.js applications. A 200 ms callback hogging the libuv event loop is exactly the "phantom" tail latency that doesn't show up in CPU profiles. The reference config attaches to `node::InternalCallbackScope`'s ctor + dtor, brackets each outer libuv callback, and emits one record per scope close with a precise duration measured in BPF.

## What's added

- **`probes/` package**:
  - YAML config schema: `{id, file_match (regex), entry_symbol, exit_symbol, main_thread_only, min_duration_ms}`.
  - eBPF uprobe program (paired entry + exit) with a per-tid scope counter that emits only when the outer scope closes. Nested scopes roll into the outer's measured duration. Main-thread filter via `tid == tgid` directly in BPF; libuv worker-pool / V8 background threads are silently dropped.
  - User-space drains the BPF ringbuf and emits one OTel `log.Record` per completed outer scope via `Logger("parca-agent.probes")`, so consumers can filter on `attributes_scope.name` to slice probe records vs other agent logs.
- **`--probe-config <path>`** CLI flag (default: disabled). Requires a remote-store.
- **Reference `probes/testdata/probes.yaml.sample`** with the Node.js `InternalCallbackScope` ctor + dtor pair pre-verified on Node v18.20 / v22.17 / v24.4 (same mangled symbol on all three, exported via `.dynsym` so stripped builds work too).
- **Build system**: single arch-agnostic `probe.bpf.o` (no `PT_REGS_*` macros used). `Makefile` + `.goreleaser.yml` `before:` hook + CI workflow + `probes/bpf/README.md` set up to produce it on every release.

## Data on the wire

Each probe-fire record:

| Field | Value |
|---|---|
| `body` | `"node.callback_scope"` (stable; queryable) |
| `timestamp` | exit ktime → unix ns via shared `times.KTime` offset (aligned with CPU sample timestamps; range queries over `[start_ns, end_ns]` against the profile-samples table don't drift) |
| `attributes` | `start_ns`, `end_ns`, `duration_ns`, `pid`, `tid`, `comm`, `is_main`, `spec_id`, `probe_id`, `level` |
| `resource` | inherited from the OTel SDK `LoggerProvider`: `service.name`, `service.version`, `host.name` |
| `scope` | `"parca-agent.probes"` |

## Test plan

- [x] `make probes-bpf` produces `probes/bpf/probe.bpf.o`
- [x] `go build ./...` clean
- [x] `go test ./probes/... ./reporter/... -count=1` — all pass
- [x] Local end-to-end on linux/amd64 against PolarSignals backend: ran `node blocker.js` doing synthetic 200 ms blocks. Agent attached, fired 142 times over 8 seconds. Duration histogram: 112 sub-ms callbacks (libuv noise), 27 records at 199 ms and 3 at 200 ms (the blocker). Exactly the expected signal.
- [x] Reviewer: verify on an arm64 build. BPF bytecode is arch-agnostic but exercising the cross-compile pipeline this PR sets up is worthwhile.

## Stacked on

Was stacked on #3190 (OTel logs support), which has since merged. This branch is now rebased directly on top of main.


.github/workflows/build.yml
```diff
@@ -37,6 +37,15 @@ jobs:
         with:
           go-version: '~1.25'
 
+      # The goreleaser-cross image is Debian-based and ships clang, but not
+      # libbpf-dev or linux-libc-dev. Install them so the goreleaser
+      # `before:` hook (make probes-bpf-all) can compile the probes BPF
+      # objects that //go:embed bpf needs.
```
umanwizard: I don't think the comment needs to be this verbose.
```suggestion
      # needed for building the BPF objects
```

flags/flags.go
```diff
@@ -164,6 +164,8 @@ type Flags struct {
 	MergeGpuProfiles bool `default:"false" help:"Report GPU kernel timing and GPU PC sampling under a single gpu_time/nanoseconds sample_type, differentiated by a gpu_view label (pc_sample|kernel_time). When false (the default), they are reported as separate sample_types (gpu_kernel_time/nanoseconds and gpu_pcsample/count) with no per-sample labels."`
 
 	OTLPLogging bool `default:"false" help:"Forward parca-agent's own logrus output to the remote-store as OTLP log records (in addition to local stderr). Requires a remote-store; ignored in offline mode."`
+
+	ProbeConfig string `default:"" help:"Path to a YAML file declaring uprobe attachments. When set, parca-agent attaches a uprobe per matching binary and streams probe-fire events to the configured remote-store as OTLP/Arrow logs. Empty disables the feature."`
```
umanwizard: The name `ProbeConfigPath` would be more consistent with other options (e.g. `ConfigPath`).

probes/bpf/probe.bpf.c
```diff
@@ -0,0 +1,151 @@
...
+// tracking the lifetime of the OUTERMOST scope only. probe_entry captures
+// a timestamp when the outer scope opens (top transitions 0->1) and just
+// increments a counter for nested entries. probe_exit decrements; emits a
+// ringbuf event only when the outer scope closes (top transitions back to
+// 0). Inner scopes roll into the outer's measured duration -- they do not
+// get their own records.
+//
+// Cookie layout (set by user-space, same on both attach points):
+//   bits 63..32 : spec_id  (uint32, 1-based)
+//   bits 31..1  : min_duration_ms (31 bits)
```
umanwizard, line 15: 2^31 ms is ~24 days; we probably don't need 31 bits for this field. It's not harming anything, but just a thought in case we'd rather reserve some of the bits or use them for something else.

probes/bpf/probe.bpf.c
```diff
@@ -0,0 +1,151 @@
...
+};
+
+struct probe_event {
+	__u64 ktime_ns;     // exit ktime (CLOCK_MONOTONIC) of the outer scope
+	__u64 duration_ns;  // exit_ktime - entry_ktime of the outer scope
+	__u32 pid;          // tgid
+	__u32 tid;
+	char  comm[16];
+	__u32 spec_id;
+	__u32 is_main;      // 1 iff tid == tgid
```
umanwizard: Why do we need this when we already have `pid` and `tid` ?

probes/service.go
```diff
@@ -0,0 +1,238 @@
...
+		record.SetSeverity(severity)
+		record.SetSeverityText(levelText)
+		record.AddAttributes(
+			otellog.Int64("start_ns", startNs),
+			otellog.Int64("end_ns", endNs),
+			otellog.Int64("duration_ns", int64(raw.DurationNs)),
+			otellog.Int64("pid", int64(raw.PID)),
+			otellog.Int64("tid", int64(raw.TID)),
+			otellog.String("comm", trimComm(raw.Comm[:])),
+			otellog.Int64("is_main", int64(raw.IsMain)),
```
umanwizard: can we just set this to `raw.PID == raw.TID` rather than needing it to be a field on `raw`?

probes/attach.go
```diff
@@ -0,0 +1,154 @@
...
+type attachReq struct {
+	filePath string
+	fileID   libpf.FileID
+	specs    []ProbeSpec
+}
+
+// attacher owns the state for matching new binaries against the spec list and
+// attaching paired entry/exit uprobes to them on a worker goroutine. The
+// hot-path callback only holds the mutex long enough to dedupe by FileID and
+// run regexes.
```
umanwizard: This is inaccurate: regex are run outside the mutex in the below code.

probes/attach.go
```diff
@@ -0,0 +1,154 @@
...
+	for _, s := range a.specs {
+		if s.FileMatchRE.MatchString(filePath) {
+			matched = append(matched, s)
+		}
+	}
+	if len(matched) == 0 {
+		log.Debugf("probes: received %s (fileID=%s) — no spec matched", filePath, fileID.StringNoQuotes())
+		return
+	}
+	log.Debugf("probes: received %s (fileID=%s) — %d spec(s) matched", filePath, fileID.StringNoQuotes(), len(matched))
```
umanwizard: I have a slight preference against using non-ASCII punctuation in log messages; it makes it more annoying to grep for them.

probes/attach.go
```diff
@@ -0,0 +1,154 @@
...
+
+	select {
+	case a.queue <- attachReq{filePath: filePath, fileID: fileID, specs: matched}:
+	default:
+		// Queue full: log and forget. We've already marked this fileID as
+		// "seen" so we won't try again, which is fine for v1 — the user
+		// can restart with a smaller probe-config or a deeper queue.
+		log.Warnf("probes: attach queue full, dropping %s (fileID=%s)", filePath, fileID.StringNoQuotes())
```
umanwizard: I think this is a pretty big footgun, no? We know a lot of users aren't regularly alerting on logs.

probes/attach.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+}
+
+// run is the attachWorker goroutine. Returns when the queue is drained after
+// ctx is cancelled.
```
umanwizard, line 83: It doesn't actually seem to drain the queue on cancellation, it just returns immediately.

probes/attach.go
```diff
@@ -0,0 +1,154 @@
...
+		exitLink, err := ex.Uprobe(s.ExitSymbol, a.progExit, &link.UprobeOptions{
+			Cookie: cookie,
+			PID:    0,
+		})
+		if err != nil {
+			log.Warnf("probes: attach exit %s @ %s: %v", s.ExitSymbol, req.filePath, err)
+			// Roll back the entry uprobe so we don't have orphan stack pushes
+			// with no exit to drain them.
+			if cerr := entryLink.Close(); cerr != nil {
+				log.Warnf("probes: rollback entry %s: %v", s.EntrySymbol, cerr)
```
umanwizard, line 123: is this expected to happen in real life for any good reason? If not, I wonder if we should just assert here because we're in such a weird state that I'm not sure it's worth continuing...

IDK though, that might be crazy, what do you think?

probes/bpf/probe.bpf.c
```diff
@@ -0,0 +1,151 @@
...
+// Two SEC programs (probe_entry, probe_exit) share a per-tid LRU hash map
+// tracking the lifetime of the OUTERMOST scope only. probe_entry captures
+// a timestamp when the outer scope opens (top transitions 0->1) and just
+// increments a counter for nested entries. probe_exit decrements; emits a
+// ringbuf event only when the outer scope closes (top transitions back to
+// 0). Inner scopes roll into the outer's measured duration -- they do not
+// get their own records.
+//
+// Cookie layout (set by user-space, same on both attach points):
+//   bits 63..32 : spec_id  (uint32, 1-based)
```
umanwizard, line 14: Can you clarify what "one-based" means? When I was reading this I assumed it meant that values were offset by one from normal uint32 (i.e., a bit-battern of 0x00000000 semantically means "1", and so on), but reading the rest of the code I'm now unsure. Also I'm not sure why 0 isn't a valid spec_id -- do we use it as a sentinel value?

probes/config.go
```diff
@@ -0,0 +1,114 @@
...
+// fallback. Probe regexes should accept either form, e.g. `(^|/)node$`.
+func LoadConfig(path string) ([]ProbeSpec, error) {
+	b, err := os.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("read probe config %s: %w", path, err)
+	}
+	return parseConfig(b)
+}
+
+func parseConfig(b []byte) ([]ProbeSpec, error) {
```
umanwizard, line 72: I really miss `serde` from Rust 😂

probes/loader.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/ringbuf"
+)
+
+//go:embed bpf
+var bpfFS embed.FS
+
+// probeEventSize is the on-the-wire size of `struct probe_event` from
+// probe.bpf.c. Keep the Go decoder in sync if the C struct changes.
+const probeEventSize = 48
```
umanwizard: can we generate this constant from a C header? If not, let's add a `_Static_assert` on the size to the C file, and a comment there that this needs to be kept in sync. Otherwise I'll definitely forget to change this if I ever modify something in C.

probes/loader.go
```diff
@@ -0,0 +1,134 @@
...
+
+// rawProbeEvent is the layout produced by bpf/probe.bpf.c. Decoded from
+// little-endian raw bytes; do not change field order without updating
+// the C struct.
```
umanwizard: same, can this be generated?

probes/service.go
```diff
@@ -0,0 +1,238 @@
...
+
+	"github.com/parca-dev/parca-agent/reporter"
+)
+
+// logBody is the OTLP LogRecord.body value emitted for every paired-scope
+// event. Stable string so dashboards can filter on it.
+const logBody = "node.callback_scope"
+
+// logScope is the instrumentation-scope name on the OTel Logger we obtain
+// from the reporter. PolarSignals surfaces this as attributes_scope.name, so
```
umanwizard: Polar Signals is two words! Also, is it possible to state this in terms of Parca rather than PS since this is in the open-source project? (I also asked the same on Discord; not sure if logging support is a PS-only feature)

flags/flags.go
gnurizen: Seems like overkill, we have options that end in "File" and "Dir" and "Path".   File seems like the best option to me.

probes/bpf/probe.bpf.c
gnurizen: I think if we need more per-spec information we would just add a per-spec map.

probes/bpf/probe.bpf.c
gnurizen: 0 is special because bpf_get_attach_cookie returns 0 when there is no cookie.   So we need to start with a non-zero value to differentiate from kernel/user-agent failures related to setting the cookie.   Otherwise if the kernel/user-mode failed to set the cookie for some reason we'd think it was the first probe.   Although we don't actually enforce that which I think is also fine, if a probe_event arrives in userland with 0 for spec id, we'll just log a warning and ignore it.

probes/loader.go
gnurizen: Good idea

gnurizen:
Review comments addressed and mods pushed as a new commit I'll squash later.   Forgot how cool bpf2go is, used it for oomprof too.

reporter/parca_reporter.go
```diff
@@ -241,6 +285,17 @@ func (r *arrowReporter) Logger(scope string) otellog.Logger {
...
+// Tracer returns an OTel Tracer bound to the given scope name. In offline
+// mode (no gRPC conn was supplied at construction) the SDK TracerProvider
+// is nil; we return the OTel no-op Tracer so callers can treat Tracer as
+// unconditional and Start/End become inert.
+func (r *arrowReporter) Tracer(scope string) oteltrace.Tracer {
+	if r.tracerProvider == nil {
+		return tracenoop.NewTracerProvider().Tracer(scope)
+	}
+	return r.tracerProvider.Tracer(scope)
+}
```
umanwizard, line 277: Hmm, should we error instead if you try to use traces in offline mode, rather than silently dropping them?

reporter/parca_reporter.go
```diff
@@ -843,6 +898,23 @@ func (r *arrowReporter) ReportExecutable(args *reporter.ExecutableMetadata) {
 		r.uploader.Upload(context.TODO(), mf.FileID, mf.FileName.String(), mf.GnuBuildID, open)
 	}
 
+	// Notify the probes hook on every ELF report, before the cache early-return.
+	// The attacher dedupes by FileID, so re-notification is harmless; this
+	// ensures binaries cached before SetProbes was wired still get a chance to
+	// attach when subsequently re-reported, and the SetProbes catch-up handles
+	// the case where the profiler never re-reports a long-cached binary.
```
umanwizard: I don't understand why we need to re-report when we already have the catch-up. Under what circumstances will we report something here that wouldn't have been caught anyway by the catch-up?

main.go
```diff
@@ -443,6 +444,28 @@ func mainWithExitCode() flags.ExitCode {
...
+		if grpcConn == nil {
+			return flags.Failure("--probe-config requires a remote-store; cannot be used in offline mode")
+		}
+		probesSvc, err := probes.Start(mainCtx, probes.StartConfig{
+			ConfigPath: f.ProbeConfigFile,
+		}, parcaReporter)
+		if err != nil {
+			// Probes are a non-essential feature; a bad config (typo in
+			// probes.yaml, missing symbol, etc.) shouldn't take the whole
+			// agent down. Log and continue without probe support.
```
umanwizard: How do we know if they are non-essential to the user's workflow? There may be users for whom probes not working is just as bad as any other possible misconfiguration of the agent; I'm not sure why we shouldn't just exit early if it's broken.

probes/attach.go
umanwizard: this is still inaccurate, unless I'm missing something.

probes/attach.go
```diff
@@ -0,0 +1,154 @@
...
+	a.mu.Unlock()
+
+	var matched []ProbeSpec
+	for _, s := range a.specs {
+		if s.FileMatchRE.MatchString(filePath) {
+			matched = append(matched, s)
+		}
+	}
+	if len(matched) == 0 {
+		log.Debugf("probes: received %s (fileID=%s) — no spec matched", filePath, fileID.StringNoQuotes())
```
umanwizard: Can we get rid of the non-ASCII characters in log messages?

probes/attach.go
umanwizard: If we're not going to attempt to do anything about this, it should at least be an error, since it completely prevents the feature from working.

reporter/parca_reporter.go
gnurizen: Not sure we have enough use cases lined up yet to answer that.  I mean we probably should be putting the traces into the offline profile data but for now I think its sufficient that we error if you have probes enabled and are in offline mode.

probes/attach.go
gnurizen: Happy to make it an error

probes/attach.go
gnurizen: sure I'll do a sweep

probes/attach.go
gnurizen: :+1:

reporter/parca_reporter.go
gnurizen: I can remove the noop bit since this is deadcode structurally, if grcpConn exists traceProvider won't be nil and else we won't get here.

reporter/parca_reporter.go
gnurizen: All this stuff was unnecessary and just Claude being too defensive, cleaned it up.

umanwizard (approved):

PR #3184, opened 2026-05-28
gpu: GPU PC sampling support in the agent

gnurizen:
Adds GPU profiling driven by parcagpu's CUPTI shim. The agent attaches to
the shim's USDT probes, receives per-pid GpuConfig (sampling factor,
clock_khz, SM count) from the cuda_gpu_config probe, and reports two GPU
views as Parca sample types.

Reporting model (non-merged, the default):

  gpu_kernel_time / nanoseconds   exact kernel durations (CUDA callbacks)
  gpu_pcsample    / count         PC samples — all PC activity, not just
                                  stalls; value is the raw sample count,
                                  period = NsPerSample = 2^SamplingFactor
                                  / clock_hz, so value × period totals ns

PC sampling observes all PC activity (normally scheduled instructions as
well as stalls), so it is reported as a count of samples with a per-sample
time weight in the period — mirroring CPU sampling (samples/count :
cpu/nanoseconds) — rather than as a pre-converted "stall time". The count
stays correct even if a sample arrives before its GpuConfig probe; only
the period is unknown until the config event lands (one-shot warn per pid
via gpu.WarnMissingGpuConfig).

MergeGpuProfiles (default false) opts into a single gpu_time/nanoseconds
sample type that folds both views together, differentiated by a gpu_view
label (kernel_time|pc_sample); there the PC count is converted to ns
inline.

Bumps opentelemetry-ebpf-profiler to c9ad5c39 (parca-dev fork main, where
the pc-sampling work has now been merged). The pc-sampling change attaches
the cuda_gpu_config USDT probe (so the gpu_config probe semaphore is
raised and the shim delivers sampling factor / clock / SM count to the
agent) and makes TraceEventMeta.OffTime carry the raw PC sample count.

Adapts parcagpu.go to two API shifts that came with the merge:
gpu.StoreStallReasonMap now takes []libpf.String (intern each name), and
process.Mapping was renamed process.RawMapping (whose Path field is a
plain string — callers libpf.Intern only when storing long-term).


flags/flags.go
```diff
@@ -160,6 +160,8 @@ type Flags struct {
 
 	EnableOOMProf       bool `default:"false" help:"Enable OOMProf profiling integration."`
 	EnableOOMProfAllocs bool `default:"false" help:"Enable OOMProf alloc counts."`
+
+	MergeGpuProfiles bool `default:"false" help:"Report GPU kernel timing and GPU PC sampling under a single gpu_time/nanoseconds sample_type, differentiated by a gpu_view label (pc_sample|kernel_time). When false (the default), they are reported as separate sample_types (gpu_kernel_time/nanoseconds and gpu_pcsample/count) with no per-sample labels."`
```
umanwizard, line 164: Did we come to a consensus on this? Should we talk about it in the meeting on Mon.?

umanwizard (approved):

flags/flags.go
gnurizen: I think everyone was fine with the keeping them separated approach but I wanted to hedge our bet by leaving this in.   Can probably remove it when we get some more feedback from the field.

PR #3185, opened 2026-06-01
feat: Snap OTLP tracing support

mmkay:
This PR adds support for passing `otlp-address` and `otlp-exporter` to the snap's config options. Although it is possible to define any set of parameters via `config-path`, allowing users to configure traces export without needing to manage their own config file would be useful for our parca-agent charm. 

The changes have been tested locally:

image


brancz (approved):

brancz:
Thank you!

PR #3186, opened 2026-06-08
reporter: emit CUDA PC-sample frames (cuda-pc) for per-function symbolization

gnurizen:
Agent-side counterpart to parca-dev/opentelemetry-ebpf-profiler#300. Reports GPU PC samples as a single `cuda-pc` frame so multi-kernel cubins can be symbolized **per function** while keeping one mapping per cubin (build_id = cubin CRC).

## Change

Adds the `libpf.CUDAPCFrame` case to **both** location-writer paths in the reporter:
- **V2** (`appendLocationV2`, the path the cloud/polarfusion symbolizer reads)
- **V1** (the record path)

For each: one mapping per cubin (`build_id = cubin CRC FileID`, never a per-function id), the kernel's mangled name as the **system name** of a placeholder line (`line 0`, no function name), and the `"cuda-pc"` frame type. The backend resolves the real source line per function off the kernel name (gated on the frame type) — that backend work is separate.

The CUPTI **stall reason** and decoded **SASS instruction** are no longer frames; they arrive as the `cuda_stall_reason` / `cuda_sass_instruction` sample labels (set on `trace.CustomLabels` in the ebpf-profiler fork, consumed generically here), independent of the frame.

### DIFF
--- a/reporter/parca_reporter.go
+++ b/reporter/parca_reporter.go
@@ -614,6 +614,32 @@ func (r *ParcaReporter) appendLocationV2(frame libpf.Frame) uint32 {
 			StartLine:  0,
 		}))
 
+	case libpf.CUDAPCFrame:
+		// CUDA PC sample: a function-relative kernel offset. One mapping per
+		// cubin (build ID = cubin CRC FileID, never a per-function ID). The
+		// kernel's mangled name rides as the system name of a placeholder line
+		// (line 0); the backend resolves the real source line per function,
+		// gated downstream on the "cuda-pc" frame type.
+		b.locFrameType.AppendString(frame.Type.String())
+
+		var fid libpf.FileID
+		if frame.Mapping.Valid() {
+			mf := frame.Mapping.Value().File.Value()
+			fid = mf.FileID
+			b.locMappingFile.AppendString(mf.FileName.String())
+		} else {
+			b.locMappingFile.AppendNull()
+		}
+		b.locMappingID.AppendString(fid.StringNoQuotes())
+
+		b.lineNumber.Append(0)
+		b.lineColumn.Append(0)
+		b.funcIndices.Append(b.funcDict.AppendFunction(FunctionV2{
+			SystemName: frame.FunctionName.String(),
+			Filename:   "",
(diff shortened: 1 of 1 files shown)


reporter/parca_reporter.go
```diff
@@ -614,6 +614,32 @@ func (r *ParcaReporter) appendLocationV2(frame libpf.Frame) uint32 {
...
+
+		var fid libpf.FileID
+		if frame.Mapping.Valid() {
+			mf := frame.Mapping.Value().File.Value()
+			fid = mf.FileID
+			b.locMappingFile.AppendString(mf.FileName.String())
+		} else {
+			b.locMappingFile.AppendNull()
+		}
+		b.locMappingID.AppendString(fid.StringNoQuotes())
```
brancz, line 633: The PR description says CRC but this uses the file id?

gnurizen: The CRC is used to create the fileID

reporter/parca_reporter.go
```diff
@@ -614,6 +614,32 @@ func (r *ParcaReporter) appendLocationV2(frame libpf.Frame) uint32 {
 			StartLine:  0,
 		}))
 
+	case libpf.CUDAPCFrame:
+		// CUDA PC sample: a function-relative kernel offset. One mapping per
+		// cubin (build ID = cubin CRC FileID, never a per-function ID). The
+		// kernel's mangled name rides as the system name of a placeholder line
```
umanwizard, line 620: What do you mean by a placeholder line? System name is a property of functions AFAIK, not of lines.

brancz: I think the comment is confusing, because the reality is confusing. We need the system name to be able to resolve the line number. But usually the symbolizer only knows to symbolize because in fact there are no lines in the lines array at all.

brancz (approved):

reporter/parca_reporter.go
gnurizen: Yeah, placeholder just means the line number is zero and will be filled in later.

PR #3199, opened 2026-07-14
reporter: OTLP TracerProvider with configurable backend

gnurizen:
## Summary

- Adds an OTel `TracerProvider` in the reporter, sibling to the existing `LoggerProvider`. Both are constructed on the shared gRPC connection and stamped with the same `Resource` (`service.name` / `service.version` / `host.name`). `Tracer(scope)` joins `Logger(scope)` on the `ParcaReporter` interface; offline mode returns the OTel no-op tracer so callers can treat `Tracer()` as unconditional. Batch tuning matches the log pipeline (4096 queue / 512 batch / 250 ms interval); `AlwaysSample` because producers filter at the source (e.g. probes drop by `min_duration_ms` in-kernel before a record ever reaches user space).
- Makes the trace destination configurable. `newTracerProvider` now takes any `sdktrace.SpanExporter` instead of a gRPC conn. `reporter.New` accepts a new `traceExporter` argument: an explicit exporter wins, otherwise it falls back to piggybacking on `grpcConn` (the remote-store connection), preserving today's default. When `--otlp-address` is set, `main.go` builds a second `flags.Exporter` (grpc | http | stdout, from `--otlp-exporter`) and hands it in, so reporter-emitted spans can be pointed at any OTLP endpoint (a collector, Honeycomb, etc.) independent of the profile-data path.
- Prep for the probes BPF service, which will be the first user of this pipeline (separate PR).

## Test plan

- [ ] `go build ./...` and existing test suite pass.
- [ ] With `--otlp-address` unset, no functional change: `Tracer()` returns a working tracer when `remote-store` is configured (spans ride the same gRPC conn as profiles) and a no-op in offline mode.
- [ ] With `--otlp-address` set to an OTel collector, spans arrive at the collector.
- [ ] With `--otlp-address` set and `--offline-mode-storage-path` set, spans still ship (previously all trace paths required a remote-store).


gnurizen:
Folded into #3181 as its first commit — the two changes ship as one PR now.

PR #3205, opened 2026-08-28
reporter: add OTLP/profiles as a third export format

gnurizen:
Adds OTLP/profiles as a third profile encoding alongside arrow v1 and v2,
selected with `--remote-store-format=otlp`. Verified end to end against two
independent OTLP/profiles receivers, both rendering symbolized flamegraphs.

Each commit builds and tests on its own:

- **`reporter: extract the backend-independent parts of arrowReporter`** — pure
  refactor, no behavior change. The label pipeline, symbol uploader, metrics
  bridge and frame classifier move out of `arrowReporter`, and `reporter.New`
  takes a `Config` instead of 24 positional parameters. `--remote-store-format`
  supersedes the `--use-v2-schema` bool, with two values at this point.
- **`reporter: add OTLP/profiles as a third export format`** — the encoder, plus
  `--remote-store-compression` and `--debuginfo-address`. The conversion is ours
  rather than upstream's, because `pprofile.Profile` carries one SampleType and
  upstream rejects the GPU origins before the encoding is reached.
- **`main: add --symbolize-go`** — independent of the encoding. Resolves Go
  frames from gopclntab in-agent, for backends with no symbolizer service. Off
  by default.
- **`reporter: only build the self-logs provider when logs are exported`** — the
  batch processor was running in every non-offline configuration regardless of
  `--otlp-logging`, allocating ~9% of the agent's total bytes.
- **`reporter: benchmark both encoders against a null gRPC endpoint`** — replays
  a real `.padata` corpus through both paths so the encoders can be compared
  without eh_frame extraction swamping the result.
- **`reporter: append a stack's location indices in one call`** — the largest
  single allocator in the OTLP path; −19% allocations.

On the same 20,000-sample corpus arrow v2 is ~1.6x cheaper in CPU and allocates
2.2x fewer objects, while OTLP is ~2.4x smaller on the wire. Both are a fraction
of a percent of a core at realistic sample rates.

Two known gaps, both on the store side rather than the agent: off-CPU is
`wallclock` on arrow and `off_cpu` on OTLP, which is a naming decision still to
be made; and OTLP has no field for aggregation temporality, so a consumer has to
derive delta from the sample type. Rationale for each change is in its commit
message.


flags/compression.go
```diff
@@ -0,0 +1,52 @@
...
+// unrelated to the LZ4 inside an Arrow v2 payload.
+//
+// It must run before any dial: SetLevel mutates the registered compressor's
+// writer pool, which grpc-go documents as initialization-time only.
+func ConfigureCompression(name string) (grpc.CallOption, bool, error) {
+	switch name {
+	case CompressionNone, "":
+		return nil, false, nil
+
+	case CompressionGzip:
```
brancz, line 41: I think the agent needs to work together with a collector and in-cluster network traffic spending the cpu is not necessary. Doesn't hurt to have this.

reporter/pprofile.go
```diff
@@ -0,0 +1,518 @@
...
+
+	sample := profile.Samples().AppendEmpty()
+	sample.SetStackIndex(stackIdx)
+	sample.TimestampsUnixNano().Append(s.Timestamp)
+	// One value per timestamp, matching the profile's single SampleType.
+	sample.Values().Append(s.Value)
+
+	attrs := sample.AttributeIndices()
+	s.SampleLabels.Range(func(l labels.Label) {
+		switch l.Name {
```
brancz, line 457: are these really the only 4 that semconv values exist for? Maybe it would be better if we integrated this mapping directly into the metadata providers directly so it's easier to maintain?

brancz (approved):

flags/compression.go
gnurizen: Yeah I made a mental note to turn this off in our local testing rig and it should probably go in our docs somewhere that none is preferred if you are just sending profiles to a sidecar pod over loopback.

reporter/pprofile.go
gnurizen: Okay I stewed on this for a bit and made some related changes, please take another look!

PR #3229, opened 2026-09-21
reporter: make the OTLP profiles output conformant

gnurizen:
The OTLP/profiles output was not conformant. Checked with
[`profcheck`](https://github.com/open-telemetry/sig-profiling), the
conformance checker the ebpf-profiler already uses in
`reporter/internal/pdata`: **0 of 24 live exports passed before, 37 of 37
pass after.**

## Dictionary sentinels

Every OTLP profiles dictionary table reserves index 0 for a zero value, so
that an index of `0` reads as "unset" rather than as a reference to the
first real entry. `reset()` only did that for the string table, so the
first real entry of every other table landed at index 0:

| table | what shipped at index 0 |
| --- | --- |
| `link_table` | nothing — the table was empty, which the spec forbids outright |
| `location_table` | a real location (`Address=0x3e82ca`, lines, attributes) |
| `stack_table` | a real stack (`LocationIndices=[0..27]`) |
| `attribute_table` | a real attribute (`KeyStrindex=6`) |

`mapping_table` and `function_table` happened not to trip the checker on
the payloads I captured, but they carry the same rule, so they get the
same treatment.

## Sample timestamps

`Build` stamped the flush window, which is when the exporter ran, not when
the kernel took the samples. A sample read from the perf buffer can predate
it by tens of milliseconds, and OTLP requires every
`timestamps_unix_nano` to satisfy `time <= ts < time + duration`. The
builder now tracks the batch's timestamp range and widens the window to
cover it.

## Note on `link_table[0]`

The proto comment says the 16/8 zero-filled form SHOULD be preferred "for
better compatibility with codecs strictly expecting 16/8 byte array
lengths", but `profcheck` compares against a bare `Link{}` with
`proto.Equal`, which a zero-filled one fails. Since we emit no links at
all, this uses the strict zero value; it is commented in place.

## Tests

`reporter/pprofile_conformance_test.go` goes through the proto wire form
rather than the pdata view, so it sees what a collector receives — these
rules are invisible in the pdata API, which is why nothing caught them:

- `TestPprofileConformance` — a populated batch through the real builder
- `TestPprofileDictionarySentinels` — index 0 reserved, and actually zero, in all six tables
- `TestPprofileConformanceSampleOutsideWindow` — samples before the start and on the exclusive end boundary

One existing assertion moved with the sentinel (`pprofile_test.go`,
`StackTable().Len()` 1 → 2).

Verified end to end by capturing real exports off the wire from an agent
built at this commit, not only in unit tests.


reporter/pprofile.go
```diff
@@ -526,14 +566,27 @@ func (b *pprofileBuilder) SampleCount() int { return b.sampleCount }
 // Build stamps the collection window onto every profile and hands over the
 // batch, leaving the builder empty for the next interval.
 func (b *pprofileBuilder) Build(start, end time.Time) pprofile.Profiles {
-	duration := uint64(end.Sub(start).Nanoseconds())
+	// Widen the window to cover every sample in the batch. OTLP requires each
+	// timestamps_unix_nano to satisfy time <= ts < time + duration, and the
+	// flush window alone does not: samples are stamped when the kernel took
+	// them, so the earliest can sit before start, and the latest can land on
+	// or after end. The end is exclusive, hence the +1.
```
umanwizard: I agree with inclusive start and exclusive end, but in that case, it should be called something other than "max".

umanwizard (approved):

PR #3230, opened 2026-09-22
reporter: name OTLP resources from OTEL_SERVICE_NAME

gnurizen:
`service.name` on the OTLP path fell back to `comm`, which is `bpf_get_current_comm()` — the **thread** name, not the process name. Any process that names its threads therefore reports as several resources rather than one.

Measured on a developer laptop against a live collector:

| process | resources emitted | examples |
|---|---|---|
| one JVM | 24 | `G1 Refine#0`, `ForkJoinPool.co`, `Asynchronous lo` |
| chrome | 12 | |
| claude | 11 | |
| a Kafka broker | — | arrived as `XT)-PLAINTEXT-1` |

## What this does

Reads the name the process declared for itself, giving the precedence:

```
APMServiceName → OTEL_SERVICE_NAME → OTEL_RESOURCE_ATTRIBUTES[service.name] → comm → basename
```

`OTEL_SERVICE_NAME` over `OTEL_RESOURCE_ATTRIBUTES` is the precedence the OTel environment variable spec defines. Both sit below `APMServiceName`, which comes from live instrumentation. `comm` still outranks the executable basename, for the reason already documented there — a basename collapses every interpreted service into `python3`.

The capture plumbing already existed behind `--include-env-var`: the profiler reads `/proc/<pid>/environ` once per process and hands back an interned map. The OTLP path now requests these two variables unconditionally, because an operator should not have to know to ask for the one variable OTel defines for this purpose. It costs two more string compares per process; the file is read once either way.

`OTEL_RESOURCE_ATTRIBUTES` is W3C Baggage-shaped, so the parser strips the `;`-delimited metadata part and percent-decodes the value.

## Caveat: process context must outrank this

This reads the environment, which is **fixed at exec**. When the agent grows process context support — a process publishing its own resource attributes for the profiler to read at sample time — that source is both authoritative and able to change during the process's life, so it must take precedence over anything recovered here. The caveat is recorded in the doc comment on `serviceNameFor` so it is in front of whoever adds that path.

## Testing

Table test extended to cover the new precedence, baggage metadata stripping, percent-decoding, an empty `OTEL_SERVICE_NAME` falling through rather than blanking the name, and `service.namespace` not matching on the shared prefix.


brancz:
TIL `OTEL_RESOURCE_ATTRIBUTES` lgtm

brancz (approved):

